PR #25, opened 2022-08-04
correcting issue where wrong result is show for arena matches

ericlytle:
gamepreviewfixed

### DIFF
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -1,4 +1,5 @@
 /* eslint import/prefer-default-export: off, import/no-mutable-exports: off */
+import internal from 'stream';
 import { startRecording, stopRecording, isRecording, isRecordingCategory}  from './main';
 
 const tail = require('tail').Tail;
@@ -10,6 +11,12 @@ let currentLogFile: string;
 let lastLogFile: string;
 let videoStartDate: Date; 
 
+type CombatantData = 
+{
+    teamId: string;
+    friendly: boolean;
+}
+
 type Metadata = {
     name: string;
     category: string;
@@ -20,6 +27,7 @@ type Metadata = {
 }
 
 let metadata: Metadata;
+let combatantInfoMap: Map<string, CombatantData> = new Map();
 
 /**
  * getLatestLog 
@@ -77,8 +85,54 @@ const handleLogLine = (line: string) => {
         handleRaidStopLine(line);
     } else if (line.includes("ZONE_CHANGE")) {
         handleZoneChange();
-    }  
-}    
+    } else if (line.includes("COMBATANT_INFO")) {
+        handleCombatantInfoLine(line);
+    } else if (line.includes("SPELL_AURA_APPLIED")){
+        handleSpellAuraAppliedLine(line);
+    }
+}
+
+/**
+ * Handles the SPELL_AURA_APPLIED line from WoW log.
+ * @param line the SPELL_AURA_APPLIED line
+ */
+const handleSpellAuraAppliedLine = (line: string) => {
+    if (combatantInfoMap.size > 0)
+    {
+        let sourcePlayerId = line.split(',')[2];
+        let sourceGUID = line.split(',')[3];
+        let playerData = combatantInfoMap.get(sourcePlayerId)
+        if (playerData != undefined)
+        {
+            playerData.friendly = isFriendlyUnit(parseInt(sourceGUID));
+        }
+    }
+}
+
+/**
+ * Handles the COMBATANT_INFO line from WoW log.
+ * @param line the COMBATANT_INFO line
+ */
+const handleCombatantInfoLine = (line: string) => {
+    let combatantData: CombatantData;
+    combatantData = {
+        friendly: false,
+        teamId: "",
+    }
+    const playerKey = line.split(',')[1];
+    combatantData.teamId = line.split(',')[2];
+    combatantInfoMap.set(playerKey, combatantData);
+}
+
+/**
+ * Determines if the source GUID is from a friendly unit.
+ * @param sourceGUID the source GUID
+ * @returns true if sourceGUID belongs to a friendly unit; false otherwise. 
+ */
+const isFriendlyUnit = (sourceGUID: number): boolean => {
+    const masked = sourceGUID & 0x000000f0;
+    return masked == 0x00000010;
+}
 
 /**
  * Handle a line from the WoW log. 
@@ -105,8 +159,19 @@ const handleArenaStartLine = (line: string) => {
     const videoStopDate = new Date();
     const milliSeconds = (videoStopDate.getTime() - videoStartDate.getTime()); 
     metadata.duration = Math.round(milliSeconds / 1000);
-    metadata.result = Boolean(parseInt(line.split(',')[1]));
+    metadata.result = determineArenaMatchResult(line);
     stopRecording(metadata);
+}
+
+/**
+ * Determines the arena match result
+ * @param line the line from the WoW log. 
+ * @returns true if you won the match; otherwise false
+ */
+const determineArenaMatchResult = (line: string): boolean => {
+    const [playerData] = combatantInfoMap.values();
+    const winningTeamId = line.split(',')[1];
+    return playerData.friendly ? playerData.teamId == winningTeamId : !(playerData.teamId == winningTeamId);
 }
 
 /**
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -160,6 +160,7 @@ const handleArenaStartLine = (line: string) => {
     const milliSeconds = (videoStopDate.getTime() - videoStartDate.getTime()); 
     metadata.duration = Math.round(milliSeconds / 1000);
     metadata.result = determineArenaMatchResult(line);
+    combatantInfoMap.clear();
     stopRecording(metadata);
 }
 
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -105,6 +105,7 @@ const handleArenaStartLine = (line: string) => {
     const videoStopDate = new Date();
     const milliSeconds = (videoStopDate.getTime() - videoStartDate.getTime()); 
     metadata.duration = Math.round(milliSeconds / 1000);
+    metadata.result = Boolean(parseInt(line.split(',')[1]));
     stopRecording(metadata);
 }
 


aza547:
Hi and thanks for the PR. I'm a little nervous about this as I had some trouble with the ARENA_MATCH_END event and eventually convinced myself that it was bugged and wasn't reliable as per my issue notes. 

See: https://wowpedia.fandom.com/wiki/COMBAT_LOG_EVENT#ARENA_MATCH_START_/_END 

> ARENA_MATCH_START: instanceID, unk, matchType, teamId
ARENA_MATCH_START,1505,0,Skirmish,0

> ARENA_MATCH_END: winningTeam, matchDuration, newRatingTeam1, newRatingTeam2
ARENA_MATCH_END,0,58,0,0

Here's 3 games I played last night start/end events in logs:

-- Loss
> 8/3 22:01:00.971  ARENA_MATCH_START,2509,33,2v2,1
8/3 22:04:12.329  ARENA_MATCH_END,0,191,1581,1709

-- Win
> 8/3 22:06:16.036  ARENA_MATCH_START,1825,33,2v2,1
8/3 22:08:01.293  ARENA_MATCH_END,0,105,1659,1550

-- Win
> 8/3 22:09:58.548  ARENA_MATCH_START,2547,33,2v2,1
8/3 22:12:14.889  ARENA_MATCH_END,0,136,1673,1668

Maybe I'm missing something obvious here, but I've no idea what to make of this! Notice that no combination of the "winningTeam" argument seems to be correct. Nor does it appear to be a boolean to indicate win/loss. 

Your change assumes that if the first argument of arena match end is 1 then it's a win (but that contradicts the above). Thoughts? Would be interested to see your logs.

I also asked this question to the friendly guys at wowarenalogs and got this response, bad haven't really looked at it till now. 

> I haven't personally seen a case where ARENA_MATCH_END was written to the log with wrong info -- do you have a record of that? I have seen the _END event not get written at all which can be problematic. Our core log parser is in https://github.com/wowarenalogs/wowarenalogs/tree/main/packages/wow-combat-log-parser although it's fairly complex now.

aza547:
See here: https://github.com/wowarenalogs/wowarenalogs/blob/d07ca844b404db0aa478554bbeb13053df7bfdf4/packages/wow-combat-log-parser/src/CombatData.ts#L380

I think the logic here is actually circumventing whatever nonsense is in the combatlog and taking the last players death reaction to determine win/loss.

That seems robust enough, although there is probably some crazy timing window for a hunter to hit feign death and break it?

aza547:
Actually there is probably still a condition for it to go wrong here. 
- If you are a hunter, or have a hunter on your team. 
- Kill an enemy player, UNIT_DIED event fires. 
- Hunter feign deaths. UNIT_DIED event fires (despite being a feign)
- Other enemies leave. 

Here's an example, where I win the game. 
- My hunter (Bogson) kills the enemy maybe Ravyne. 
- He Feign deaths after. Enemy rogue leaves. 
- ARENA_MATCH_EVENT fires. Last unit to die in the log was my hunter. 

> $ cat gamelog.txt  | grep "UNIT_\|ARENA_MATCH"
8/3 22:09:58.548  ARENA_MATCH_START,2547,33,2v2,1
8/3 22:10:26.566  UNIT_DIED,0000000000000000,nil,0x80000000,0x80000000,Player-1329-09C34603,"Bogson-Ravencrest",0x10512,0x0,1
8/3 22:10:54.219  UNIT_DIED,0000000000000000,nil,0x80000000,0x80000000,Player-1329-09C34603,"Bogson-Ravencrest",0x10512,0x0,1
8/3 22:12:04.738  UNIT_DIED,0000000000000000,nil,0x80000000,0x80000000,Player-580-0A3E1BD5,"Ravyne-Blackmoore",0x10548,0x0,0
8/3 22:12:11.356  UNIT_DIED,0000000000000000,nil,0x80000000,0x80000000,Player-1329-09C34603,"Bogson-Ravencrest",0x10512,0x0,1

I might try upload this log to wowarenalogs and see what it makes of it...

aza547:
Interestingly it got it correct: https://wowarenalogs.com/matches/db04a341fa59f3b1275d032db70dfb34

I'll keep digging in their code...

ericlytle:
Oh interesting. No i have not had that issue at all. let me grab a couple of my log files and post them. I did dig through my matches from the previous night and they seemed to have the correct item in event log.

ericlytle:
Here are a couple of mine:
`8/4 13:34:43.753  ARENA_MATCH_END,1,86,0,0`

`8/4 00:25:43.935  ARENA_MATCH_END,1,71,1159,1910`

`8/4 00:18:06.633  ARENA_MATCH_END,0,90,1648,1657`

all of these have the correct results to them

aza547:
Suspect that sample size of 3 could just be getting lucky, do you have some more?

Chatting to wowarenalogs dev just now, I think we need to parse out `COMBATANT_INFO` to determine the players teamID and then use the ARENA_MATCH_END event which tells us the winningTeamID.

ericlytle:
The more i look at this. I wonder if the parameter is misleading. Maybe it is not a boolean to indicate if YOUR team won, but really to indicate WHICH team one. Meaning if your team is "team1" that would be index 0 and if your team is "team2" then your index is 1

so in this case and we have the follow scenarios and results, it would make sense
Team 1 wins over team 2
8/4 00:18:06.633 ARENA_MATCH_END,0,90,1648,1657

team 2 wins over team 1
8/4 00:18:06.633 ARENA_MATCH_END,1,90,1648,1657


if this is the case, then maybe it just needs to be figured out which "team" you are apart of and then use this index to correlate the two to determine if you won. 

Ill dig into this a little when i have more time.

ericlytle:
Looking at ARENA_MATCH_START, this looks to be the case. this should be very easy to implement
```
ARENA_MATCH_START: instanceID, unk, matchType, teamId
ARENA_MATCH_START,1505,0,Skirmish,0

ARENA_MATCH_END: winningTeam, matchDuration, newRatingTeam1, newRatingTeam2
ARENA_MATCH_END,0,58,0,0
```

START event tells you which team YOU belong to and then end tells you WHICH team won. so you would just have to compare the two. If the END event winning team id matches your team id, then you won the match. otherwise you lost

ericlytle:
I will test this tonight when i can get into some wargames, but i have the code in my local branch ready to test. 

Ill get an executable built out and give it to my team and the opposing team to confirm this fix works
New global variable for team id
```
let currentTeamId: number;
```
Modification to arena start event
```
const handleArenaStartLine = (line: string) => {
    const zoneID = parseInt(line.split(',')[1]);
    const category = line.split(',')[3];
    videoStartDate = new Date();
    currentTeamId = parseInt(line.split(',')[4]);
```
Modification to arena stop event
```
 const handleArenaStopLine = (line: string) => {
    const videoStopDate = new Date();
    const milliSeconds = (videoStopDate.getTime() - videoStartDate.getTime()); 
    metadata.duration = Math.round(milliSeconds / 1000);
    metadata.result = Boolean(parseInt(line.split(',')[1]) == currentTeamId);
```

aza547:
Hm, I still don't think that works, check my log snippets again.

I think the last arg in ARENA_MATCH_START is actually if it's a ranked game or not, and wowpedia is incorrect.

ericlytle:
Dang you are right. sorry about that. i should have looked back at your logs. interesting.

ericlytle:
You are for sure right on that. 
```
8/4 13:33:17.582  ARENA_MATCH_START,1911,33,Skirmish,0
```
says 0 here even though i won and the end shows
```
8/4 13:34:43.753  ARENA_MATCH_END,1,86,0,0
```

aza547:
OK here's a strategy shamelessly stolen from wowarenalogs:

- Get a random SPELL event e.g. SPELL_AURA_APPLIED.
- That has sourceGUID, sourceFlags. 
- SourceGUID can be used to link back to COMBATANT info event. 
- SourceFlags is a bitwise operator which tells you the reaction from the observers point of view (i.e. your reaction to them).
    - See [wowpedia](https://wowpedia.fandom.com/wiki/UnitFlag) and [wowarenalogs code](https://github.com/wowarenalogs/wowarenalogs/blob/d07ca844b404db0aa478554bbeb13053df7bfdf4/packages/wow-combat-log-parser/src/utils.ts#L34).
- From that info you can determine which team the observer (player) is, which you can then link to the ARENA_MATCH_END event indicator. 

Fancy having a go at implementing that? Bit more complex than it started!

ericlytle:
I will give it a shot! that was close to my idea as well i was posting -- 

Difficult part is just determining your player GUID to determine if you are on the winning team or not. 

One concern i do have is the combatant info defines the "team" as faction here: https://wowpedia.fandom.com/wiki/COMBAT_LOG_EVENT#COMBATANT_INFO

so i wonder how this plays out now that cross faction is enabled. 

But as you have demonstrated, this wiki is not always right, so lets hope that doesnt cause any issues.

aza547:
Fingers crossed that's just a remnant of the old system - I think back in the day only horde vs alliance games happened? Maybe I've made that up. Good luck, look forward to seeing the code!

prsauer:
The wiki isn't right -- the teamId doesn't map to faction. In fact, I don't think there is a good way (other than racials) to determine faction/race for players. It isn't reported in the COMBATANT_INFO event like one might expect. Could query the armory with their name but that's not guaranteed...

ericlytle:
Can you confirm something for me, can you find a spell aura applied that is from yourself and tell me if the GUID is 0x511, i am seeing this for every item in my log but i want to double check this is a constant used by blizzard and something i can code against. 

if this is the case, it will be much easier to code up

prsauer:
> 

Can check https://github.com/wowarenalogs/wowarenalogs/blob/d07ca844b404db0aa478554bbeb13053df7bfdf4/packages/wow-combat-log-parser/src/utils.ts#L34 for reaction flags.

Recommend joining the wow arena logs discord at https://discord.gg/64jwtUkM

ericlytle:
I have this working. just gotta clean up the code a little. should have a pr sometime tomorrow

src/main/logutils.ts
```diff
@@ -77,8 +85,54 @@ const handleLogLine = (line: string) => {
...
+ */
+const handleCombatantInfoLine = (line: string) => {
+    let combatantData: CombatantData;
+    combatantData = {
+        friendly: false,
+        teamId: "",
+    }
+    const playerKey = line.split(',')[1];
+    combatantData.teamId = line.split(',')[2];
+    combatantInfoMap.set(playerKey, combatantData);
```
prsauer, line 124: `combatantInfoMap` needs to be reset at some point between matches

ericlytle: addressed on this commit: https://github.com/aza547/wow-recorder/pull/25/commits/f493c35b35fe44a5feb4c120180f86ed19596fe2

ericlytle: addressed here: https://github.com/aza547/wow-recorder/pull/25/commits/f493c35b35fe44a5feb4c120180f86ed19596fe2

prsauer: Still broken when players afk or disconnect, end event won’t get seen. Better off putting the reset in start

prsauer: Pretty sure this will be buggy when they play bgs/raids as well -- iirc the combatant_info event fires for those as well, which will add players to the map that werent part of the arena

prsauer: @ericlytle @aza547 👀

aza547: Good points @prsauer. I've already merged... so some actions below: 

The afk/disconnect stuff: currently I think way worse things will happen on a DC (we'll just record forever). I should raise an issue to address that. Actually sounds like a tricky one to deal with. 

BG support: something I'm hoping to add over the weekend. I'll come back to this then. 

Raids: I think we just don't want to hit this code in raids, the outcome is trivial as it's in the ENCOUNTER_END event. Probably should gate this behind some condition. 

Another thought I had is performance. We only really need to hit the handleSpellAuraAppliedLine() once per arena game. I think currently we'll hit it every time the SPELL event fires. More conditional logic to fix this I think is needed.

prsauer: It might make sense to write a single end() function and call it every time something happens that would end a session (zone change, game closing, _end events, etc)

Issue #22, opened 2022-07-31
Start-up is fragile if video folder isn't in correct state


aza547:
e.g. if there is a mp4 not accompanied by a json it will hit an error.

better behaviour would be just to omit that video from the ones displayed. 

Easy to hit this as json file isn't created till end of recording -- such as if user closes app mid recording.

aza547:
This is in main. Will be fixed when 2.2.0 is released. Thanks @ericlytle!

PR #31, opened 2022-08-05
Correct https://github.com/aza547/wow-recorder/issues/22

ericlytle:
Also removed unused methods that existed in the file.

### DIFF
--- a/src/main/util.ts
+++ b/src/main/util.ts
@@ -59,19 +59,20 @@ const loadAllVideos = (storageDir: any, videoState: any) => {
  const loadVideoDetails = (video: any, videoState: any) => {
     const dateObject = new Date(fs.statSync(video.name).mtime)
     const metadata = getMetadataForVideo(video)
-
-    videoState[metadata.category].push({
-        index: videoIndex[metadata.category]++,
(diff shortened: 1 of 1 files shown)


src/main/util.ts
```diff
@@ -59,19 +59,20 @@ const loadAllVideos = (storageDir: any, videoState: any) => {
...
-        zone: getVideoZone(metadata.zoneID, metadata.category),
-        zoneID: metadata.zoneID,
-        encounter: getVideoEncounter(metadata),
-        encounterID: metadata.encounterID,
-        duration: metadata.duration,
-        result: metadata.result, 
-        date: getVideoDate(dateObject),
-        time: getVideoTime(dateObject)
-    });
+    if (metadata != undefined) {
```
aza547, line 62: I think you want `!==` here (see https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a).

ericlytle: interesting. didnt even know that existed. 
```suggestion
    if (metadata !== undefined) {
```

aza547 (approved):
Thanks!

PR #36, opened 2022-08-06
Fix ModuleError

jgorset:
Hi! Just a small fix here that I found.

We're trying to import 'Settings.css', but the file is called 'settings.css', and on some systems
(such as Linux) this is case sensitive:

  ERROR in ./src/settings/App.tsx 13:0-25
  Module not found: Error: Can't resolve './Settings.css' in '/home/jgorset/code/jgorset/wow-recorder/src/settings'
  @ ./src/settings/index.tsx 10:30-46


aza547:
I think given our discussion in discord we don't need this now - dev mode isn't supported on linux and will no doubt hit a bunch of other issues.

Issue #37, opened 2022-08-06
Bad quality on high resolution. 


aza547:
Seems our bitrate setting isn't getting plumbed through to OBS correctly. 

Videos seem to have 2-4M bitrate when its set to 60M.

aza547:
![image](https://user-images.githubusercontent.com/32389325/183248949-209fada6-334a-4cc5-aa40-7a2ede9d8fe9.png)

aza547:
Fixed. Closing. Will see how we get on with file sizes.

PR #39, opened 2022-08-07
Quality improvements

aza547:
- Remove bitrate cap. 
- Fix resolution so not 1920x1080 hardcoded.
- Remove some random packages we don't use. 
- Small tweaks to code style elsewhere.


aza547:
image

aza547:
Mentioned I am a bit worried about file size, but actually 120MB a min is probably not terrible. It will let the size monitor actually do its job.

aza547:
Will fix https://github.com/aza547/wow-recorder/issues/37

aza547:
Hopefully some weird env thing but I did try a skirmish and the entire recording didn't happen. Going out now so no time to investigate. Surely fine, as python test script all behaves as expected.

PR #48, opened 2022-08-28
Recording buffer

aza547:
Initial go at buffering recording.


src/main/recorder.ts
```diff
@@ -0,0 +1,181 @@
...
+     * 
+     */
+    set isRecordingBuffer(value) {
+        this._isRecordingBuffer = value;
+    }
+
+    /**
+     * init
+     */
+    init = () => {
```
ericlytle, line 62: is there any reason we really need this init to be called? init methods are a little awkward for the consumers to use, most of the time this type of stuff can just be done in the constructor. 

if you prefer to have this, do we want any protection around this -- IE illegal state exceptions if this hasnt been called first

aza547: no reason- i'll add to the constructor

PR #56, opened 2022-08-30
set window title to "Warcraft Recorder"

### LINKED ISSUE #55: Window title (Taskbar/alt+tab etc) is "Arena Recorder"

csutcliff:
fixes #55

### DIFF
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -38,6 +38,11 @@ let playerCombatant: Combatant | undefined;
 let isWowRunning: boolean = false;
 
 /**
+ * Is Challenge Mode recording in progress?
+ */
+let isRecordingChallengeMode: boolean = false;
+
+/**
  * wowProcessStarted
  */
 const wowProcessStarted = () => {
@@ -120,6 +125,10 @@ const handleLogLine = (line: string) => {
         handleEncounterStartLine(line);
     } else if (line.includes("ENCOUNTER_END")) {
         handleEncounterStopLine(line);
+    } else if (line.includes("CHALLENGE_MODE_START")) {
+        handleChallengeModeStartLine(line);
+    } else if (line.includes("CHALLENGE_MODE_END")) {
+        handleChallengeModeStopLine(line);
     } else if (line.includes("ZONE_CHANGE")) {
         handleZoneChange(line);
     } else if (line.includes("COMBATANT_INFO")) {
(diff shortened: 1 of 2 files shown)


ericlytle:
PR title is a little misleading as this also seems to be attempting to add M+ support

src/main/logutils.ts
```diff
@@ -117,9 +122,13 @@ const handleLogLine = (line: string) => {
...
     } else if (line.includes("ENCOUNTER_START")) {
-        handleRaidStartLine(line);
+        handleEncounterStartLine(line);
     } else if (line.includes("ENCOUNTER_END")) {
-        handleRaidStopLine(line);
+        handleEncounterStopLine(line);
+    } else if (line.includes("CHALLENGE_MODE_START")) {
+        handleChallengeModeStartLine(line);
+    } else if (line.includes("CHALLENGE_MODE_END")) {
+        handleChallengeModeStopLine(line);
```
ericlytle, line 131: if a group disbands, i dont think we get the challenge mode end event, right?

src/main/logutils.ts
```diff
@@ -262,6 +286,69 @@ const determineArenaMatchResult = (line: string): any[] => {
...
+     * so in theory if you start a new keystone before a failed key has been stopped the
+     * video will still be stopped/started in quick succession to ensure a separation.
+     */
+
+    const encounterID = parseInt(line.split(',')[1]);
+    const category = "ChallengeModes";
+
+    videoStartDate = getCombatLogDate(line);
+
+    metadata = {
```
ericlytle, line 304: we probably want to capture the key level too

this is actually why i stopped working on this. I originally was just going to add all this detail to this metadata but then i realized we are going to have alot of dead properties depending on the gameplay type. so the best way was to create an abstract class with the common attributes and then have other gameplay types extend that depending on their use case.

csutcliff:
> PR title is a little misleading as this also seems to be attempting to add M+ support

ah that's my fault, I didn't make this a seperate branch so when I continued working on the m+ support the PR here reflected that. Still fairly new to Git. I'll see what I can do to sort it out.

src/main/logutils.ts
csutcliff: No, I was planning on working on a combat log timeout for this before I sent it in a PR. Reading the comments I believe that could help for BGs too?

src/main/logutils.ts
csutcliff: That makes sense, the key level is certainly something useful to display to the user.

PR #58, opened 2022-08-30
set window title to "Warcraft Recorder"

### LINKED ISSUE #55: Window title (Taskbar/alt+tab etc) is "Arena Recorder"

csutcliff:
fixes #55 

resubmitted as I messed up with #56


aza547:
Thanks for the fix!

PR #59, opened 2022-08-30
[WIP] Mythic plus

csutcliff:
I've made a start with mythic plus (challenge mode).

After I messed up the #56 PR I figured it could useful to post it here even in the current state for feedback etc.

I'm planning on using a timeout on the combat log to handle leaving a keystone before completion without then starting another. Key attempts in quick succession should already work as, at least in my logs, starting a key appears to fire CHALLENGE_MODE_END before it fires CHALLENGE_MODE_START even if you were not previously in a key.


aza547:
Had a skim. Looks promising! Looking forward to seeing more!

tdaugaard:
I'm working on my own implementation on this as well, but it's currently a bit more elaborate and will probably be split into multiple PRs due to the size and refactoring. Check it out if you want https://github.com/tdaugaard/wow-recorder/tree/feature/implement_mythic_plus_support

Issue #23, opened 2022-08-02
Failed to launch, fix config and restart


jgorset:
Hi! Really looking forward to using this, but having an issue with getting it up and running.

![image](https://user-images.githubusercontent.com/195266/182384692-a9bc1cc8-61a7-4c2b-a15f-31e4014801ed.png)

This _should_ work, right? All these paths exist, and I did relaunch it, yet it keeps giving me the same error. Is there something I should be looking out for?

jgorset:
That's weird, after a few hours now it's changed to "waiting for a game to start", which seems promising! In the mean time, I installed Obs by itself (figuring maybe it was a prerequisite) and updated Windows with some minor patches. Maybe one of those things fixed it?

aza547:
Thanks for the report. I had a bug recently that the recording would fail to start if there was no combatlog files in the logs directory. I thought I had fixed it but I wonder if that could be the cause. 

Is it possible that you have never used combatlogging before and thus had no combatlog files present initially? If so I'd like to re-open this to make a proper fix!

Windows updates or OBS installation are unlikely to be the difference here.

aza547:
Your config looks good fyi, and the waiting for the game to start is the working state.

If you're tech savvy you could check our the python scripts under the test folder, they are very basic and would let you test it out with minor modifications to the script without even opening WoW (or alternatively just queue up for a skirmish)!

Suspect it's the no combat log files thing and you have indirectly created some by either installing the SCL addon or typing /combatlog.

aza547:
I'm opening this again - I should investigate this. It sucks if new users get hit with this (especially if above is correct and it's already diagnosed!).

jgorset:
Sorry for my late reply! I'm pretty sure it's exactly as you said -- I didn't have combat logs prior to installing this, so there weren't any in the directory, and that probably caused the error message!

aza547:
Still want to fix this, it's a crap first time user experience.

aza547:
Are you sure? This issue is worked around by closing and re-opening the application. It shouldn't appear again after that. Sounds like it might be another issue?

computerchipdotbiz:
I am VERY sure, also, when I go to close it from the system tray, it throws this error every time as well. 



![wcr](https://user-images.githubusercontent.com/5994477/187772038-9942831e-34f3-42d3-8db0-3e012f1bee6e.png)

aza547:
Aware of this one actually. I'll make a fix for that. 

[...]

PR #62, opened 2022-08-30
[Issue 23] Correct launch config issue

ericlytle:
Correct https://github.com/aza547/wow-recorder/issues/23


src/main/main.ts
```diff
@@ -151,11 +151,7 @@ const createWindow = async () => {
     if (!mainWindow) throw new Error('"mainWindow" is not defined');
 
     // Check we have all config, and at least one log is in the log directory.
```
aza547, line 153: Don't need this comment anymore - fn name is obvious

src/main/main.ts
```diff
@@ -246,7 +242,23 @@ const openPathDialog = (event: any, args: any) => {
...
+  if (mainWindow === null) return false; 
+  return isConfigReady(cfg) && (getLatestLog(baseLogPath));
+}
 
+/**
+ * Updates the status icon for the application.
+ * @param status the status number
+ */
+const updateStatus = (status: number) => {
+  if (mainWindow === null) return;
```
aza547, line 259: logic is a bit odd, how about 

`if (mainWindow !== null) mainWindow.webContents.send('updateStatus', status);`

src/main/main.ts
```diff
@@ -291,10 +303,19 @@ ipcMain.on('settingsWindow', (event, args) => {
...
   if (args[0] === "quit") {
     console.log("User closed settings");
     settingsWindow.close();
+    storageDir = getStringConfigSafe(cfg, 'storage-path');
+    baseLogPath = getStringConfigSafe(cfg, 'log-path');
+    maxStorage = getStringConfigSafe(cfg, 'max-storage');
+    if (checkConfig()) {
+      updateStatus(0);
+      recorder = new Recorder(storageDir, maxStorage);  
+      watchLogs(String(baseLogPath));
```
aza547, line 316: Why is String() needed here?

src/main/util.ts
```diff
@@ -451,6 +451,16 @@ const cutVideo = async (initialFile: string, finalDir: string, desiredDuration:
     });
 }
 
+/**
+ * Gets string value from the config in a more reliable manner.
+ * @param cfg the config store
+ * @param preference the preference
```
aza547, line 457: preference as a var name is a bit unfamiliar to me in this context, I was expecting `attribute` or something similar, but pleased to be educated if this is a common thing?

src/main/util.ts
```diff
@@ -451,6 +451,16 @@ const cutVideo = async (initialFile: string, finalDir: string, desiredDuration:
...
+ * @returns the string config
+ */
+const getStringConfigSafe = (cfg: any, preference: string) : string => {
+  return cfg.has(preference) ? cfg.get(preference) + "/" : "";
```
aza547, line 461: I hated writing the + "/" stuff. Even in early days it felt like a lazy choice :P 
Can we use path.join instead?

src/main/main.ts
ericlytle: leftover from a debugging session where i was getting an error on this. i doubt its needed now. good catch.

src/main/main.ts
ericlytle: yeah we can do that, i just copied this from other places in the code.

src/main/util.ts
ericlytle: well, i was thinking preference because it is a user pref they are setting. im fine with attribute too.

src/main/util.ts
```diff
@@ -452,6 +452,26 @@ const cutVideo = async (initialFile: string, finalDir: string, desiredDuration:
...
 }
 
+/**
+ * Gets string value from the config in a more reliable manner.
+ * @param cfg the config store
+ * @param key the key
+ * @returns the string config
+ */
+const getPathConfigSafe = (cfg: any, key: string) : string => {
+  return cfg.has(key) ? path.join(cfg.get(key), "/") : "";
```
aza547, line 462: As sorry I think I should have been clearer. I don't like that we have to add the "/" at all here, we should just use path methods everywhere when we are handling paths. I'm going to revert this so it can go in and I'll probably fix it another time.

src/main/main.ts
```diff
@@ -246,7 +241,21 @@ const openPathDialog = (event: any, args: any) => {
     console.log(err);
   })
 } 
+/**
+ * Checks the app config.
+ * @returns true if config is setup, false otherwise. 
+ */
+const checkConfig = () : boolean => {
+  return mainWindow !== null ? isConfigReady(cfg) && (getLatestLog(baseLogPath)) : false;
```
aza547, line 249: I think we wanted rid of the getLatestLog() check here?

Can we remove it? I'm not 100% sure, need to check if watchLogs will do the right thing which I think is nothing till a log is created then. 

A test case would be to move all your logs out the folder and see what happens when you drop one back in with the app running.

src/main/main.ts
```diff
@@ -291,10 +300,19 @@ ipcMain.on('settingsWindow', (event, args) => {
...
   if (settingsWindow === null) return; 
-
+  
   if (args[0] === "quit") {
     console.log("User closed settings");
     settingsWindow.close();
+    storageDir = getPathConfigSafe(cfg, 'storage-path');
+    baseLogPath = getPathConfigSafe(cfg, 'log-path');
+    maxStorage = getNumberConfigSafe(cfg, 'max-storage');
+    if (checkConfig()) {
```
aza547, line 310: I'm a stickler for a new line before a condition like this x)

src/main/main.ts
```diff
@@ -291,10 +300,19 @@ ipcMain.on('settingsWindow', (event, args) => {
...
+      updateStatus(0);
+      recorder = new Recorder(storageDir, maxStorage);
```
aza547, line 312: Can you merge main into this branch? I think when you do that this should have a 3rd 'monitor index' argument of the integer 0. Is stuff I added tonight.

src/main/util.ts
```diff
@@ -452,6 +452,26 @@ const cutVideo = async (initialFile: string, finalDir: string, desiredDuration:
...
+
+/**
+ * Gets number value from the config in a more reliable manner.
+ * @param cfg the config store
+ * @param preference the preference
+ * @returns the number config
+ */
+ const getNumberConfigSafe = (cfg: any, preference: string) : number => {
+    return cfg.has(preference) ? parseInt(cfg.get(preference)) : -1;
+  }
```
aza547, line 473: Indentation

src/main/util.ts
```diff
@@ -452,6 +452,26 @@ const cutVideo = async (initialFile: string, finalDir: string, desiredDuration:
...
+}
+
+/**
+ * Gets number value from the config in a more reliable manner.
+ * @param cfg the config store
+ * @param preference the preference
+ * @returns the number config
+ */
+ const getNumberConfigSafe = (cfg: any, preference: string) : number => {
+    return cfg.has(preference) ? parseInt(cfg.get(preference)) : -1;
```
aza547, line 472: Is there something better than -1 we could return here? Realise that for now this is fine, but if we add number config in the future that is allowed to be negative then this breaks. Is there a NaN equivalent in JS?

aza547 (commented):
Sorry, more picky comments.

src/main/main.ts
aza547: Also I think parenthesis could be neater as:

`return mainWindow !== null ? (isConfigReady(cfg) && getLatestLog(baseLogPath)) : false;`

src/main/util.ts
aza547: Reasoning is it's hard from a maintainability perspective if we need to remember if a path string has a "/" on the end or not. If we just use path.join() everywhere we don't need to worry. I just didn't know about path.join at the time of writing.

ericlytle: So this is a little tricky because if we dont have this the way it is and then we get from the store when there isnt anything there (first time user), this will result in a string of "/", which gets interpreted as a non-string and is the reason we were throwing an error before. the way i have this now resolves this and gives us a good way to get Paths. 

we can discuss alternate solutions if you dont liek this one.

src/main/main.ts
ericlytle: I am going to be removing all of those latest logs checks in a later PR when i work on the no logs in the directory issue, didnt want to clutter this reivvew. 

Also, i actually disagree on the parenthesis, only because if we ever get a linter it will throw a fit with unnecessary parenthesis. -- i know this because our work linter does this for this exact code lol

src/main/main.ts
ericlytle: agree on this. makes the code easier to read.

src/main/main.ts
ericlytle: really odd its not telling me thats a merge conflict

src/main/util.ts
ericlytle: NaN exists. i like it

src/main/main.ts
aza547: Hmm OK - in that case can remove the parenthesis around `(getLatestLog(baseLogPath))` ? 

Whatever, not that big a deal, not gonna let it block merging.

src/main/main.ts
aza547: yeah fuck knows why not - I tried to merge in main myself but it didn't work so I immediately gave up. We'll just deal with it if there is a problem.

aza547:
Nice. Thanks again. Only other thing is changelog, but I'll update that.

aza547 (approved):

Issue #43, opened 2022-08-20
Add Mythic+ recording support [Logs wanted]


aza547:
See https://github.com/aza547/wow-recorder/issues/6

aza547:
Lots of people are asking for this now. Pretty high on the priority list.

aza547:
It would help greatly if people could submit their M+ logs here, so I can build an allow-list of internal ZONE_CHANGE events that shouldn't stop the recording.

aza547:
Should consider adding a "boss only" mode, or having some recognition of boss encounters to save people searching long vods. We could watch the combat log for damage events vs a certain mob etc. There might even be actual boss pull events but wouldn't be suprised if not.

DThielke:
The boss events you're looking for are "ENCOUNTER_START" and "ENCOUNTER_END".

aza547:
https://github.com/Nnoggie/MythicDungeonTools/blob/bf3bd4b83c77df7425473927983ab7342982e720/MythicDungeonTools.lua#L2613

>MDT.zoneIdToDungeonIdx = {
  [934] = 15, --atal
  [935] = 15, --atal
  [936] = 16, --fh
  [1004] = 17, --kr
  [1039] = 18, --shrine
  [1040] = 18, --shrine
  [1162] = 19, --siege
  [1038] = 20, --temple
  [1043] = 20, --temple
  [1010] = 21, --motherlode
  [1041] = 22, --underrot
  [1042] = 22, --underrot
  [974] = 23, --toldagor
  [975] = 23, --toldagor
  [976] = 23, --toldagor
  [977] = 23, --toldagor
  [978] = 23, --toldagor
  [979] = 23, --toldagor
  [980] = 23, --toldagor
  [1015] = 24, --wcm
  [1016] = 24, --wcm
  [1017] = 24, --wcm
  [1018] = 24, --wcm
  [1029] = 24, --wcm
  [1490] = 25, --lower mecha
  [1491] = 26, --upper mecha
  [1493] = 26, --upper mecha
  [1494] = 26, --upper mecha
  [1497] = 26, --upper mecha
  [1663] = 30, --halls of atonement
  [1664] = 30, --halls of atonement
  [1665] = 30, --halls of atonement
  [1666] = 35, --necrotic wake
  [1667] = 35, --necrotic wake
  [1668] = 35, --necrotic wake
  [1669] = 31, --mists of tirna scithe
  [1674] = 32, --plaguefall
[...]

PR #84, opened 2022-09-05
Implement Mythic+ support

tdaugaard:
As discussed in issue #43, here's a preliminary implementation of Mythic+/Keystone support with time skips for trash/boss segments and what have we.

It's a draft for now, but I'm opening it to talk about the implementation/features/etc.


tdaugaard:
Be aware that I have no previous experience with React, so I might've done something in a way that isn't great - I'm not sure. I've googled things I wasn't sure of and such, so it should look somewhat okay.

aza547:
Thanks, I'll try review this today. Had a quick skim this morning. Thomas, you should DM be on discord and join our development channel if you plan to stay involved. Also good place to discuss things like this!

tdaugaard:
If you want me to split it up into several PRs, or if there are parts you don't want/like/have comments about, feel free to let me know - I'll sort it out however you'd like it :)

tdaugaard:
I've tested it pretty thoroughly with M+ dungeons and believe I've found and fixed all bugs related to that.

I've split some commits into more which might ease the review process.

I have NOT tested this in raid or PvP content and therefore don't know if any of the changes has created bugs for those, but I _don't_ think so. I've tried to keep the changes compatible with the current code.

aza547:
Having a look at this now - initial observation is there is no tests, can you check in an example dungeon log and add a dungeon test? See all_tests.py for existing infrastructure.

src/renderer/Layout.tsx
```diff
@@ -164,23 +186,6 @@ export default function Layout() {
     )
   };
 
-  /**
```
aza547, line 167: Awesome to get rid of this!

src/renderer/VideoButton.tsx
```diff
@@ -79,7 +88,7 @@ export default function VideoButton(props: any) {
   /**
   * Delete a video.
   */
-  function deleteVideo(filePath: string) {6
```
aza547, line 82: as if I checked this into main... :| 
this is why I need to write some unit tests

src/renderer/rendererutils.ts
```diff
@@ -26,24 +7,16 @@
 
     // Not sure how we can decide who won or lost yet. 
     // Combat log doesn't make it obvious.
-    if ((category == "Battlegrounds") || (category == "Solo Shuffle")) {
-        return "";
-    }
+    switch (category) {
+        case VideoCategory.MythicPlus:
+            return isGoodResult ? "Timed" : "Fail";
```
aza547, line 12: Deplete better than fail?

src/renderer/rendererutils.ts
```diff
@@ -26,24 +7,16 @@
 
     // Not sure how we can decide who won or lost yet. 
     // Combat log doesn't make it obvious.
-    if ((category == "Battlegrounds") || (category == "Solo Shuffle")) {
```
aza547, line 29: We've lost this logic as far as I can tell?

src/renderer/images.ts
```diff
@@ -27,25 +29,15 @@ for (const id of arenaIDs) {
...
+const dungeonIDs = Object.keys(dungeonsByZoneId).map(v=> parseInt(v, 10))
 
 let dungeon: imageObject = {};
 
 for (const id of dungeonIDs) {
-    dungeon[id] = require(`../../assets/dungeon/${id}.jpg`);
+    try {
+        dungeon[id] = require(`../../assets/dungeon/${id}.jpg`);
+    }
+    catch {};
```
aza547, line 40: probably need some logging in this branch

tdaugaard:
Will add tests - thanks for pointing that out :)

aza547:
Hi @tdaugaard, thanks for the PR, you've clearly put a lot of work into this. I'm really keen to get this PR into a state where I'm comfortable merging it. I think there is a bit of work to go still we're there, thoughts below. 

I've had a quick skim over as it's all I have time for tonight and made some light comments in line.

I've also run some tests on this branch and general feel is things seem pretty broken and I think you need to run the tests and do some debugging before further review. 

A few things I noticed in my testing:
- Arena test fails to get the player, spec and result. 
- Raid encounter test fails to finish and records forever. 
- I added a M+ test log and python script to main, and recording failed entirely. Pushed these tests to main, see [here](https://github.com/aza547/wow-recorder/commit/707b921dd294727e9d0605dd1e539f31db6f785a) and [here](https://github.com/aza547/wow-recorder/commit/0490bfa72ebb237d721cbca11310e6ad359d1d45).
- VideoButtons sizing have changed and looks bad.

You can find instructions on running tests [here](https://github.com/aza547/wow-recorder/blob/main/CONTRIBUTING.md#building-and-packaging).

aza547 (changes requested):

src/main/keystone.ts
```diff
@@ -0,0 +1,90 @@
...
+enum VideoSegmentType {
+    BossEncounter = 'Boss',
+    Trash = 'Trash'
+};
+
+class ChallengeModeVideoSegment {
+    logEnd: Date
+    result?: Boolean
+
+    constructor(
```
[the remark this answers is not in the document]
tdaugaard: Yeah, it's a short hand constructor that takes arguments and makes them as public members of the class as well. It's the same result as the following, in this particular case, but much more concise:

```typescript
class ChallengeModeVideoSegment {
    logEnd: Date;
    result?: Boolean;
    segmentType: VideoSegmentType;
    logStart: Date;
    ts: number;
    encounterId?: number;

    constructor(
        segmentType: VideoSegmentType,
        logStart: Date,
        ts: number,
        encounterId?: number
    ) {
        this.segmentType = segmentType;
        this.logStart = logStart;
        this.ts = ts;
        this.encounterId = encounterId;

        // Initially, let's set this to log start date to avoid logEnd
        // potentially being undefined.
        this.logEnd = logStart;
    }
};
```

src/main/logutils.ts
```diff
@@ -117,41 +123,173 @@ const tailFile = (path: string) => {
 }    
 
 /**
+ * Splits a WoW combat line intelligently with respect to quotes,
+ * lists, tuples, and what have we.
+ * 
+ */
+const splitLogLine = (line: string): any => {
```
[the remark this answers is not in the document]
tdaugaard: The reason it looks like C code (although it's a port from a Python script I made) is that it does something low level that JS cannot do by itself.

It basically parses the log file format _accurately_ and creates arrays as needed for things like affix lists, converts the tuples from talents and such into lists.

I understand the need to keep things simple, but parsing this log file isn't simple as such. Splitting on comma without regard to encounters that contain commas is a recipe for trouble.

Example is the boss name `Skulloc, Son of Gruul` - this previously cut it short to `Skulloc`, which isn't correct.

> I'm even more nervous if this can handle the COMBATANT_DATA (which are fairly complex lines) events we rely on for getting specs/teams in arena.

This is exactly what it can do. I'd be happy to implement some tests for this, if you want.

src/renderer/VideoButton.tsx
tdaugaard: I actually thought that _I_ had made a mistake and couldn't for the life of me figure out where, but accepted that it just needed to be fixed regardless. 🤣

src/renderer/rendererutils.ts
tdaugaard: Yes, that's what I was looking for! Thanks.

src/renderer/images.ts
tdaugaard: There was nothing before, as you can see, and I thought the logging would be pointless but I'd be happy to add as much `console.log` as you'd want me to :)

src/renderer/rendererutils.ts
tdaugaard: It's no longer neccessary. See `getResultText()`, default switch case takes care of that.

src/main/logutils.ts
```diff
@@ -117,41 +123,173 @@ const tailFile = (path: string) => {
...
+    let in_quote = false;
+    let list_items: string[][] = [];
+    let tuple_items: string[][] = [];
+    let in_what: string[] = [];
+    let args_list: any = [];
+    let value: any = '';
+
+    // eslint-disable-next-line no-plusplus
+    // Timestamp is always a fixed width of 18, and we want to skip that here.
```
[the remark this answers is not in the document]
tdaugaard: You're right, it's not true - but there's never more than 18 characters of date:

Log:

```
12/19 20:41:11.786  COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.1.5,PROJECT_ID,1
8/22 21:38:27.944  COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.2.7,PROJECT_ID,1
1/2 15:14:32.191  COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.1.5,PROJECT_ID,1
```

Input:

```js
console.log('12/19 20:41:11.786  COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.1.5,PROJECT_ID,1'.substring(18));
console.log('8/22 21:38:27.944  COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.2.7,PROJECT_ID,1'.substring(18));
console.log('1/2 15:14:32.191  COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.1.5,PROJECT_ID,1'.substring(18));
```

Output:

```
  COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.1.5,PROJECT_ID,1
 COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.2.7,PROJECT_ID,1
COMBAT_LOG_VERSION,19,ADVANCED_LOG_ENABLED,1,BUILD_VERSION,9.1.5,PROJECT_ID,1
```

src/main/logutils.ts
```diff
@@ -117,41 +123,173 @@ const tailFile = (path: string) => {
...
-    } else if (line.includes("COMBATANT_INFO")) {
-        handleCombatantInfoLine(line);
-    } else if (line.includes("SPELL_AURA_APPLIED")){
-        handleSpellAuraAppliedLine(line);
+    // Skip timestamp
+    let lineToken = line.substring(18);
+    // Get log line token
+    lineToken = lineToken.substring(0, lineToken.indexOf(','));
+
+    switch (lineToken) {
```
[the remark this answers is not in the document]
tdaugaard: switch..case is always faster than doing many if..else's, especially when the value to test is a simple string and doesn't have other conditions than that.

I've no strong feelings one way or the other regarding this, besides this being faster, performance wise.

src/main/keystone.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    // Remove the last video segment
+    removeLastSegment() {
+        this.videoSegments.pop();
+    }
+};
+
+const calculateCompletionResult = (mapId: number, duration: number): number => {
```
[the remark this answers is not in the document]
tdaugaard: Fixed in 40848d1

src/main/logutils.ts
```diff
@@ -205,35 +344,132 @@ const handleArenaStartLine = (line: string) => {
...
+
+    activeChallengeMode.addVideoSegment(new ChallengeModeVideoSegment(
+        VideoSegmentType.Trash, videoStartDate, 0
+    ));
+
+    console.debug("[ChallengeMode] Starting Challenge Mode dungeon")
+
+    metadata = {
+        name: lineArgs[1],
+        encounterID: parseInt(lineArgs[1], 10),
```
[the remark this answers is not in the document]
tdaugaard: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt describes that `radix` (the base) should not be assumed to be 10, and I'd rather not it be interpreted differently.

`radix` has a default, and in most cases it's 10, but it's up to the JS implementation to decide what the default is.

It's normally good practice to include it to make sure you get what you want.

src/main/logutils.ts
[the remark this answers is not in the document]
aza547: Ah, good point. Look forward to discussing further. You might be interested in taking a look at this: https://github.com/wowarenalogs/wowarenalogs/tree/main/packages/wow-combat-log-parser/src

src/main/logutils.ts
```diff
@@ -500,17 +752,16 @@ const pollWowProcess = () => {
 /**
  * getCombatLogDate
  */
-const getCombatLogDate = (line: string) => {
-    const [date, time] = line.split(" ");
-    const [month, day] = date.split("/");
-    const [hours, mins, secs] = time.split(":");
-
+const getCombatLogDate = (line: string): Date => {
```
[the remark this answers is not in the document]
tdaugaard: It's my habit of optimization, either for readability (for me, this is easier to read - believe it or not), or performance. I have no strong feelings one way or the other.

src/renderer/rendererutils.ts
tdaugaard: Fixed in 8a2d3cf

src/main/logutils.ts
```diff
@@ -117,41 +123,173 @@ const tailFile = (path: string) => {
...
-    } else if (line.includes("ENCOUNTER_END")) {
-        handleRaidStopLine(line);
-    } else if (line.includes("ZONE_CHANGE")) {
-        handleZoneChange(line);
-    } else if (line.includes("COMBATANT_INFO")) {
-        handleCombatantInfoLine(line);
-    } else if (line.includes("SPELL_AURA_APPLIED")){
-        handleSpellAuraAppliedLine(line);
+    // Skip timestamp
+    let lineToken = line.substring(18);
```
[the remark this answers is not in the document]
tdaugaard: See https://github.com/aza547/wow-recorder/pull/84#discussion_r964173053 for comment about this.

src/renderer/VideoButton.tsx
aza547: hilariously it somehow isn't broken... no idea why

src/main/logutils.ts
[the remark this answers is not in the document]
aza547: ah - good point

src/main/keystone.ts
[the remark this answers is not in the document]
aza547: neat - thanks for the education!

src/renderer/images.ts
aza547: That'd be great - thanks!

src/main/logutils.ts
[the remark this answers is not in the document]
tdaugaard: That looks very overengineered, considering it's a partial, heavily PvP centric/specific parser, at a glance.

src/main/logutils.ts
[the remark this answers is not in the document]
tdaugaard: I've clarified/corrected the comment to accurately describe the situation, that is, a max length rather than a fixed length in 92821f4

src/renderer/VideoButton.tsx
tdaugaard: Just javascript things, really. Javascript will take what you write and do the best it can, mostly ignoring things that doesn't make sense. 🤷‍♂️🤣

src/renderer/images.ts
tdaugaard: Fixed in 372e3c6

tdaugaard:
> Arena test fails to get the player, spec and result.

Fixed in b0710fa - I was parsing values starting with `0x` wrongly due to parseInt(..., 10) - needs to be parseInt(..., 16) for hexadecimal values.

tdaugaard:
> VideoButtons sizing have changed and looks bad.

Shouldn't have changed. Screenshot from the "Test" button test of an Arena match:

image

aza547:
> > VideoButtons sizing have changed and looks bad.
> 
> Shouldn't have changed. Screenshot from the "Test" button test of an Arena match:
> 
> image

But this is how it looked before:

image

tdaugaard:
I had to overlay mine on yours to see the difference, but I see it now - there are some slight font size differences and the background image looks off (due to `background: contain;`). I will fix that :)

tdaugaard:
BTW, I _can_ fix this entire PR to _not_ include the changes to `splitLogLine`, `getCombatLogDate`, and `handleLogLine` and make a separate PR for those things.

They are not necessary for this feature, but I'd like for at least the changes to `splitLogLine` and `handleLogLine` to stay, as they do improve the performance and accuracy significantly.

Specifically, using `.split()` every time you want a value from the line is very ineffcient, as you're splitting every single time for a single value.

Using `.contains()` on multiple lines is quite a performance hit, as we have to test every `if` expression for lines we're not interested in, which is _most_ of the log lines.

I have a fix for that, too, using an array of interesting events, then doing the  `switch (line)` on only the log event type - but that's probably better suited for a different PR altogether, too.

Let me know your thoughts as well as whether you see this whole feature as I've implemented it as likely to be merged in either state as explained above.

aza547:
I don't think I mind this all being in one PR and agree that lots of stuff is inefficient. Convinced by your points regarding the splitLogLine function too.  

What's the current state of the code? Have you ran the tests? Should I re-review?

tdaugaard:
I'm not done with the tests yet, and I still have to test the issue with the raid recording not stopping.

I'll let you know when it's ready to re-review.

One final question, I am unable to reproduce the font size difference on the VideoButton component. I can _see_ there is a difference in some things, but there are currently (after the last commit) no changes to `App.css` that isn't scoped to the `.dungeon` class, which is only applied to VideoButton's from M+ dungeons. Perhaps I'm missing a font or something since the difference is not in size, but width.

Two Arena buttons overlaid in "Difference" mode:

image

aza547:
> I'm not done with the tests yet, and I still have to test the issue with the raid recording not stopping.
> 
> I'll let you know when it's ready to re-review.
> 
> One final question, I am unable to reproduce the font size difference on the VideoButton component. I can _see_ there is a difference in some things, but there are currently (after the last commit) no changes of mine to `App.css` that isn't scoped to the `.dungeon` class, which is only applied to VideoButton's from M+ dungeons. Perhaps I'm missing a font or something since the difference is not in size, but width.
> 
> Two Arena buttons overlaid in "Difference" mode:
> 
> image

Strange, I'll keep this in mind when reviewing. Maybe I'll spot the cause.

aza547:
Something to consider, the test script timestamps are off because it only records for a few seconds and obviously the log used in the test spans an entire dungeon duration. In other tests I've hacked the timestamps to give better test results.

You should consider that here. Obviously makes testing slightly less representative of reality but it can hit the features then rather than trying to seek to 40 min in a 10 sec vid.  

For example, see the first line [here](https://raw.githubusercontent.com/aza547/wow-recorder/main/resources/example-logs/raid.txt). 

Suspect to make the seeking work in the test you might need to carefully hack several time stamps throughout the log? Or apply some clever interpolation programmatically to them all?

tdaugaard:
> Suspect to make the seeking work in the test you might need to carefully hack several time stamps throughout the log? Or apply some clever interpolation programmatically to them all?

I had considered using wall clock for the time stamps instead of the log timestamps, as they might be more accurate and also would work fine with testing, but I'd have to test that first in real-time. Will probably get an opportunity to do so tomorrow.

tdaugaard:
Thoroughly tested this evening and it works super well, including the time stamps - they are not off by more than 500-1200 ms.
image

tdaugaard:
Closing because it has been replaced by PR #111

PR #87, opened 2022-09-07
Fix memory leak from EventEmitter

### LINKED ISSUE #86: Possible EventEmitter memory leak
### Problem as logged

```
(node:14384) MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
11 refreshState listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
```
### Code

The only place that's attaching a listener for `refreshState` is in [`src/main/renderer/Layout.tsx`, line 146](https://github.com/aza547/wow-recorder/blob/0490bfa72ebb237d721cbca11310e6ad359d1d45/src/renderer/Layout.tsx#L146) 

### How to reproduce

Have two videos in one category, then repeatedly switch between them some 11+ times.

tdaugaard:
Fixes #86.

The listener on `ipc.on('refreshState')` was attached on every re-render, and this commit fixes that by making sure the listener only attached once per component load/unload.

This caused memory leaks as logged in the console:

```
(node:14384) MaxListenersExceededWarning: Possible EventEmitter memory leak detected.
11 refreshState listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
```

`React.useEffect()` triggers on every state update, but when given an empty array as its 2nd argument, it changes that to only trigger on component load/unload.

The scenario is described in the React documentation here:

https://www.robinwieruch.de/react-function-component/#react-function-component-lifecycle

under section "React Function Component: Lifecycle".

### DIFF
--- a/src/renderer/Layout.tsx
+++ b/src/renderer/Layout.tsx
@@ -140,17 +140,28 @@ export default function Layout() {
   */
   const { classes: styles } = useStyles();
 
-  /**
-   * Refresh handler.
-   */
-  ipc.on('refreshState', () => {
-    setState(prevState => {
-      return {
-        ...prevState, 
-        videoState: ipc.sendSync('getVideoState', categories)
-      }
-    })
-  });
+  // This is effectively equivalent to componentDidMount() in
+  // React Component classes
+  React.useEffect(() => {
+      /**
+       * Refresh handler.
+       */
+      ipc.on('refreshState', () => {
+        setState(prevState => {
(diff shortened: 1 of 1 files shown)


aza547:
Awesome. Thanks for fixing this. I had previously raised an issue for this but closed it in error.

aza547 (approved):

tdaugaard:
Sorry, completely forgot the changelog 🙄

PR #88, opened 2022-09-08
Fix EventEmitter memory leak in settings

tdaugaard:
In the Settings window, if "**Run on Startup?**" is repeatedly clicked or if some other action causes redraw of the component, it would attach an event listener for `'settingsWindow'` on every redraw.

This caused a potential memory leak which this commit fixes.

Related to issue #86


aza547 (approved):
Cheers!

PR #89, opened 2022-09-08
Feature: Select audio device

### LINKED ISSUE #17: mic recording may default to webcam
Mori noticed this while testing, not the bug of the century but probably should make it configurable.

tdaugaard:
Implement a way to seelct input and output devices for recording via Settings. Fixes #17

### DIFF
--- a/src/main/obsRecorder.ts
+++ b/src/main/obsRecorder.ts
@@ -182,6 +182,7 @@ const setupSources = (scene: any) => {
           setSetting('Output', `Track${currentTrack}Name`, device.name);
           source.audioMixers = 1 | (1 << currentTrack-1); // Bit mask to output to only tracks 1 and current track
           osn.Global.setOutputSource(currentTrack, source);
+          source.release()
           currentTrack++;
         });
     }
@@ -194,6 +195,7 @@ const setupSources = (scene: any) => {
           setSetting('Output', `Track${currentTrack}Name`, device.name);
           source.audioMixers = 1 | (1 << currentTrack-1); // Bit mask to output to only tracks 1 and current track
           osn.Global.setOutputSource(currentTrack, source);
+          source.release()
           currentTrack++;
         });
     }
--- a/src/main/obsRecorder.ts
+++ b/src/main/obsRecorder.ts
@@ -1,7 +1,6 @@
 import { fixPathWhenPackaged } from "./util";
 import WaitQueue from 'wait-queue';
 import { getAvailableAudioInputDevices, getAvailableAudioOutputDevices } from "./obsAudioDeviceUtils";
-import Store from 'electron-store';
 const waitQueue = new WaitQueue<any>();
 const path = require('path');
 const { byOS, OS } = require('./operatingSystems');
@@ -14,16 +13,16 @@ let scene = null;
 /*
 * Reconfigure the recorder without destroying it.
 */
-const reconfigure = (outputPath: string, monitorIndex: number) => {
+const reconfigure = (outputPath: string, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string) => {
   configureOBS(outputPath);
   scene = setupScene(monitorIndex);
-  setupSources(scene);
+  setupSources(scene, audioInputDeviceId, audioOutputDeviceId);
 }
 
 /*
 * Init the library, launch OBS Studio instance, configure it, set up sources and scene
 */
-const initialize = (outputPath: string, monitorIndex: number) => {
+const initialize = (outputPath: string, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string) => {
   if (obsInitialized) {
     console.warn("OBS is already initialized");
     return;
@@ -32,7 +31,7 @@ const initialize = (outputPath: string, monitorIndex: number) => {
   initOBS();
   configureOBS(outputPath);
   scene = setupScene(monitorIndex);
-  setupSources(scene);
+  setupSources(scene, audioInputDeviceId, audioOutputDeviceId);
   obsInitialized = true;
 }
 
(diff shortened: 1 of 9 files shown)


tdaugaard:
I'm unsure exactly how the device scenes work - does it actually capture audio from _all_ devices as it is now, or does it just select one device out of however many there is?

I created two extra options besides the actual audio devices: "All" and "None", but as I'm unsure what it does exactly, maybe only "None" + devices are required. Let me know, if you know :)

aza547:
Looks great as usual. You should increase the height of the settingsWindowin main.ts so that it all fits. 

> I'm unsure exactly how the device scenes work - does it actually capture audio from _all_ devices as it is now, or does it just select one device out of however many there is?
> 


> I created two extra options besides the actual audio devices: "All" and "None", but as I'm unsure what it does exactly, maybe only "None" + devices are required. Let me know, if you know :)

Not sure of the top of my head how this works, I'll have a look when I get some time and get back to you.

aza547:
I don't really understand why I can't see obsRecorder.reconfigure() here. Have you branched off an old version of main? I think you need to ensure that function works (think you will need to plumb some config through as args).

src/main/obsRecorder.ts
```diff
@@ -158,44 +160,45 @@ const setupScene = (monitorIndex: number) => {
...
-  dummyDevice.release();
-  return devices;
-};
-
 /*
 * setupSources
 */
 const setupSources = (scene: any) => {
+  const cfg = new Store();
+  const selectedInputAudioDevice = (cfg.get('audio-input-device', 'all') as string);
```
aza547, line 168: The existing model is to keep track of config in main.ts and pass values through to the Recorder class as args. I think this makes reconfiguring easier.

tdaugaard: Honestly, I hadn't even noticed that `cfg.get()` was exposed via an IPC channel. 🤦‍♀️ I will fix, of course :)

tdaugaard: Fixed to be in line with the existing way of passing configuration to the `Recorder`.

tdaugaard:
I have fixed the issue you addressed here. I was contemplating creating an object/type to contain and pass the settings to the recorder through, to avoid having to pass 4 arguments everywhere, but I held back and kept to the actual feature 😂

aza547:
Looks good.

aza547:
Have you tested? Particularly interested in the reconfigure function.

aza547:
Had a quick go at testing. Didn't get any sound in the video. My config below for ref: 

image

Also some errors in react:

image

tdaugaard:
> Had a quick go at testing. Didn't get any sound in the video. My config below for ref:

I was testing it quite a lot yesterday with no audio, only mic, no output device, other way around, different mic (BT headset) etc. and it worked fine exactly as intended.

I did not see the errors in the console - I tend to close the inspector tab because it takes up too much space when testing 😬 I'm fixing those now.

tdaugaard:
> You should increase the height of the settingsWindow in main.ts so that it all fits. This page is probably due a revamp at some stage to be better setup.

Definitely, but it'll do for now no problem. I changed the size to 650x460 to allow two columns instead of having it become very tall. If you prefer a tall window, I'll make it tall =)

aza547:
> > You should increase the height of the settingsWindow in main.ts so that it all fits. This page is probably due a revamp at some stage to be better setup.
> 
> Definitely, but it'll do for now no problem. I changed the size to 650x460 to allow two columns instead of having it become very tall. If you prefer a tall window, I'll make it tall =)

Wide is good too :)

tdaugaard:
I was able to reproduce the problem you mentioned about no output audio being recorded. I'll investigate and test further.

tdaugaard:
I had mixed up the input/output device selection _gravely_ which affected the feature in a way that made it never record any output audio. This has been fixed.

There's a small bug in the settings window where it won't show the selected option after selecting it, but it'll save it properly and reload it as intended when the settings window is opened once again.

I'll investigte this and fix it, but that should be the last thing.

aza547:
Had a test, got sound which was nice!

Had a problem where I did a recording with sound, then changed settings to No input / No output.

Recorded again and still got sound.

tdaugaard:
> Had a problem where I did a recording with sound, then changed settings to No input / No output.

I am able to reproduce this - am looking into it.

tdaugaard:
If I select a device, record, then I get audio. Then select None, and I still get audio. Then select a device, THEN select None, then Update and now I get no sound. Interesting.

tdaugaard:
It was due to the way audio tracks are added; it's like the old tracks aren't removed when `setupSources()` runs subsequently. 

Temporily fixed by using `source.muted` flag to mute the sources that we haven't selected, but we should find a way to clear the sources before configuring them again.

tdaugaard:
After some more testing, it _is_ possible to simply delete existing sources first:

```typescript
for (let i = 0; i < 12; i++) {
  osn.Global.setOutputSource(i, null);
}
```

But this has the downside that we need to know how many sources there were, since `osn.Global` doesn't expose that, as far as I can tell.

I'll let this one stay as it is, because it works.

aza547:
This is a really awesome feature. I didn't find any bugs in the latest test. Nice work.

src/main/util.ts
```diff
@@ -492,13 +493,13 @@ const cutVideo = async (initialFile: string, finalDir: string, desiredDuration:
...
- * Gets string value from the config in a more reliable manner.
+ * Gets a path (string) value from the config in a more reliable manner.
  * @param cfg the config store
  * @param key the key
  * @returns the string config
  */
-const getPathConfigSafe = (cfg: any, key: string): string => {
-    return cfg.has(key) ? path.join(cfg.get(key), "/") : "";
+const getPathConfigSafe = (cfg: ElectronStore, key: string): string => {
+    return cfg.has(key) ? path.join(getStringConfigSafe(cfg, key), path.sep) : "";
```
aza547, line 502: path.sep, nice!

src/main/operatingSystems.ts
```diff
@@ -5,7 +5,7 @@ const OS = {
     Mac: 'darwin',
   }
   
-  function byOS(handlers) {
+  function byOS(handlers: any) {
```
aza547, line 8: Pretty sure this is all useless code FYI, inherited from the obsRecorder copy paste.

tdaugaard: Oh, yeah it's there to basically support MacOS as well from what I can gather. If you don't plan on supporting that, we can remove it - if you want to keep the support in (for that part of the code anyway) then we should keep it.

PR #90, opened 2022-09-09
Fix sizeMonitor() blocking the thread

### LINKED ISSUE #75: Size monitor shouldn't block saving
App responsiveness thing. Currently size monitor will prevent the app returning to a ready state after a video has just recorded. We'll stick on the "saving" icon until finished.

That doesn't make sense, we can run the size monitor whenever and it shouldn't block a user watching a video back in the GUI.

tdaugaard:
- Replace `glob` with `glob-promise` to be able to perform async globbing.

- Import `fs-promise` as `fs`and rename `fs` to `fscb` as `fs-promise` cannot be renamed on its import, as far as I know.

- Create async `getSortedVideos()` to find and return all video files sorted by modification time. This method will read all metadata for every video which by itself is a performance penalty, but considering modern hardware it should be neglible and not noticable.

- Remove `isVideoProtected()` as `getSortedVideos()` replace this.

- Create async `writeMetadataFile()` to reduce duplicate code.

- Refactor `deleteOldestVideo()` and `getNewestVideo()` to be async.

- Create `saveMetadataForVideo()`.

- Refactor `toggleVideoProtected()` to use `getMetadataForVideo()`/`saveMetadataForVideo()`.

This fixes #75

### DIFF
--- a/src/main/recorder.ts
+++ b/src/main/recorder.ts
@@ -213,8 +213,8 @@ const glob = require('glob');
             // finish up with the video file. Maybe this can be done better. 
             setTimeout(async () => {
                 const bufferedVideo = await getNewestVideo(this._bufferStorageDir);
-                await cutVideo(bufferedVideo, this._storageDir, metadata.duration);
-                writeMetadataFile(this._storageDir, metadata).then(resolve);
+                const videoPath = await cutVideo(bufferedVideo, this._storageDir, metadata.duration);
+                writeMetadataFile(videoPath, metadata).then(resolve);
             }, 
             2000)
         });   
--- a/src/main/recorder.ts
+++ b/src/main/recorder.ts
@@ -255,7 +255,8 @@ const glob = require('glob');
     /**
      * Reconfigure the underlying obsRecorder. 
      */
-    reconfigure = (outputPath: string, monitorIndex: number) => {      
+    reconfigure = (outputPath: string, maxStorage: number, monitorIndex: number) => {
+        this._maxStorage = maxStorage;
 
         if (this._isRecording) {
             obsRecorder.stop();       
--- a/src/main/recorder.ts
+++ b/src/main/recorder.ts
@@ -181,7 +181,10 @@ const glob = require('glob');
 
             // Run the size monitor to ensure we stay within size limit.
             // Need some maths to convert GB to bytes
-            runSizeMonitor(this._storageDir, this._maxStorage); 
+            runSizeMonitor(this._storageDir, this._maxStorage)
+                .then(() => {
+                    if (mainWindow) mainWindow.webContents.send('refreshState');
+                });
 
             // Clean-up the temporary recording directory. 
             this.cleanupBuffer();
--- a/src/main/recorder.ts
+++ b/src/main/recorder.ts
@@ -209,10 +209,9 @@ const glob = require('glob');
             // It's a bit hacky that we async wait for 2 seconds for OBS to 
             // finish up with the video file. Maybe this can be done better. 
             setTimeout(async () => {
-                const bufferedVideo = getNewestVideo(this._bufferStorageDir);
+                const bufferedVideo = await getNewestVideo(this._bufferStorageDir);
                 await cutVideo(bufferedVideo, this._storageDir, metadata.duration);
-                writeMetadataFile(this._storageDir, metadata);  
-                resolve();       
+                writeMetadataFile(this._storageDir, metadata).then(resolve);
             }, 
             2000)
         });   
--- a/src/main/util.ts
+++ b/src/main/util.ts
@@ -344,7 +344,7 @@ const runSizeMonitor = async (storageDir: string, maxStorageGB: number): Promise
         return files.shift().name;
      }
 
-     return Promise.reject('no files');
+     return Promise.reject('No video files were found while looking for the most recent video.');
 }  
 
 /**
--- a/src/main/util.ts
+++ b/src/main/util.ts
@@ -18,8 +18,9 @@ const ffprobePath = fixPathWhenPackaged(require('@ffprobe-installer/ffprobe').pa
 const ffmpeg = require('fluent-ffmpeg');
 ffmpeg.setFfmpegPath(ffmpegPath);
 ffmpeg.setFfprobePath(ffprobePath);
-const fs = require('fs');
-const glob = require('glob');
+import fs from 'fs/promises';
+const fscb = require('fs')
+import glob from 'glob-promise';
 
 let videoIndex: { [category: string]: number }  = {};
 
@@ -55,7 +56,7 @@ const getEmptyState = () => {
  */
 const loadAllVideos = (storageDir: any, videoState: any) => {
     const videos = glob.sync(storageDir + "*.mp4")        
-        .map((name: any) => ({name, mtime: fs.statSync(name).mtime}))
+        .map((name: any) => ({name, mtime: fscb.statSync(name).mtime}))
         .sort((A: any, B: any) => B.mtime - A.mtime);
 
     for (const category of categories) {
@@ -72,7 +73,7 @@ const loadAllVideos = (storageDir: any, videoState: any) => {
  */
  const loadVideoDetails = (video: any, videoState: any) => {
     const today = new Date();
-    const videoDate = new Date(fs.statSync(video.name).mtime)
+    const videoDate = new Date(fscb.statSync(video.name).mtime)
     const isVideoFromToday = (today.toDateString() === videoDate.toDateString());
 
     const metadata = getMetadataForVideo(video)
@@ -105,13 +106,11 @@ const loadAllVideos = (storageDir: any, videoState: any) => {
 /**
  * Get the date a video was recorded from the date object.
  */
-const getMetadataForVideo = (video: any) => {
-    const videoFileName = path.basename(video.name, '.mp4');
-    const videoDirName = path.dirname(video.name);
-    const metadataFile = videoDirName + "/" + videoFileName + ".json";
+const getMetadataForVideo = (video: string) => {
+    const metadataFile = getMetadataFileForVideo(video)
 
-    if (fs.existsSync(metadataFile)) {
-        const metadataJSON = fs.readFileSync(metadataFile);
+    if (fscb.existsSync(metadataFile)) {
+        const metadataJSON = fscb.readFileSync(metadataFile);
         const metadata = JSON.parse(metadataJSON);
         return metadata;
     } else {
@@ -120,14 +119,21 @@ const getMetadataForVideo = (video: any) => {
     }
 }
 
+const saveMetadataForVideo = (videoPath: any, metadata: any) => {
+    const metadataFile = getMetadataFileForVideo(videoPath);
+
+    const newMetadataJsonString = JSON.stringify(metadata, null, 2);
+    fscb.writeFileSync(metadataFile, newMetadataJsonString);
+}
+
 /**
- * Get the date a video was recorded from the date object.
+ * Get the filename for the metadata file associated with the given video file
  */
- const getMetadataFileForVideo = (video: any) => {
+ const getMetadataFileForVideo = (video: string) => {
     const videoFileName = path.basename(video, '.mp4');
     const videoDirName = path.dirname(video);
-    const metadataFilePath = videoDirName + "/" + videoFileName + ".json";
-    return metadataFilePath;
+
+    return path.join(videoDirName, videoFileName + '.json');
 }
 
 /**
@@ -247,13 +253,14 @@ const getVideoState = (storageDir: unknown) => {
 }    
 
 /**
- *  writeMetadataFile
+ *  Writes video metadata asynchronously and returns a Promise
  */
-const writeMetadataFile = (storageDir: string, metadata: Metadata) => {
+const writeMetadataFile = async (storageDir: string, metadata: Metadata) => {
+    const file = await getNewestVideo(storageDir);
     const jsonString = JSON.stringify(metadata, null, 2);
-    const newestVideoPath = getNewestVideo(storageDir);
-    const newestVideoName = path.basename(newestVideoPath, '.mp4');
-    fs.writeFileSync(storageDir + newestVideoName + ".json", jsonString);
+    const metadataFileName = getMetadataFileForVideo(file);
+
+    return await fs.writeFile(metadataFileName, jsonString);
 }    
 
 /**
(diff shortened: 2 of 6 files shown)


tdaugaard:
I haven't tested it yet, as I don't have WoW on this PC, but I will test it tomorrow. Async functionality for `getSortedVideso()` has been tested separately on `src/main/*.ts` files, so I know that works at least - hence why this is a draft.

src/main/util.ts
```diff
@@ -12,14 +12,21 @@ const chalk = require('chalk');
...
 const { exec } = require('child_process');
 const ffmpegPath = fixPathWhenPackaged(require('@ffmpeg-installer/ffmpeg').path);
 const ffprobePath = fixPathWhenPackaged(require('@ffprobe-installer/ffprobe').path);
 const ffmpeg = require('fluent-ffmpeg');
 ffmpeg.setFfmpegPath(ffmpegPath);
 ffmpeg.setFfprobePath(ffprobePath);
-const fs = require('fs');
-const glob = require('glob');
+import fs from 'fs/promises';
+const fscb = require('fs')
```
aza547, line 28: why `fscb`  and not `fsSync` or so? What does cb mean?

src/main/util.ts
```diff
@@ -120,14 +125,21 @@ const getMetadataForVideo = (video: any) => {
     }
 }
 
+const saveMetadataForVideo = (videoPath: string, metadata: any) => {
+    const metadataFile = getMetadataFileForVideo(videoPath);
+
```
aza547, line 130: random new line

src/main/util.ts
```diff
@@ -247,74 +259,98 @@ const getVideoState = (storageDir: unknown) => {
...
-    }
+    return { name: videoPath, size, mtime };
+};
 
-    if (totalSize > maxStorageBytes) { 
-        deleteOldestVideo(storageDir);
-        runSizeMonitor(storageDir, maxStorageBytes);
-    } 
-}   
+const getSortedVideos = async (storageDir: string): Promise<VideoInfo[]> => {
```
aza547, line 283: Worth a comment to say which sorted order this is, can work it out  from the func but it's a not totally obvious.

src/main/util.ts
```diff
@@ -247,74 +259,98 @@ const getVideoState = (storageDir: unknown) => {
...
+            const metadataFileName = getMetadataFileForVideo(file.name);
+            const data = fscb.readFileSync(metadataFileName);
+            const metadata = JSON.parse(data.toString());
+            return { ...file, metadata, };
+        } catch (e) {
+            console.log(`Unable read and/or parse JSON from metadata file: ${file.name}`);
+        }
+    });
+
+    let totalVideoFileSize = 0;
```
aza547, line 311: this can probably move down a few lines

src/main/util.ts
tdaugaard: `fscb` was just chosen because the `fs` module can take a callback (cb) to the non-Sync versions of the fs-calls, like `readdirSync` has a corrosponding `readdir` method, that uses a callback to deliver the results.

So, in turn 'fs' isn't inherently synchronous, onlny the `*Sync` versions of the methods, and that's why I picked that name.

src/main/util.ts
tdaugaard: It gets the filename and path for the metadata for the given video fil in order to save the metadata. Before, a few places that saved or read metadata would first get filename, the call the appropriate function to save/load it.

Now, the method that saves/loads will determine the filename based on the video path. See also `getMetadataForVideo()`, `writeMetadataFile()`, and `deleteVideo()` where it's used as well.

aza547:
Found a mainline bug with `getSortedVideos()`:

image

aza547:
Didn't see any examples of `await glob` here: https://www.npmjs.com/package/glob-promise. I wonder if that's valid syntax?

aza547:
Actually should this perhaps just be `glob.glob()`?

image

aza547:
Seems not... 

```
glob(pattern [, options])
Alias for glob.promise
```

aza547:
I'm stumped. Can't tell what is wrong but the new glob module doesn't seem to return any files, it's the same in other uses too. The app doesn't load any videos anymore. Any ideas?

aza547:
Maybe easier to go back the previous glob module and wrap it ourselves? Either manually or with https://github.com/isaacs/node-glob/issues/228#issuecomment-375170162

aza547:
Another theory that this might be something to do with the fs module change.

tdaugaard:
I can tell you that it works fine here. I have no problem removing the `fs/promise` module and going back to callback based logic, but it looks worse and it's harder to catch errors with `.catch()` if need be.

Let me test a bit more locally and see if I can reproduce this in any way - it looks like it's in the cleanup phase where it removes the "buffer videos", correct?

> Didn't see any examples of `await glob`

`await` is always valid when something returns a promise, as long as it's used in an `async` function/method, so the documentation doesn't need to say that it's valid, because it always is.

aza547:
We debugged this and the reason for the differences was Node 14 vs Node 16

tdaugaard:
It works for me now on Node v16.

aza547:
> It works for me now on Node v16. I had to remove `node_modules` and reinstall them.

Still no luck here, trying a few things...

aza547:
No luck, still same symptoms. Have deleted node_modules and reinstalled, no change.

aza547:
> No luck, still same symptoms. Have deleted node_modules and reinstalled, no change.

Seems it's glob 8.x.x. If I change package.json to 7.2.0 (seems to be what we use on main), it's all fine...

tdaugaard:
This is perhaps because `glob` wasn't added to the project as a dependency, but `glob` was available because it was a peer dependency from several other packages, which coincidentally was `glob` 7.1/7.2. I can add that specifically with that version.

aza547:
> This is perhaps because `glob` wasn't added to the project as a dependency, but `glob` was available because it was a peer dependency from several other packages, which coincidentally was `glob` 7.1/7.2. I can add that specifically with that version.

Strange if glob 8.x.x worked for you but not me? Was that definitely the version you had?

aza547:
Also wonder if the promises version of this PR would have worked with glob 7?

tdaugaard:
> Strange if glob 8.x.x worked for you but not me? Was that definitely the version you had?

Unfortunately, I can't tell now since I deleted my `node_modules` but it works with ^7.1.0 at least. It's okay, I don't think we actually made use of the promise version of glob, only `fs/promise` was important.

tdaugaard:
Okay, tested and didn't work with line 291. Reintroduced a promise version of `glob` (using `util.promisify()`) and now it works - for me.

aza547:
Seems good to me. 

Two minor things I saw testing:
- Seems to not pick up new maxStorage config without an app restart.
- We don't log if we fail to delete a video. See deleteVideo - can you add one?

Could you consider what is needed to fix that? Steps to reproduce:
- Reduce size monitor config.
- Run a test without restarting app.
- Observe old config still in use e.g. `[Size Monitor] Running (max size = 4 GB) ` 

I'll do a final pass over the diff tomorrow then we can merge?

tdaugaard:
> Seems to not pick up new maxStorage config without an app restart.

`recorder` class was missing a `maxStorage` parameter for `reconfigure()`.

New result now, attempting to reproduce it:

```
(...)
23:08:32.458 > Recorder: Start recording buffer
23:08:32.459 > obsRecorder: start
23:08:32.464 > [Size Monitor] Running (max size = 100 GB)
23:08:32.609 > Asserted OBS signal: recording start
23:08:37.112 > User clicked open settings
(...)
23:08:46.457 > Setting in config store:  max-storage 120
(...)
23:09:05.494 > Recorder: Start recording buffer
23:09:05.495 > obsRecorder: start
23:09:05.501 > [Size Monitor] Running (max size = 120 GB)
23:09:05.635 > Asserted OBS signal: recording start

```
> We don't log if we fail to delete a video. See deleteVideo - can you add one?

In `tryUnlinkSync()`, line 374, there's logging when a file can't be deleted. I've tested it. Does it not work for you?

aza547:
> > Seems to not pick up new maxStorage config without an app restart.
> 
> `recorder` class was missing a `maxStorage` parameter for `reconfigure()`.
> 
> New result now, attempting to reproduce it:
> 
> ```
> (...)
> 23:08:32.458 > Recorder: Start recording buffer
> 23:08:32.459 > obsRecorder: start
> 23:08:32.464 > [Size Monitor] Running (max size = 100 GB)
> 23:08:32.609 > Asserted OBS signal: recording start
> 23:08:37.112 > User clicked open settings
> (...)
> 23:08:46.457 > Setting in config store:  max-storage 120
> (...)
> 23:09:05.494 > Recorder: Start recording buffer
> 23:09:05.495 > obsRecorder: start
> 23:09:05.501 > [Size Monitor] Running (max size = 120 GB)
> 23:09:05.635 > Asserted OBS signal: recording start
> ```
> 
> > We don't log if we fail to delete a video. See deleteVideo - can you add one?
> 
> In `util.ts`, `tryUnlinkSync()` (line 374), there's logging when a file can't be deleted. I've tested it. Does it not work for you?

Ah so there is. I must have misunderstood something, I expect it works.

src/main/recorder.ts
```diff
@@ -209,10 +212,9 @@ const glob = require('glob');
             // It's a bit hacky that we async wait for 2 seconds for OBS to 
             // finish up with the video file. Maybe this can be done better. 
             setTimeout(async () => {
-                const bufferedVideo = getNewestVideo(this._bufferStorageDir);
+                const bufferedVideo = await getNewestVideo(this._bufferStorageDir);
                 await cutVideo(bufferedVideo, this._storageDir, metadata.duration);
-                writeMetadataFile(this._storageDir, metadata);  
-                resolve();
```
aza547, line 215: We've lost the call to resolve here. Don't we need that? writeMetadataFile now returns a resolved promise, is that enough to resolve the promise on line 207? I'm new to this stuff - so keen to understand!

src/main/util.ts
```diff
@@ -105,10 +114,8 @@ const loadAllVideos = (storageDir: any, videoState: any) => {
 /**
  * Get the date a video was recorded from the date object.
  */
-const getMetadataForVideo = (video: any) => {
-    const videoFileName = path.basename(video.name, '.mp4');
-    const videoDirName = path.dirname(video.name);
-    const metadataFile = videoDirName + "/" + videoFileName + ".json";
+const getMetadataForVideo = (video: string) => {
+    const metadataFile = getMetadataFileForVideo(video)
```
aza547, line 118: My VSCode is upset with line 122 here, not sure if it's as a result of this change:
`Argument of type 'Buffer' is not assignable to parameter of type 'string'`

aza547: I think the fix is simply to add `.toString()` on the end. Would you sort that as you're in the area? I think it's pre-existing.

aza547: `const metadata = JSON.parse(metadataJSON.toString());` 

https://stackoverflow.com/questions/46190511/typescript-error-ts2345-error-ts2345argument-of-type-buffer-is-not-assignabl

src/main/util.ts
```diff
@@ -247,74 +264,101 @@ const getVideoState = (storageDir: unknown) => {
 }    
 
 /**
- *  writeMetadataFile
+ *  Writes video metadata asynchronously and returns a Promise
  */
-const writeMetadataFile = (storageDir: string, metadata: Metadata) => {
+const writeMetadataFile = async (storageDir: string, metadata: Metadata) => {
+    const file = await getNewestVideo(storageDir);
```
aza547, line 270: I probably wrote this behaviour, but I think it's a bit crap that this function writes Metadata for the newest video. Seems like doing that will expose us to all sorts of confusion. Can we make the video to write metadata for an arg we pass in? Fine if you want to leave for me as this is again pre-existing.

src/main/util.ts
```diff
@@ -247,74 +264,101 @@ const getVideoState = (storageDir: unknown) => {
...
 /**
- * Delete the oldest video, unprotected video.
+ * Asynchronously delete the oldest, unprotected videos to ensure we don't store
+ * more material than the user has allowed us.
  */
-const deleteOldestVideo = (storageDir: any) => {
-    const sortedVideos = glob.sync(storageDir + "*.mp4")
-        .map((name: any) => ({name, mtime: fs.statSync(name).mtime}))
-        .sort((A: any, B: any) => B.mtime - A.mtime)
+const runSizeMonitor = async (storageDir: string, maxStorageGB: number): Promise<void> => {
```
aza547, line 303: This function is so much better. Love it.

src/main/util.ts
```diff
@@ -247,74 +264,101 @@ const getVideoState = (storageDir: unknown) => {
...
+            return { ...file, metadata, };
+        } catch (e) {
+            console.log(`Unable read and/or parse JSON from metadata file: ${file.name}`);
+        }
+    });
 
-    let videoForDeletion = sortedVideos.pop();
-    let deleteNotAllowed = isVideoProtected(videoForDeletion.name);
+    // Filter files with metadata
+    files = files.filter((file: any) => file && file.hasOwnProperty('metadata'));
```
aza547, line 321: Do we need `file &&` here? Maybe we do as a null check?

Also another comment - 

This will let videos without metadata build up. Maybe we should just delete them? Not sure what the best behaviour here is. I think I'm leaning towards delete anything without a metadata file.

The app won't display a video without metadata, the counter argument is that it could still be salvaged manually. I doubt many users are going to do that though. 

I've definitely already seen logs from users who have lots of leaked video files without metadata, it would be nice to protect against that.

src/main/util.ts
```diff
@@ -247,74 +264,101 @@ const getVideoState = (storageDir: unknown) => {
...
+ const getNewestVideo = async (storageDir: any): Promise<string> => {
+    const files = await getSortedVideos(storageDir);
+
+    if (files.length > 0) {
+        //@ts-ignore 'object is possibly undefined'
+        // but it can't, due to the 'if' above.
+        return files.shift().name;
+     }
+
+     return Promise.reject('no files');
```
aza547, line 361: no files could be a bit more descriptive

aza547 (commented):
Few final comments. Keen to get this in. Seems great.

src/main/util.ts
tdaugaard: Yeah, it was always there - I'll include it.

src/main/recorder.ts
tdaugaard: `.then()` takes a callback that will be called when the promise resolves, and in this case that would be the outer promise that will be resolved by that.

Basically, it has the same effect as doing:

```typescript
writeMetadataFile(this._storageDir, metadata).then(() => {
    resolve()
});
```

But without the extra callback (which just calls .. a .. callback).

src/main/util.ts
tdaugaard: Fixed.

src/main/util.ts
tdaugaard: We could use `file?.hasOwnProperty()` yes.

I think it'd be a good idea to simply remove those that don't have metadata. The videos are primarily intended to be viewed and used i the app, so it makes sense to clean up properly after us.

Besides, the filter for no metadata was simply made as a guard against missing metadata which would cause the protected test to fail, but they we could remove the metadata check and simply have the protected check be:

```typescript
!file.hasOwnProperty('metadata') || !Boolean(file.metadata.protected)
```

If it doesn't have metadata, it will be considered for deletion.
If it _has_ metadata, but isn't protected it will also be considered for deletion.

Let me know what you think and I'll fix it whichever way you like the best.

aza547: Yeah - lets do it!

PR #93, opened 2022-09-10
Fixes cleanup errors if files are locked

### LINKED ISSUE #78: Closing app while video open in VLC gives error.
Can probably just put the cleanup behind a try block and let it fail.

tdaugaard:
If a video is open in another application, like VLC, and the application is closed - or if you attempt to delete that video - the app would throw an exception that we didn't catch.

This commit changes that so that we'll catch the error and log it to the console.
Fixes #78


aza547:
Another awesome fix. Thanks!

aza547 (approved):

PR #94, opened 2022-09-12
Feature: Raid instance difficulty

tdaugaard:
Closes #53. As the title says, introduces the saving and display of the difficulty for raids.

There isn't much space left on the VideoButton, but I managed to fit it in.

PR #95, opened 2022-09-13
Replace `string.split()` with `splitLogLine()`

tdaugaard:
### `splitLogLine`

Split log lines intelligently with respect to quotes, lists, tuples and what have we such that we can better use the data contained within the log lines.

`splitLogLine`  has a `maxSplits` argument that, when given, will limit the maximum number of parsed arguments returned. Technically, it's advisable to only parse up to as many arguments as needed, but that isn't done yet as that'd probably be a pre-emptive overoptimization.

The method now returns an instance of the new class `LogLine`, which contains both the timestamp and the line arguments. At the same time, all `handle*Line` methods have been updated to accept an argument of `LogLine` instead of a string.

`getCombatLogDate` has been moved to `LogLine` and simply called `date()`, all occurances of the former has been replaced by the latter.

This is not so much a performance improvement more than it's a logical grouping of methods/data.

### `handleLogLine`

`handleLogLine` has been refactored to use a `switch..case` style matching to avoid doing text searching with `string.contains()` for every line, which in a worst case scenario would perform this check as many times as there are `if`-conditions, which would be most of the time since we're only interested in a very limited subset of the combat log events that WoW logs.

`handleLogLine` now checks the line token type, like `COMBATANT_INFO` against an array of `interestingCombatLogEvents` and will only parse the full log line if one matches to improve the processing speed of combat events.

`handleRaid(Start|Stop)Line` has been renamed to `handleEncounter(Start|Stop)Line` as `ENCOUNTER_(START|STOP)` isn't exclusive to raids.

### DIFF
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -54,15 +54,19 @@ class LogLine {
      * Split the line by any delimiter that isn't a number
      */
     date (): Date {
-        const [month, day, hours, mins, secs, msec] = this.ts.split(/[^0-9]/, 6);
+        const timeParts = this.ts
+            .split(/[^0-9]/, 6)
+            .map(v => parseInt(v, 10))
+            .reverse();
+        const [msec, secs, mins, hours, day, month] = timeParts;
         const dateObj = new Date();
 
-        dateObj.setDate(parseInt(day, 10));
-        dateObj.setMonth(parseInt(month, 10));
-        dateObj.setHours(parseInt(hours, 10));
-        dateObj.setMinutes(parseInt(mins, 10));
-        dateObj.setSeconds(parseInt(secs, 10));
-        dateObj.setMilliseconds(parseInt(msec, 10))
+        if (day) dateObj.setDate(day);
+        if (month) dateObj.setMonth(month);
+        dateObj.setHours(hours);
+        dateObj.setMinutes(mins);
+        dateObj.setSeconds(secs);
+        dateObj.setMilliseconds(msec);
 
         return dateObj;
     }
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -16,23 +16,29 @@ let metadata: Metadata;
 let combatantMap: Map<string, Combatant> = new Map();
 let playerCombatant: Combatant | undefined;
 let testRunning: boolean = false;
+let isChallengeModeActive: boolean = false;
+
+declare interface LogLineHandlerType {
+    (line: LogLine): void
+};
+
+type InterestingCombatLogEventsType = { [key: string]: LogLineHandlerType }
 
 /**
- * Array of combat events we want to handle.
- *
- * This array is only created to avoid parsing every line, because
- * that'd be a heavy workload for nothing.
+ * List of of combat events we want to handle associated with their
+ * respective handler function.
  */
-const interestingCombatLogEvents = [
-    'ARENA_MATCH_START',
-    'ARENA_MATCH_END',
-    'ENCOUNTER_START',
-    'ENCOUNTER_END',
-    'ZONE_CHANGE',
-    'COMBATANT_INFO',
-    'SPELL_AURA_APPLIED',
-    'UNIT_DIED'
-];
+const interestingCombatLogEvents: InterestingCombatLogEventsType = {
+    'ARENA_MATCH_START': handleArenaStartLine,
+    'ARENA_MATCH_END': handleArenaStopLine,
+    'CHALLENGE_MODE_START': handleChallengeModeEndLine,
+    'CHALLENGE_MODE_END': handleChallengeModeStartLine,
+    'ENCOUNTER_START': handleEncounterStartLine,
+    'ENCOUNTER_END': handleEncounterStopLine,
+    'ZONE_CHANGE': handleZoneChange,
+    'COMBATANT_INFO': handleCombatantInfoLine,
+    'SPELL_AURA_APPLIED': handleSpellAuraAppliedLine,
+};
 
 /**
  * A parsed line from the WoW combat log
@@ -41,7 +47,7 @@ class LogLine {
     constructor (
         // Timestamp in string format, as-is, from the log
         // Example: '8/3 22:09:58.548'
-        public ts: string,
+        public timestamp: string,
 
         // Multi-dimensional array of arguments
         // Example: 'ARENA_MATCH_START', '2547', '33', '2v2', '1'
@@ -54,7 +60,7 @@ class LogLine {
      * Split the line by any delimiter that isn't a number
      */
     date (): Date {
-        const timeParts = this.ts
+        const timeParts = this.timestamp
             .split(/[^0-9]/, 6)
             .map(v => parseInt(v, 10))
             .reverse();
@@ -175,10 +181,21 @@ const tailFile = (path: string) => {
  * Splits a WoW combat line intelligently with respect to quotes,
  * lists, tuples, and what have we.
  *
+ * @param line Log line straight from the combat log
+ * @param maxSplits Madximum number of elements to find (same as `limit` for `string.split()` )
  */
 const splitLogLine = (line: string, maxSplits?: number): LogLine => {
+    // Avoid evaluating this on each iteration of the for(..) loop below
     const line_len = line.length
-    const list_items: string[][] = [];
+
+    // Array of items that has been parsedin the current scope of the parsing.
+    //
+    // This can end up being multidimensional in the case of some combat events
+    // that have complex data stored, like `COMBATANT_INFO`.
+    const list_items: any[] = [];
+
+    // Final argument list as parsed, which can contain one or more instances of `list_items`
+    // depending on the complexity of the combat log line
     const args_list: any[] = [];
     let in_quote = false;
     let open_lists = 0;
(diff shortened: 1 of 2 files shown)


tdaugaard:
This has been taken out of the M+ PR because of some issues with PR #94 that will be fixed with this feature.

tdaugaard:
Not tested yet, will do later.

tdaugaard:
Tested and found one "bug", which is fixed in a5188cc. Ready for review and merge, if approved.

src/main/logutils.ts
```diff
@@ -17,6 +17,61 @@ let combatantMap: Map<string, Combatant> = new Map();
...
+];
+
+/**
+ * A parsed line from the WoW combat log
+ */
+class LogLine {
+    constructor (
+        // Timestamp in string format, as-is, from the log
+        // Example: '8/3 22:09:58.548'
+        public ts: string,
```
aza547, line 44: Can we call this `timestamp`? I'm going to get `ts` confused with `typescript`

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
+
+    // Check if we are interested in this event, and if not, discard it
+    if (interestingCombatLogEvents.indexOf(logLine.args[0]) === -1) {
+        return;
+    }
+
+    // Parse the full line
+    logLine = splitLogLine(line)
+
+    switch (logLine.args[0]) {
```
aza547, line 274: Seems a shame to have to hardcode all the interesting events again in this switch block. Is there some neater syntax we could use here? I'm not aware of one - but maybe you have some ideas.

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
 
     // If all goes to plan we don't need this but we do it incase the game
     // crashes etc. so we can still get a reasonable duration.
-    videoStartDate = getCombatLogDate(line);
+    videoStartDate = line.date();
 
     metadata = {
         name: "name",
-        category: category,
+        category: line.args[3],
```
aza547, line 314: I think this is totally fine for now, but we could have a subclass that extends LogLine for each "interesting" event type with simple a .metadata() function to avoid having to have carefully set indexes in here.

src/main/logutils.ts
```diff
@@ -177,9 +337,9 @@ const handleArenaStartLine = (line: string) => {
     // Helpfully ARENA_MATCH_END events contain the game duration. Solo shuffle
     // ARENA_MATCH_END duration only counts the last game so needs special handling. 
     if (metadata.category !== "Solo Shuffle") {
-        duration = parseInt(line.split(',')[2]);
+        duration = parseInt(line.args[2], 10);
```
aza547, line 340: We chatted about this 10 arg already. Is it technically possible to redefine parseInt? e.g.

```
const parseInt (i) => {
  return parseInt(i, 10);
}
```

aza547: If not, whatever - not a big deal.

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
+}
 
 /**
+ * Splits a WoW combat line intelligently with respect to quotes,
+ * lists, tuples, and what have we.
+ *
+ */
+const splitLogLine = (line: string, maxSplits?: number): LogLine => {
+    const line_len = line.length
+    const list_items: string[][] = [];
```
aza547, line 181: Could do with a comment explaining the purpose.

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
     });
-}    
+}
 
 /**
+ * Splits a WoW combat line intelligently with respect to quotes,
+ * lists, tuples, and what have we.
+ *
+ */
+const splitLogLine = (line: string, maxSplits?: number): LogLine => {
```
aza547, line 179: maxSplits should be mentioned in comment I think - not obvious at a glance what it is

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
+
+        if (in_quote) {
+            if (c === '"') {
+                in_quote = false;
+                continue;
+            }
+
+        } else {
+            switch (c) {
+            case ',':
```
aza547, line 206: indentation a bit off? a few lines below too

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
+                if (open_lists > 0) {
+                    list_items.at(-1)?.push(value);
```
aza547, line 208: What is the purpose of the `?` here? Null handling? Is that a plausible scenario

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
+    const timestamp = line.substring(0, tsIndexEnd);
+
+    for (let ptr = tsIndexEnd + 2; ptr < line_len; ptr++) {
+        const c = line.charAt(ptr);
+        if (c === '\n') {
+            break;
+        }
+
+        if (in_quote) {
+            if (c === '"') {
```
aza547, line 199: Does this logic need to be so heavily nested? Can we just pull this to the start and have:

```
if (c === '"') {
  in_quote = !in_quote;
  continue;
}
```

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
+                    args_list.push(value);
+                    if (maxSplits && args_list.length >= maxSplits) {
+                        break;
+                    }
+                }
+
+                value = '';
+                continue;
+
+            case '"':
```
aza547, line 219: see comments on in_quote above

src/main/logutils.ts
```diff
@@ -110,48 +165,153 @@ const tailFile = (path: string) => {
...
+
+        value += c;
+    }
+
+    if (value) {
+        args_list.push(value)
+    }
+
+    if (open_lists > 0) {
+        throw `Unexpected EOL. There are ${open_lists} open tuples/lists.`
```
aza547, line 253: tuples probably doesn't need mentioned here?

aza547:
> Tested and found one "bug", which is fixed in [a5188cc](https://github.com/aza547/wow-recorder/commit/a5188cc6975af418c7f9fbc7c932df46e5e11435). Ready for review and merge, if approved. It can still cause problems if some lines have all the date parts and some don't, but it's only the tests - regular log files won't have this problem.

FWIW this is a test script bug as far as I'm concerned - do we need this change? Happy to fix test scripts. Not fussed either way. Although I can't see any test scripts as you describe? I checked all the *.txt files I think. 

> Split log lines intelligently with respect to quotes, lists, tuples and what have we such that we can better use the data contained within the log lines.

I mentioned this in-line - I think you're referring to "tuples" as things in `()` and lists as things in `[]` - not a huge fan of that terminology for commenting and such as to me tuple means an immutable list (python) which I think isn't what this is! 

Great stuff overall.

aza547 (changes requested):

src/main/logutils.ts
tdaugaard: There is, but I wasn't too sure if you'd like it due to it perhaps looking too complex. I'll try it out and you can decide for yourself.

src/main/logutils.ts
tdaugaard: Oh, you've no idea how many things I wanted to do to this stuff, haha. I think that should go in a seperate PR because it'll be quite the change - but it'll come.

src/main/logutils.ts
tdaugaard: No, you can't redefine it, but we can literally just make a "parseInt10()" or something if you prefer a kind of syntactic sugar of sorts - shorthand or whatever.

src/main/logutils.ts
tdaugaard: Yes, unfortunately it does. Basically, the logic is:

- Are we inside a quoted string?
  - Yes: Is this the end of the quote?
    - Yes: Set flag false, continue loop
    - No: Do nothing, and let the current character be added to `value`
  - No: Check the current character aginst the special characters (which have no special meaning inside a quoted string)
    - Matches: Perform logic and contonue loop, or break out depending on the case
    - Doesn't match: Add to the current character to `value`

**If** I were to move it into the `case '"'` of the `switch` block, a comma, and brackets/parenthesis would suddenly have special meaning inside a quoted string.

src/main/logutils.ts
tdaugaard: Yes, exactly that. It's called _optional chaining (?.)_  (documented here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and is used because `.at()` can return `any`, like a string or number. Possible? Yes. Plausible? Not really. I just like to cover my bases since it would avoid an exception of calling `.push()` on a string, which would fail hard and can't be caught, since it'd be a runtime exception - If I recall correctly.

src/main/logutils.ts
tdaugaard: Yeah, it looks like it, but VSCode likes `case` to be at the same indentation level as `switch`. I guess it's too avoid a too high level of intendation (one extra level).

src/main/logutils.ts
tdaugaard: It's also possible to _remove_ the `radix` parameter and let JS sort it out - it _did_ work before and would probably cause no problems whatsoever to _not_ specify it.

src/main/logutils.ts
tdaugaard: Check out lastest commit - it does exactly this. If you don't like it, I'll remove it again :)

src/main/logutils.ts
tdaugaard: Nah

src/main/logutils.ts
tdaugaard: _Alternatively_, I could create methods on `LogLine` like:

```typescript
arg(index: number): any;
intArg(index: number): number;
hexArg(index: number): number;
```

That would then facilitate this and wrap it in those. That wouldn't take even a minute to fix.

aza547: Eh, not going to lose sleep over this then. Fine as is.

src/main/logutils.ts
aza547: I like it!

src/main/logutils.ts
aza547: Happy to leave as per other comment.

aza547 (approved):

PR #100, opened 2022-09-14
Remember volume and mute state when changing video

### LINKED ISSUE #99: remember muted state
by switching between videos the muted state is reset to unmuted. I'd love it to remember the setting.
also: really, really great work! keep it up

tdaugaard:
Simply as the title states. When changing video, save the current state and then restore it when `state.videoIndex` is changed. Fixes #99


src/renderer/Layout.tsx
```diff
@@ -102,9 +102,28 @@ export default function Layout() {
...
+    videoState: ipc.sendSync('getVideoState', categories),
+    videoMuted: false,
+    videoVolume: 1, // (Double) 0.00 - 1.00
   });
 
+  const getVideoPlayer = () => {
+    return (document.getElementById('video-player') as HTMLMediaElement);
+  }
+
+  const updateCurrentVideoSettings = () => {
```
aza547, line 114: Function name isn't totally clear. Sounds like it's updating a video element but actually it's changing the state. Can you change it or add a comment?

src/renderer/Layout.tsx
```diff
@@ -102,9 +102,28 @@ export default function Layout() {
...
+    const video = getVideoPlayer()
+    if (video) {
+      setState(prevState => {
```
aza547, line 117: Does changing the state here and again at line 144 cause a double render? Can we avoid that if so?

aza547 (changes requested):
Thanks for implementing this. Minor changes/questions as usual.

src/renderer/Layout.tsx
tdaugaard: No, it shouldn't double render because the `useEffect()` for updating it has a condition of only firing when `state.videoIndex` changes.

aza547: Doesn't the call to setState re-render the DOM (unrelated to the useEffect())? And you call useState twice here

aza547: (Unresolving as not yet convinced)

aza547: I think the useEffect runs after the component renders IF the state.videoIndex has changed. 

But this code path is if the user changes video, which will update the state once with to pull out the videos muted / volume status and then again to actually change the video index. 

Actually writing that out maybe it's fine, the first state update is cheap, we don't re-read all the videos from disk for either anyway, that's what I was worried about. I still think we're rendering twice here though.

tdaugaard: I'll see if I can find a way to determine if we double render.

tdaugaard: Wait, perhaps I'm not completely understanding how state works in React. Would it be possible to simply set the state like so:

```typescript
state.videoMuted = video.muted;
state.videoVolume = video.volume;
```

But without _using_ the state, causing a re-render? We don't actually need to re-render for this action.

tdaugaard: Yes, it works fine without the extra `useState()`. Today I learned! =)

aza547: This wasn't totally what I was expecting, but if it works then... ship it

tdaugaard: I had a manager once whose mantra was _fuck it, ship it_. I'm not sure it's great in the real world, but it sounded fun!

I think the idea is that while I was changing the state, I didn't want to apply the state until _after_ a new video was selected but I _did_ and that caused the double render.

PR #101, opened 2022-09-14
Few bug fixes

aza547:
- Time stamp in afk out test was causing us to clip video wrong and I hadn't noticed.
- Size monitor mainline but fix to add a `.name` so we pass a string not an object.
- Size monitor bug fix to always remove dangling video files (not just when they are in the set that are over the size budget)


aza547:
No idea how this ended up named feb22 backup. That was a mistake obviously.

src/main/util.ts
```diff
@@ -295,37 +295,47 @@ const getSortedVideos = async (storageDir: string): Promise<VideoInfo[]> => {
...
     });
 
-    // Consider files with NO metadata (dangling video files)
-    // and files that ARE NOT protected
-    files = files.filter((file: any) => !file.hasOwnProperty('metadata') || !Boolean(file.metadata.protected));
+    // Files without metadata are considered dangling and are cleaned up. 
+    const danglingFiles = files.filter((file: any) => !file.hasOwnProperty('metadata') || !file.metadata);
+    const unprotectedFiles = files.filter((file: any) => file.hasOwnProperty('metadata') && file.metadata && !Boolean(file.metadata.protected));
+
+    if (danglingFiles.length !== 0) {
```
tdaugaard, line 313: I'm thinking to make a list of "videosToDelete" instead and add to that, both those over size and also those dangling ones and _then_ delete all those instead of having two sections with deleteVideo?

Unless you want the log to reflect what is being done in detail

tdaugaard:
Looks good to me! :)

tdaugaard (approved):
Looks fine! :)

src/main/util.ts
aza547: Probably would be a bit neater but just going to stick with this.

Issue #96, opened 2022-09-13
Only recording a sub-section of screen, probably detecting resolution wrongly? 


aza547:
Issue from MrFlyingBananas on discord.

![image](https://user-images.githubusercontent.com/32389325/189993473-ff574dde-3cca-4bc0-8732-1c237d9a461f.png)

[WarcraftRecorder-2022-09-13.log](https://github.com/aza547/wow-recorder/files/9560263/WarcraftRecorder-2022-09-13.log)

aza547:
![image](https://user-images.githubusercontent.com/32389325/190268346-9131e8e8-d9c9-4145-962f-a48b736fce7e.png)

aza547:
Fixed in next release.

aza547:
Re-opening this because I spotted this in passing and might be a better solution:
https://github.com/Envek/obs-studio-node-example/pull/16/files

Issue #67, opened 2022-09-03
Quality should be configurable


aza547:
Right now video quality is quite high - files are large. 

We should make it a slider or something in settings so people can reduce it if they have disk space/upload concerns.

aza547:
We did this a while back!

PR #102, opened 2022-09-15
Partial recorder bug fix + app version included in logs / titlebar

NickIs211:


src/main/obsRecorder.ts
```diff
@@ -135,19 +135,54 @@ const setupScene = (monitorIndex: number) => {
...
+  // Checks if val is +-1 compare
+  const isClose = (val: number, compare: number) => {
+    return Math.abs(compare - val) <= 1;
+  };
+
+  // Checks if resolution is close enough to monitor, and if so, sets the settingString.
+  const checkRes = (resolution: string) => {
+    const [resW, resH] = resolution.split('x');
+
+    if (
```
tdaugaard, line 149: This could literally be:
```typescript
return isClose(parseInt(resW, 10), physicalWidth) && isClose(parseInt(resH, 10), physicalHeight);
```

src/main/obsRecorder.ts
```diff
@@ -135,19 +135,54 @@ const setupScene = (monitorIndex: number) => {
...
+      return true;
+    }
+
+    return false;
+  };
+
+  // TODO: Output should eventually be moved into a setting field to be scaled down. For now it matches the monitor resolution.
+  obsSettings.forEach((subCategory: any) => {
+    subCategory.parameters.forEach((param: any) => {
+      if (param.name === 'Base' || param.name === 'Output') {
```
tdaugaard, line 162: It's usually better to either return early or use `.filter()` to remove the unwanted elements than to indent everything one more level. Fx.:

```typescript
obsSettings.forEach((subCategory: any) => {
  subCategory.parameters.forEach((param: any) => {
    if (!(param.name === 'Base' || param.name === 'Output')) {
      return;
    }

    console.log(param.name, param.currentValue, param.values);
    param.values.forEach((vpair: { [key: string]: string }) => {
      const res = Object.keys(vpair)[0];
      if (checkRes(res)) {
        param.currentValue = res;
        console.info('Setting Video', param.name, 'to', res);
      }
    });
  });
});
```

tdaugaard: By the way, could this not easily be implemented using the existing methods to avoid duplicating the settings code?
```typescript
getAvailableValues (category: any, subcategory: any, parameter: any);
setSetting (category: any, parameter: any, value: any);
```

tdaugaard:
I think there might be a better way to fix the issue of this +/- 1 pixel thing due to the scaling; someone said somewhere on the Internet that you could simply use `Math.floor()` on the calculation where it multiplies by the scaling factor. Perhaps that would be better to test out?

aza547:
> I think there might be a better way to fix the issue of this +/- 1 pixel thing due to the scaling; someone said somewhere on the Internet that you could simply use `Math.floor()` on the calculation where it multiplies by the scaling factor. Perhaps that would be better to test out?

I think this is what @NickIs211 initially suggested in discord and I steered away from it. Might have been right all along. Not had a chance to look over this beyond the comments yet so not sure of my opinion. Will get a chance at the weekend.

src/main/obsRecorder.ts
NickIs211: ;;;;;;; 
don't look at me I'm hideous

src/main/obsRecorder.ts
NickIs211: I wrote the code to do everything in the same loop, rather than looping 4x to find, set, find, set when the functions are used

tdaugaard: I'd argue that the reduction of duplicate code outweighs the performance penalty, which is neglible, but ultimately it's up to @aza547. :)

As long as it's not a method that is called hundreds of times per second, it's not a problem.

src/main/main.ts
```diff
@@ -152,13 +154,18 @@ const createWindow = async () => {
...
   mainWindow.on('ready-to-show', () => {
     if (!mainWindow) throw new Error('"mainWindow" is not defined');
 
-    updateStatus(checkConfig() ? 0 : 2);
+    const initialStatus = checkConfig() ? AppStatus.WaitingForWoW : AppStatus.InvalidConfig;
+
+    updateStatus(initialStatus);
+
+    // This shows the correct version on a release build, not during development.
+    mainWindow.webContents.send('updateTitleBar', 'Warcraft Recorder v' + app.getVersion());
```
aza547, line 168: Where does the 18.0.3 come from I see on the dev version?

aza547 (commented):
Hi, thanks for the PR, sorry for taking so long to get to it (and I've still only had a second glance). There is a bunch of changes here that I think are due to other PRs. Could you catch up with main by either merging it in or rebasing please? 

I've quickly checked the version stuff which is awesome, curious why I see 18.0.3 on dev builds? You mentioned in comments that it works fine for release builds. Why the difference? I searched the repo for 18.0.3 and couldn't find a reference to it! Where is it coming from? 

I've not managed to review the OBS bits as I'm not awake enough, but I promise I'll get to them soon!

src/main/main.ts
NickIs211: app.getVersion() changes return value depending on how it's launched, see [here](https://github.com/electron/electron/issues/7085#issuecomment-244584704).

package.json:192 means any version of electron that is 18.x.x, so that's where 18.0.3 is coming from.
`"electron": "^18.0.1",`

NickIs211: Also caught up

aza547: Diff still not quite right, but I'll try to make do

src/main/obsRecorder.ts
aza547: I would prefer this to call to setSetting - even if it sacrifices a bit of efficiency, I don't expect this to be a bottleneck.

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
   // Correct the monitorIndex. In config we start a 1 so it's easy for users. 
   const monitorIndexFromZero = monitorIndex - 1; 
-  console.info("monitorIndexFromZero:", monitorIndexFromZero);
+  console.info("[OBS] monitorIndexFromZero:", monitorIndexFromZero);
   const { physicalWidth, physicalHeight } = displayInfo(monitorIndexFromZero);
 
+  const obsSettings = osn.NodeObs.OBS_settings_getSettings('Video').data;
+
+  // Checks if val is +-1 compare
+  const isClose = (val: number, compare: number) => {
```
aza547, line 142: Can this function go in utils please!

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
+    return Math.abs(compare - val) <= 1;
+  };
+
+  // Checks if resolution is close enough to monitor, and if so, sets the settingString.
+  const checkRes = (resolution: string) => {
```
aza547, line 147: Prefer functions not to be nested in other functions. Also this comment isn't quite right here, this doesn't do any setting.

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
+  // TODO: Output should eventually be moved into a setting field to be scaled down. For now it matches the monitor resolution.
+  obsSettings.forEach((subCategory: any) => {
+    subCategory.parameters.forEach((param: any) => {
+      if (!(param.name === 'Base' || param.name === 'Output')) {
+        return;
+      }
+
+      param.values.forEach((vpair: { [key: string]: string }) => {
+        const res = Object.keys(vpair)[0];
+        if (checkRes(res)) {
```
aza547, line 162: What if this is not close? I think we should throw an error and make it very obvious in logs

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
+  // Checks if val is +-1 compare
+  const isClose = (val: number, compare: number) => {
+    return Math.abs(compare - val) <= 1;
+  };
+
+  // Checks if resolution is close enough to monitor, and if so, sets the settingString.
+  const checkRes = (resolution: string) => {
+    const [resW, resH] = resolution.split('x');
+
+    return isClose(parseInt(resW, 10), physicalWidth) && isClose(parseInt(resH, 10), physicalHeight);
```
aza547, line 150: Can we make this extra clear:

```
const isWidthclose = ...
const isHeightCLose = ...

return (isWitdhClose && isHeightClose);
```

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
+  };
+
+  // TODO: Output should eventually be moved into a setting field to be scaled down. For now it matches the monitor resolution.
```
aza547, line 153: Where did this information come from? It's news to me - I'm quite keen to understand OBS better so if you have knowledge please share!

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
+  const obsSettings = osn.NodeObs.OBS_settings_getSettings('Video').data;
+
+  // Checks if val is +-1 compare
+  const isClose = (val: number, compare: number) => {
+    return Math.abs(compare - val) <= 1;
+  };
+
+  // Checks if resolution is close enough to monitor, and if so, sets the settingString.
+  const checkRes = (resolution: string) => {
+    const [resW, resH] = resolution.split('x');
```
aza547, line 148: resW -> resWidth
resH -> resHeight

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
   const { physicalWidth, physicalHeight } = displayInfo(monitorIndexFromZero);
 
+  const obsSettings = osn.NodeObs.OBS_settings_getSettings('Video').data;
+
+  // Checks if val is +-1 compare
+  const isClose = (val: number, compare: number) => {
+    return Math.abs(compare - val) <= 1;
+  };
+
+  // Checks if resolution is close enough to monitor, and if so, sets the settingString.
```
aza547, line 146: Comment style should be consistent with others 

```
/*
* comment here
*/
```

src/main/obsRecorder.ts
```diff
@@ -133,22 +133,51 @@ const setupScene = (monitorIndex: number) => {
...
+  const checkRes = (resolution: string) => {
+    const [resW, resH] = resolution.split('x');
+
+    return isClose(parseInt(resW, 10), physicalWidth) && isClose(parseInt(resH, 10), physicalHeight);
+  };
+
+  // TODO: Output should eventually be moved into a setting field to be scaled down. For now it matches the monitor resolution.
+  obsSettings.forEach((subCategory: any) => {
```
aza547, line 154: I think this should all go in a separate appropriately named function with a big ol' header comment to make it clear exactly why we need this.

aza547 (changes requested):
Hi, happy enough with the function, but a few code style comments for you to fix up. Thanks!

src/main/obsRecorder.ts
NickIs211: @aza547 updated

> I would prefer this to call to setSetting - even if it sacrifices a bit of efficiency, I don't expect this to be a bottleneck.

src/main/obsRecorder.ts
NickIs211: image

Video.Base translates to Base Canvas Resolution, which is the size of the scene we are creating. For current purposes, this should always match monitor resolution.
Video.Output translates to Output Resolution, which is the video resolution we're recording at. For users where storage space is a concern, they should be able to change the output resolution to a size that they find acceptable, say 720p. Currently, with this setup, users with 4k monitors will record at 4k.

NickIs211:
@aza547 code refactored based on comments above

src/main/obsRecorder.ts
aza547: Ah I see. That's interesting for https://github.com/aza547/wow-recorder/issues/67 if you fancy tackling another issue after!

src/main/obsRecorder.ts
```diff
@@ -125,30 +125,70 @@ const displayInfo = (displayIndex: number) => {
...
+
+  // TODO: Output should eventually be moved into a setting field to be scaled down. For now it matches the monitor resolution.
+  setOBSVideoResolution(physicalWidth, physicalHeight, 'Output');
+
+  const videoSource = osn.InputFactory.create(byOS({ [OS.Windows]: 'monitor_capture', [OS.Mac]: 'display_capture' }), 'desktop-video');
+
   // Update source settings:
   let settings = videoSource.settings;
   settings['monitor'] = monitorIndexFromZero;
-  settings['width'] = physicalWidth;
```
aza547, line 142: Were these doing anything at all?

aza547 (changes requested):
Nice. This looks good, with the only problem being I'm not sure what happens if I try to merge this given the massive "files changed" page. I'm not good enough with git to know how to resolve (thought merge / rebase would do it) - can you either fix that via some git means (if you are better with it than me!) or re-create the PR so it's got a clean diff? Then I'm happy to merge!

aza547:
Also in the new PR, please add a changelog fixed entry linked to https://github.com/aza547/wow-recorder/issues/96.

src/main/obsRecorder.ts
NickIs211: They weren't being respected when creating anything really, and looking at 

https://github.com/stream-labs/obs-studio-node/blob/5e22aaa87103ddc4f234d6b0f73d173e9b76b507/tests/osn-tests/util/input_settings.ts#L106
https://github.com/stream-labs/obs-studio-node/blob/d048994aac4c3988e5f300d835feabdf8e57a3df/tests/osn-tests/src/test_osn_input.ts#L134

They don't seem to be used properties during osn testing. Also couldn't find the related fields in the libobs docs.

NickIs211:
> Nice. This looks good, with the only problem being I'm not sure what happens if I try to merge this given the massive "files changed" page. I'm not good enough with git to know how to resolve (thought merge / rebase would do it) - can you either fix that via some git means (if you are better with it than me!) or re-create the PR so it's got a clean diff? Then I'm happy to merge!

Should be all good now. I initially merged main branch instead of bug-fixes ;;;;;;

aza547:
> > Nice. This looks good, with the only problem being I'm not sure what happens if I try to merge this given the massive "files changed" page. I'm not good enough with git to know how to resolve (thought merge / rebase would do it) - can you either fix that via some git means (if you are better with it than me!) or re-create the PR so it's got a clean diff? Then I'm happy to merge!
> 
> Should be all good now. I initially merged main branch instead of bug-fixes ;;;;;; @aza547

Hmm I still see 22 files changed (and lots of stuff from other commits). I think I was expecting merging main to be the correct thing to do (I'm not sure what bug-fixes is/was).

PR #104, opened 2022-09-15
 Cleanup application status values and logic

tdaugaard:
This is to make it more easier to identify the status values in the code and allow stronger typing on the values.

- Create an enum for the application status values in a new module, `main/types`.

- Replace `if..else` with an object of messages mapped to the new `AppStatus` enums.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -448,6 +448,5 @@ app
 
 export {
   mainWindow,
-  recorder,
-  AppStatus
+  recorder
 };
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -11,6 +11,7 @@ import Store from 'electron-store';
 const obsRecorder = require('./obsRecorder');
 import { Recorder } from './recorder';
 import { getAvailableAudioInputDevices, getAvailableAudioOutputDevices } from './obsAudioDeviceUtils';
+import { AppStatus } from './types';
 let recorder: Recorder;
 
 /**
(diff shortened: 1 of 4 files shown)


src/renderer/Status.tsx
```diff
@@ -15,54 +32,22 @@ export default function Status() {
    *   3 - watching
    *   4 - saving
    */
-  const [status, setStatus] = React.useState(0);
+  const [status, setStatus] = useState(AppStatus.WaitingForWoW);
```
aza547, line 35: Comment above this can go

src/renderer/Status.tsx
```diff
@@ -3,7 +3,24 @@ import eyeIcon from  '../../assets/icon/sleep-icon.png';
...
+  title: string,
+  type: StatusMessageType,
+  icon: string,
+};
+
+const statusMessages: { [key: number]: StatusMessageObjectType } = {
+  [AppStatus.WaitingForWoW]:   { type: 'status', icon: eyeIcon, title: 'Waiting for WoW to start' },
+  [AppStatus.Recording]:       { type: 'status', icon: recordIcon, title: 'Recording' },
+  [AppStatus.InvalidConfig]:   { type: 'error', icon: errorIcon, title: 'Failed to launch, check config is valid' },
+  [AppStatus.ReadyAndWaiting]: { type: 'status', icon: watchIcon, title: 'Ready and waiting'},
```
aza547, line 21: ReadyAndWaiting could be better named? We're buffering recording files at this point so doing work, not quite what is implied by this name

src/main/main.ts
```diff
@@ -443,10 +448,6 @@ app
 
 export {
   mainWindow,
-  recorder
+  recorder,
+  AppStatus
```
aza547, line 452: Weird to export AppStatus here? It's imported from types.ts

aza547 (changes requested):
Awesome clean-up. Minor comments only.

src/renderer/Status.tsx
tdaugaard: I just named them more or less according to how they were described :D I'll find a better name. I initially wanted to go with `ReadyToRecord`.

src/main/main.ts
tdaugaard: Sorry, I initially had it here and moved it to `types.ts` - I'll axe it.

Issue #76, opened 2022-09-05
Don't save raid encounter resets.


aza547:
Or atleast make it a configurable option to skip recording fights that are only a few seconds long?

Might affect legacy content runs but surely that's fine

tdaugaard:
Any idea of how to abandon a recording that has been started, that we don't want to keep - it should be easy to have a setting like "_Discard recordings of encounters shorter han X seconds_", but the actual logic behind having the recorder discard what has already been recorded, I am unsure of that.

aza547:
I think the logic is to put in a stop recording function that never calls finalizeVideo() - the finalizeVidoeo() function is responsible for cutting the temp recording to size and moving it to the storage dir. If we skip that and just restarted the buffering i think that's all we need.

PR #107, opened 2022-09-16
Add minimum encounter duration setting

DThielke:
Adds a new setting to only keep videos for raid encounters longer than X seconds (excluding the additional overrun time). Defaults to 15 seconds.

image

Fixes https://github.com/aza547/wow-recorder/issues/76

### DIFF
--- a/src/main/recorder.ts
+++ b/src/main/recorder.ts
@@ -21,18 +21,20 @@ const glob = require('glob');
     private _monitorIndex: number;
     private _audioInputDeviceId: string;
     private _audioOutputDeviceId: string;
+    private _minEncounterDuration: number;
 
 
     /**
      * Constructs a new Recorder.
      */
-    constructor(storageDir: string, maxStorage: number, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string) {
+    constructor(storageDir: string, maxStorage: number, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string, minEncounterDuration: number) {
         console.debug("Construcing recorder with: ", storageDir, maxStorage, monitorIndex);
         this._storageDir = storageDir;
         this._maxStorage = maxStorage;     
         this._monitorIndex = monitorIndex;           
-        this._audioInputDeviceId = audioInputDeviceId
-        this._audioOutputDeviceId = audioOutputDeviceId
+        this._audioInputDeviceId = audioInputDeviceId;
+        this._audioOutputDeviceId = audioOutputDeviceId;
+        this._minEncounterDuration = minEncounterDuration;
 
         // Something like: C:\Users\alexa\AppData\Local\Temp\WarcraftRecorder
         this._bufferStorageDir = path.join(app.getPath("temp"), "WarcraftRecorder"); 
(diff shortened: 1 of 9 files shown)


src/main/recorder.ts
```diff
@@ -180,9 +183,11 @@ const glob = require('glob');
             // Update the GUI to show we're processing a video. 
             if (mainWindow) mainWindow.webContents.send('updateStatus', AppStatus.SavingVideo);
 
-            // Cut the video to length and write its metadata JSON file.
-            // Await for this to finish before we return to waiting state.
-            await this.finalizeVideo(metadata);
+            if ((metadata.category != "Raids") || (metadata.duration - overrun >= this._minEncounterDuration)) {
```
tdaugaard, line 186: Please use `VideoCategory.Raids` instead of a literal string.

tdaugaard:
Ignore the comment I made - it was made with some other code I've been looking at in mind, which isn't yet merged into main. Carry on! 😂

src/main/recorder.ts
aza547, line 186: This is a tad much in a one line condition for my liking, how about:

```
const isRaid = (metadata.category != "Raids");
const isLongEnough = ((metadata.duration - overrun) >= this._minEncounterDuration);

if (isRaid && isLongEnough) {
  (...)
}


```

src/main/recorder.ts
```diff
@@ -21,18 +22,20 @@ const glob = require('glob');
...
     private _audioInputDeviceId: string;
     private _audioOutputDeviceId: string;
+    private _minEncounterDuration: number;
 
 
     /**
      * Constructs a new Recorder.
      */
-    constructor(storageDir: string, maxStorage: number, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string) {
+    constructor(storageDir: string, maxStorage: number, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string, minEncounterDuration: number) {
```
aza547, line 31: Also getting a bit busy syntactically, how about:

```
constructor(storageDir: string, 
            maxStorage: number, 
            monitorIndex: number, 
            audioInputDeviceId: string, 
            audioOutputDeviceId: string, 
            minEncounterDuration: number)
```

src/main/recorder.ts
```diff
@@ -260,8 +265,9 @@ const glob = require('glob');
     /**
      * Reconfigure the underlying obsRecorder. 
      */
-    reconfigure = (outputPath: string, maxStorage: number, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string) => {
+    reconfigure = (outputPath: string, maxStorage: number, monitorIndex: number, audioInputDeviceId: string, audioOutputDeviceId: string, minEncounterDuration: number) => {
```
aza547, line 268: can we split args per line as this is getting unreadable

src/main/util.ts
```diff
@@ -414,6 +414,12 @@ const runSizeMonitor = async (storageDir: string, maxStorageGB: number): Promise
         return false;
     }
 
+    const minEncounterDuration = getNumberConfigSafe(cfg, 'min-encounter-duration');
+
+    if ((!minEncounterDuration) || (minEncounterDuration < 0) || (minEncounterDuration > 10000)) {
```
aza547, line 419: Totally fine for now, but we're building on a quick fix I made to avoid invalid config - it's a crap user experience to get told "config invalid" and not get told why. Someone should come fix this one day!

aza547 (changes requested):
Hi and thanks for the PR! It's great to have more contributors.

Functionally this looks great. I've no real comments, but made some superficial syntax stuff that I'd love to be tidied up before merging.

Let me know if you want to discuss!

src/main/util.ts
DThielke: Agreed - this would really be better served by actually specifying a schema (including defaults) for the cfg store.

DThielke: I'll take a crack at this in a separate PR since it involves some broader changes.

aza547:
Forgot to suggest that you add a changelog and some tests for this so just did that here: https://github.com/aza547/wow-recorder/commit/9a9d60cdf0d05cf4b08494a933be4375f1a0dee3.

Thanks for the contribution!

PR #108, opened 2022-09-16
Cleanup the isUnitSelf() method

tdaugaard:
- Add `enum UnitFlag` to types for easier identification of the various flags that a unit can have.

- Add `isUnitPlayer()` to easier identify a player controlled unit.

- Refactor `isUnitSelf()` to make use of `isUnitPlayer()` and `UnitFlags`.

To prove that nothing has change logically, the old `isUnitSelf()` method checked the flags against the hexadecimal value 0x511 which consists of the flags:
```typescript
UnitFlags.AFFILIATION_MINE = 1
UnitFlags.REACTION_FRIENDLY = 16
UnitFlags.CONTROL_PLAYER = 256
UnitFlags.TYPE_PLAYER = 1024
```
1297 decimal, 0x511 hex.

### DIFF
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -2,6 +2,7 @@
 import { Combatant } from './combatant';
 import { recorder }  from './main';
 import { battlegrounds }  from './constants';
+import { UnitFlags } from './types';
 
 const tail = require('tail').Tail;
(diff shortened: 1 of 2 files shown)


aza547 (approved):
LGTM!

PR #109, opened 2022-09-16
Refactor 'categories' to be an enum type

tdaugaard:
As the title states. A simple refactor to eliminate literal strings from comparisons and assignments.

### DIFF
--- a/src/renderer/VideoButton.tsx
+++ b/src/renderer/VideoButton.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import Tab from '@mui/material/Tab';
 import Menu from '@mui/material/Menu';
 import MenuItem from '@mui/material/MenuItem';
-import { categories, videoButtonSx, specToClass }  from '../main/constants';
+import { categories, videoButtonSx, specToClass, VideoCategory }  from '../main/constants';
 import { getResultText, getFormattedDuration, getInstanceDifficulty } from './rendererutils';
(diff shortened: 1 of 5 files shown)


src/renderer/rendererutils.ts
```diff
@@ -1,24 +1,20 @@
...
+        VideoCategory.SoloShuffle,
+        VideoCategory.Battlegrounds,
     ];
 
-    if (pvpCategories.includes(category)) {
-        return true;
-    } else {
-        return false;
-    }
+    return pvpCategories.indexOf(category) !== -1;
```
aza547, line 17: Why not just `.includes()`?

src/renderer/VideoButton.tsx
```diff
@@ -21,7 +21,7 @@ export default function VideoButton(props: any) {
   const videoIndex = props.index;
 
   const categoryIndex = state.categoryIndex;
-  const category = categories[categoryIndex]
+  const category = categories[categoryIndex] as VideoCategory
```
aza547, line 24: Semicolon

aza547 (changes requested):
A few clean-up comments, otherwise looks good.

src/renderer/rendererutils.ts
tdaugaard: My bad. No idea why I thought that was a good idea to change.

Issue #74, opened 2022-09-04
Add a version check


aza547:
Something we discussed, a link back to GitHub to check for new versions and display if an upgrade is available would be nice.

Crucially it shouldn't be obnoxious, and should be easily dismissed if a user doesn't want to upgrade.

aza547:
While in the area it would also be good to log the version, we don't do that currently.

tdaugaard:
Perhaps take a look electron-builder https://www.electron.build/, which has an Auto Update feature if that's something you're interested in.

aza547:
being handled here https://github.com/aza547/wow-recorder/pull/102

aza547:
Actually that's just the adding of a version number to logs / title. We still want a check to confirm it's the latest.

PR #114, opened 2022-09-18
Feature check updates

NickIs211:
Addresses #74 
Also cleans up button bar css

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -448,7 +448,7 @@ app.on('window-all-closed', () => {
 });
 
 /**
- * Checks for updates from the releases page on github, andif there is a new version, sends a message to the main window to display a notification
+ * Checks for updates from the releases page on github, and, if there is a new version, sends a message to the main window to display a notification
  */
 const checkAppUpdate = () => {
   const options = {
@@ -457,7 +457,7 @@ const checkAppUpdate = () => {
     path: '/repos/aza547/wow-recorder/releases/latest',
     method: 'GET',
     headers: {
-      'User-Agent': 'wow-recorder'
+      'User-Agent': 'wow-recorder',
     }
   }
 
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -448,7 +448,7 @@ app.on('window-all-closed', () => {
 });
 
 /**
- * https://api.github.com/repos/aza547/wow-recorder/releases/latest
+ * Checks for updates from the releases page on github, andif there is a new version, sends a message to the main window to display a notification
  */
 const checkAppUpdate = () => {
   const options = {
@@ -473,7 +473,7 @@ const checkAppUpdate = () => {
       const release = JSON.parse(data);
       const latestVersion = release.tag_name;
       const downloadUrl = release.assets[0].browser_download_url;
-      
+
       if (latestVersion !== app.getVersion()) {
         console.log("[Main] New version available:", latestVersion);
         if (mainWindow) mainWindow.webContents.send('updateAvailable', downloadUrl);
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -470,6 +470,11 @@ const checkAppUpdate = () => {
     });
 
     response.on('end', () => {
+      if(response.statusCode !== 200) {
+        console.error(`[Main] ERROR, Failed to check for updates, status code: ${response.statusCode}`);
+        return;
+      }
+
       const release = JSON.parse(data);
       const latestVersion = release.tag_name;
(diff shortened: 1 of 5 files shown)


src/renderer/VersionUpdateWidget.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  const openReleaseDownloadUrl = () => {
+    ipc.sendMessage('openURL', [downloadUrl]);
+  };
+
+  const [widgetClass, setWidgetClass] = React.useState("hidden");
+  const [downloadUrl, setDownloadUrl] = React.useState<string>("");
+
+  React.useEffect(() => {
+    window.electron.ipcRenderer.on('updateAvailable', (downloadUrl) => {
```
tdaugaard, line 15: You have `window.electron.ipcRenderer` aliased above as `ipc` already, might as well use it :)

aza547:
Hey - this is really cool. Thanks! While I review, could you post a screenshot of what it looks like?

NickIs211:
> Hey - this is really cool. Thanks! While I review, could you post a screenshot of what it looks like?

image
image

src/main/main.ts
```diff
@@ -445,6 +447,45 @@ app.on('window-all-closed', () => {
   app.quit();
 });
 
+/**
+ * Checks for updates from the releases page on github, andif there is a new version, sends a message to the main window to display a notification
```
aza547, line 451: `andif` typo

src/main/main.ts
```diff
@@ -445,6 +447,45 @@ app.on('window-all-closed', () => {
...
+    headers: {
+      'User-Agent': 'wow-recorder'
+    }
+  }
+
+  const request = net.request(options);
+  request.on('response', (response) => {
+    let data = '';
+
+    response.on('data', (chunk) => {
```
aza547, line 468: Why do we need this? Are we expecting data to be returned in multiple chunks? I was expecting the response would be short enough that it would be in a single response. 

That being said, if this is just the done thing, it's probably fine - I've never made a HTTP request in JS before (and being robust to all possibilities is obviously a good thing).

src/renderer/VersionUpdateWidget.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [downloadUrl, setDownloadUrl] = React.useState<string>('');
+
+  React.useEffect(() => {
+    ipc.on('updateAvailable', (downloadUrl) => {
+      setWidgetClass('');
+      setDownloadUrl(downloadUrl as string);
+    });
+  }, []);
+
+  return (
```
aza547, line 20: I'm by no means a react expert (or even good with react) but we do this differently elsewhere.

I would have done

```
const [widgetShown, setWidgetShown] = React.useState<boolean>(false);

(...)

{ widgetShown &&
 ... <the TSX>
}
```

See videoButton around line 200 for examples. 

Happy to be convinced this is better, but interested in your thoughts.

aza547 (changes requested):
Thanks for the feature, this seems great. Minor comments.

src/main/main.ts
NickIs211: The response has a [data event](https://www.electronjs.org/docs/latest/api/incoming-message#event-data) that delivers chunks, meaning data might be called multiple times to build the full json object. 
See https://en.wikipedia.org/wiki/Chunked_transfer_encoding

src/renderer/VersionUpdateWidget.tsx
NickIs211: Yeah, I'm used to working on Vue rather than React, this is implemented in a similar way as described above.

src/main/main.ts
```diff
@@ -445,6 +447,45 @@ app.on('window-all-closed', () => {
...
+      data += chunk;
+    });
+
+    response.on('end', () => {
+      const release = JSON.parse(data);
+      const latestVersion = release.tag_name;
```
aza547, line 474: What is this response doesn't contain the data?

aza547: Actually I suspect the latter will just never return the event so maybe fine. Still curious about the former.

NickIs211: > What is this response doesn't contain the data?
Just added 200 check along with making sure the variables have something in them. If the status 200 page doesn't have the data, it's either a GitHub issue or a release tagging issue.

> What if I start the app with my internet unplugged?
The request will go into request.error and output to logs.

aza547:
Awesome - looks good to me! Merging.

PR #116, opened 2022-09-18
Write metadata with UTF-8 encoding

tdaugaard:
Ensure that metadata JSON is written to disk in UTF-8 encoding to support e.g. Russian character names. This relates to the new `playerDeath` metadata array which can look garbled when Russian character names appear in it.


tdaugaard:
Tested:

image

PR #117, opened 2022-09-18
Make <VideoButton/> for M+ look nice

tdaugaard:
- Remove "zone" text and include only the map name, which is the only important part for a dungeon.

- Add Keystone level at the top where raids have their difficulty letter displayed and make it LEGENDARY colored.

- Make dungeon map name centered vertically and horizontally to it looks nice

### DIFF
--- a/src/renderer/App.css
+++ b/src/renderer/App.css
@@ -199,13 +199,13 @@ h2 {
   text-align: center;
 }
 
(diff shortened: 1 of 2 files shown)


tdaugaard:
Proof that it looks nicer:

image

aza547:
Looks great!

PR #118, opened 2022-09-18
Make video player settings global

### LINKED ISSUE #99: remember muted state

tdaugaard:
Closes #99.

When a video is muted or the volume changed, that setting would only persist within the selected video category (e.g. 3v3) and if a different category was selected, the settings would reset which was undesirable.

This has been changed such that the settings are _global_ across the app regardless of category and video selection.

### In types.ts:

- New type `VideoPlayerSettings` which contains the muted/volume state values that we want to preserve.

### In preload.ts:

- New IPC channel `videoPlayerSettings` with two main functions as `args[0]`:

  `get`: retrieve the stored settings
  `set`: store the settings given in `args[1]`

### In main.ts:

- IPC listener to get/set video player settings.

### In Layout.tsx:

- New function `handleVideoPlayerVolumeChange()` which handles saving the state of the muted/volume when it changes and sets the values in the existing state, but doesn't apply it as that will happen in due time whenever a new video or category is selected.
- Formatted the `<video/>` element over multiple lines for readability.

### DIFF
--- a/src/renderer/Layout.tsx
+++ b/src/renderer/Layout.tsx
@@ -13,6 +13,7 @@ import VideoButton  from './VideoButton';
 import readyPoster  from  "../../assets/poster/ready.png";
 import notReadyPoster from  "../../assets/poster/not-ready.png";
 import { dungeon } from './images';
+import { VideoPlayerSettings } from 'main/types';
 
 /**
  * For shorthand referencing.
@@ -95,6 +96,13 @@ const tabProps = (index: number) => {
 }
 
 /**
+ * Get video player settings initially when the component is loaded,
+ * because we do want to get it from here one time, but subsequently
+ * we'll only need to use the IPC channel `videoPlayerSettings`.
+ */
+const videoPlayerSettings = (ipc.sendSync('videoPlayerSettings', ['get']) as VideoPlayerSettings);
+
+/**
  * The GUI itself.
  */
 export default function Layout() {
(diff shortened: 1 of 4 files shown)


src/renderer/Layout.tsx
```diff
@@ -94,6 +95,13 @@ const tabProps = (index: number) => {
   }
 }
 
+/**
+ * Get video player settings initially when the component is loaded,
+ * because we do want to get it from here one time, but subsequently
+ * we'll only need to use the IPC channel `videoPlayerSettings`.
```
aza547, line 101: Not sure I totally understand comment. This is hit whenever we render the component I think (i.e. change category or video) -- so the other than this we only use the IPC set method to tell main to remember the settings. Is that right?

aza547 (commented):
One comment for clarification, otherwise looks great. Have you tested?

aza547:
image

Hit an issue when changing video

aza547:
> image
> 
> Hit an issue when changing video

Fixed by removing 156 in Layout.tsx (old reference to readVideoPlayerSettings).

aza547:
Otherwise testing seems good, merging and fixing on main.

aza547:
See https://github.com/aza547/wow-recorder/commit/86a577ca077bc1bef1ad2571859d65de4de33a89

src/renderer/Layout.tsx
tdaugaard: Honestly, I'm not sure why it works - I don't quite understand the React component lifecycle well, but it works.

Issue #119, opened 2022-09-19
Application crashes when clicking on "Settings" button without a valid configuration


tdaugaard:
### Describe the bug

When the application does not have a valid configuration, as most new users won't have, it crashes without any errors in the console or anything else indicating a problem.

This has been tested to affect both the development environment as well as a packaged release version of the application.

### Affected version

Affetcs git commit fd44914d (most recent as of the time of writing this issue)

Application reports version 2.7.0.

### To Reproduce

1. Move or delete the folder `%AppData%\WarcraftRecorder` to ensure no old configuration exists
2. Clone repository
3. Run `npm install`
4. Run `npm run start`
5. Notice the warning triangle status indicating an invalid configuration state
6. Click "_Settings_" button (cog-wheel)
7. App crashes

### Expected behavior

I expect to be able to configure the application.

### Screenshots

None.

### Additional context

Console log from app start to crash:

```
[electronmon] waiting for a change to restart it
11:05:55.597 > [Main] App starting: version 18.0.3
11:05:55.655 > [Main] App ready
11:05:55.759 > (node:16060) ExtensionLoadWarning: Warnings loading extension at C:\Users\XXXXX\AppData\Roaming\Electron\extensions\fmkadmapgofadopljbjfkapdkoienihi:
  Manifest version 2 is deprecated, and support will be removed in 2023. See https://developer.chrome.com/blog/mv2-transition/ for more details.

(Use `electron --trace-warnings ...` to show where the warning was created)
<i> [webpack-dev-middleware] wait until bundle finished: /mainWindow.index.html
[16060:0919/110600.878:ERROR:CONSOLE(160)] "Electron sandbox_bundle.js script failed to run", source: node:electron/js2c/sandbox_bundle (160)
[16060:0919/110600.879:ERROR:CONSOLE(160)] "TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))", source: node:electron/js2c/sandbox_bundle (160)
11:06:33.851 > [Main] User clicked open settings
11:06:33.857 > (node:16060) ExtensionLoadWarning: Warnings loading extension at C:\Users\XXXXX\AppData\Roaming\Electron\extensions\fmkadmapgofadopljbjfkapdkoienihi:
  Manifest version 2 is deprecated, and support will be removed in 2023. See https://developer.chrome.com/blog/mv2-transition/ for more details.

11:06:34.196 > [Main] Got from config store:  storage-path null
[...]

PR #120, opened 2022-09-19
Fix crashing with invalid configuration

tdaugaard:
Fixes issue #119 which describes a problem in which an invalid configuration, as how the initial state of the app is for new users, will crash when one attempts to open the Settings window.

This is due to OBS not yet being initialized but attempted to be used to retrieve available input/output audio devices.

The fix is to simply return an empty array for these device arrays and let the config only show "All" and "None" until OBS has been initialized.


aza547:
Looks good!

PR #122, opened 2022-09-19
Fix typos in this._isRecordingBuffer

Emme:


aza547:
Not sure how we managed this! Thanks!

PR #126, opened 2022-09-19
Refactor: logutils simplification

tdaugaard:
### Refactor and simplify logutils

To reduce code duplication and therefore ease testing and lower the risk of bugs, a refactor of the logutil handlers has been done.

- `videoOverrunPerCategory` is now used to obtain the overrun per category rather than having each function specify its own.

- All logic surrounding adding `result` and `duration` to `metadata` has been moved to a new function `endRecording()` which takes options in  the form of `EndRecordingOptionsType` - all optional with reasonable   defaults:

    `duration`: calculated from `(videoStopDate - videoStartDate)`

    `result`: false

    `discardVideo`: false (obviously)

  If `discardVideo` is true, the recording will be discarded and not used, just as if a raid encounter was too short.

  Where previously a function would stop recording and set the result to `false` (`zoneChangeStop`, `battlegroundStop`, etc.) this is now done by defaults in `endRecording()` if no result is given.

- A new function to start recording is introduced `startRecording()` which takes a `VideoCategory` argument such that we have an indication of what we're recording in `currentActivity`. This will be used later to be able to handle various unexpected cases that results in a recording being ended, both to show it in the log but also to be able to differentiate between the categories.

- `combatantMap` and `playerCombatant` resets are now done in `startRecording()` to ensure it's _always_ done before a new recording starts.

- A new function `endChallengeModeDungeon()` to handle the logic of ending an in-progress Challenge Mode dungeon, because it will be used in mulitple places at a later time to account for unfinished dungeons, AFK'ing in an incomplete dungeon and such.

- `videoStopDate` is now global, like `videoStartDate` and is used by `endRecording()` to calculate video duration, if not explicitly given.

- A few `return` statements has been added to `handleZoneChange()` to make it more readable.

- `recorder.stop()` method has been enriched with a `discardVideo` argument that handles the logic that `endRecording()` neccessitates.


tdaugaard:
Force pushed due to errorneous order of statements in `logutils.ts` in `endRecording` regarding `currentActivity` being use after being unset.

aza547:
Looks great, agree this code was needing it. I'm going to need more time to review this.

src/main/logutils.ts
```diff
@@ -330,35 +381,19 @@ function handleArenaStartLine (line: LogLine): void {
...
-        metadata.playerName = playerCombatant.name;
-        metadata.playerRealm = playerCombatant.realm;
-        metadata.playerSpecID = playerCombatant.specID;        
-    }
-
-    let duration; 
+    let duration;
+    videoStopDate = line.date();
     
     // Helpfully ARENA_MATCH_END events contain the game duration. Solo shuffle
```
aza547, line 387: Is there any value in keeping this? We can just use the date events from logs and then we don't need any special handling for solo shuffle?

src/main/logutils.ts
```diff
@@ -465,21 +502,11 @@ function handleChallengeModeEndLine (line: LogLine): void {
...
-        console.debug("[ChallengeMode] Removing last timeline segment, because it's too short.")
-        activeChallengeMode.removeLastTimelineSegment();
-    }
+    const result = Boolean(parseInt(line.args[1]));
 
-    console.debug("[ChallengeMode] Ending Challenge Mode instance");
-    activeChallengeMode = undefined;
-    
-    recorder.stop(metadata, overrun);
+    endRecording({result,});
```
aza547, line 509: comma

aza547: Also I'm curious, how does this translate to the EndRecordingOptionsType, which has 3 optional fields? Maybe the comma is required but I don't understand the syntax?

src/main/logutils.ts
```diff
@@ -129,18 +131,67 @@ const wowProcessStopped = () => {
...
+};
+
+type EndRecordingOptionsType = {
+    discardVideo?: boolean, // Discard the video/don't save the recording
+    duration?: number,      // Explicitly give video duration
+    result?: boolean,       // Success/Failure result for the overall activity
+};
+
+/**
+ * Stop recording and mark as not doing anything
```
aza547, line 162: Worth stating in comment here that videoStartDate/videoStopDate should be set before calling this.

src/main/logutils.ts
tdaugaard: For PvP there might not be, no - I think only Mythic+ _requires_ the duration from the logs to be saved because it's always different from the video duration due to overrun, but I don't think any raid or PvP activity has this issue. I'll remove it!

src/main/logutils.ts
tdaugaard: No, the comma is not required. It's a shorthand way to specify that a variable with the same name a as a property in a type/object can be given by simply naming it.

`{result}` and `{result: result}` has the same .. result.

Optional fields are optional exactly for that reason, so we don't need to give them when we want them to have their default value as used in `endRecording()`.

src/main/logutils.ts
tdaugaard: Easy fix - if the method is called without a valid `videoStopDate` we'll set it to `new Date()` because it _needs_ to be set or it'll cause an invalid application state if we bug out of it or throw an error.

Also the whole `duration` thing in `EndRecordingOptionsType` is pointless - no activities specify it anymore. I'll remove it entirely.

src/main/logutils.ts
aza547: thanks for explaining!

PR #127, opened 2022-09-20
Restructure `Recorder` constructor arguments

tdaugaard:
The constructor arguments for `Recorder` had gotten a little out of hand and caused duplication code in a few places.

- All the arguments has been restructured into a type `RecorderOptionsType` that is passed to both the constructor as well as the `Recorder.reconfigure()` call. It also contains the buffer directory, as it didn't previously, because it's used in `obsRecorder.ts` and that object had the same problem and took mostly the same options.

- The default config checks has been moved to `loadRecorderOptions()` which is used to (re-) load recorder options on app load as well as when the Settings window is closed.

- `_bufferStorageDir` is now defined in `loadRecorderOptions()` as that makes the most sense.

- `obsRecorder.ts` `initialize()` now calls `reconfigure(options)` instead of duplicating the reconfiguration code.


src/main/recorder.ts
```diff
@@ -1,61 +1,48 @@
...
-        this._minEncounterDuration = minEncounterDuration;
-
-        // Something like: C:\Users\alexa\AppData\Local\Temp\WarcraftRecorder
-        this._bufferStorageDir = path.join(this._storageDir, ".temp"); 
-
-        if (!fs.existsSync(this._bufferStorageDir)) {
-            console.log("[Recorder] Creating dir:", this._bufferStorageDir);
-            fs.mkdirSync(this._bufferStorageDir);
+    constructor(options: RecorderOptionsType) {
+        this._options = options;
```
aza547, line 34: Not sure how much I like this purely due to now all over this file we have this.something.somethingelse.

We could stick with having the member variables on this class as they were, and just split out options in the constructor to set them? I think that reads better throughout this file.

You could still set this._options for passing to the obsRecorder file.

src/main/recorder.ts
```diff
@@ -279,19 +266,9 @@ const glob = require('glob');
...
-        maxStorage: number, 
-        monitorIndex: number, 
-        audioInputDeviceId: string, 
-        audioOutputDeviceId: string, 
-        minEncounterDuration: number
-    ) => {
-        this._maxStorage = maxStorage;
-        this._minEncounterDuration = minEncounterDuration;
-
+    reconfigure = (options: RecorderOptionsType) => {
```
aza547, line 269: Don't we need to set the _options here? I think we're going to end up in a state where we have reconfigured the obsRecorder but not the member variables on this class. We might just get away with it but definitely confusing.

aza547 (changes requested):
Some comments on this one.
Also conflicts to resolve I suspect because of one of the other PRs.

tdaugaard:
>  Also conflicts to resolve I suspect because of one of the other PRs.

I was fully expecting that - I will fix.

src/main/recorder.ts
tdaugaard: There's  lot of this.that.otherthing in many places and it personally doesn't bother me, but if you wish I can expand it to member variables.

src/main/recorder.ts
tdaugaard: Dang it, you're correct! This is also why it would be a bad idea to expand the options into separate member variables due to having to do it twice.

src/main/recorder.ts
aza547: Fair enough - given the other comment happy to leave

src/main/recorder.ts
```diff
@@ -1,14 +1,10 @@
 import { Metadata } from './logutils';
 import { writeMetadataFile, runSizeMonitor, getNewestVideo, deleteVideo, cutVideo, addColor, getSortedVideos } from './util';
 import { mainWindow }  from './main';
-import path from 'path';
 import { AppStatus } from './types';
 
 const obsRecorder = require('./obsRecorder');
 import fs from 'fs';
-import glob from 'glob';
```
aza547, line 10: Is this right? I thought you had added these

tdaugaard: `Recorder` no longer uses these - it uses `util.getSortedVideos()` instead :)

tdaugaard: I might have added them in a previous PR before I let `getSortedVideos()` do all the work :)

aza547:
Conflicts - then good to go!

Issue #128, opened 2022-09-20
Failure to cancel buffer restart timer 


aza547:
**Describe the bug**
See title. User did a M+. Following is in logs:

```
[2022-09-19 23:41:36.663] [info]  [OBS] obsRecorder: start
[2022-09-19 23:41:36.802] [debug] [OBS] Asserted OBS signal: recording start
[2022-09-19 23:41:43.633] [debug] [ChallengeMode] Starting Challenge Mode instance
[2022-09-19 23:41:43.634] [info]  [Recorder] Start recording by cancelling buffer restart
[2022-09-19 23:41:43.635] [info]  [Logutils] Handling zone change:  {
  timestamp: '9/19 23:41:26.922',
  args: [ 'ZONE_CHANGE', '2441', 'Tazavesh, the Veiled Market', '8' ]
}
[2022-09-19 23:42:12.780] [info]  [Logutils] Handling zone change:  {
  timestamp: '9/19 23:42:11.196',
  args: [ 'ZONE_CHANGE', '2441', 'Tazavesh, the Veiled Market', '8' ]
}
[2022-09-19 23:44:38.427] [debug] [ChallengeMode] Starting new boss encounter: Zo'phex the Sentinel
[2022-09-19 23:45:33.115] [debug] [ChallengeMode] Ending boss encounter: Zo'phex the Sentinel
[2022-09-19 23:46:33.390] [info]  [Recorder] Restart recording buffer
[2022-09-19 23:46:33.390] [info]  [OBS] obsRecorder: stop
[2022-09-19 23:46:33.404] [debug] [OBS] Asserted OBS signal: recording stopping
[2022-09-19 23:46:34.595] [debug] [OBS] Asserted OBS signal: recording stop
```

Notice the ordering:
23:41:33 - Buffer restarts as expected.
23:41:43 - Cancel buffer restart on M+ start.
23:46:33 - Buffer restarts exactly 5 mins later -- clearly it didn't get cancelled. 

The restart proceeds to continue throughout the dungeon, obviously giving the wrong size of video at the end. 

**Expected behavior**
Buffer is cancelled as expected. Video is expected size. 

**Screenshots**
https://i.imgur.com/hYkRd39.png

aza547:
[2022-09-19_15-24-24.txt](https://github.com/aza547/wow-recorder/files/9610872/2022-09-19_15-24-24.txt)
[...]

Issue #123, opened 2022-09-19
Problem with display scaling


Emme:
**Describe the bug**
When using windows display scaling, the app can't find a matching output resolution:

> [error] [OBS] ERROR! Matching resolution not found for Video Output
> [error] Error attempting to match 3841.25x2161.25 with 1920x1080,1280x720,1920x1200,1200x1920,3840x2160 for Base

**Expected behavior**
The output resolution should be set to the closest matching one

**Additional context**
The log entry for the relevant monitor:
```
  {
    id: 2841568472,
    bounds: { x: 0, y: 0, width: 2195, height: 1235 },
    workArea: { x: 0, y: 0, width: 2195, height: 1195 },
    accelerometerSupport: 'unknown',
    monochrome: false,
    colorDepth: 24,
    colorSpace: '{primaries:BT709, transfer:SRGB, matrix:RGB, range:FULL}',
    depthPerComponent: 8,
    size: { width: 2195, height: 1235 },
    displayFrequency: 60,
    workAreaSize: { width: 2195, height: 1195 },
    scaleFactor: 1.75,
    rotation: 0,
    internal: false,
    touchSupport: 'unknown'
  }
```
[WarcraftRecorder-2022-09-19.log](https://github.com/aza547/wow-recorder/files/9601497/WarcraftRecorder-2022-09-19.log)
[2022-09-19 19-29-11.txt](https://github.com/aza547/wow-recorder/files/9601498/2022-09-19.19-29-11.txt)

aza547:
Interesting! Another user had trouble with the scaleFactor rounding. We made a fix in the last release to "snap" to a close resolution so long as the resolution was within 1 pixel. Here it's 1.25 pixels off so that snapping doesn't happen. 

Easy fix that we should probably make is to make the snapping happen for a wider range (bumping it to 2 would fix it - but that might not solve for everyone...).

aza547:
@Emme I'd like to make a fix and share a rebuilt version of the app for you to test. Sound OK?

aza547:
https://drive.google.com/file/d/1rIsXBHJ_-LHptbb_M2Kf8DLAIdWIKqlo/view?usp=sharing - let me know if that sorts it!

Emme:
Yes, it's working now.

aza547:
Great. I'll mark this closed and will be fixed in 2.7.1 when it releases.

PR #131, opened 2022-09-20
Bug fix 123 & 128

aza547:
#123 
#128

### DIFF
--- a/src/main/recorder.ts
+++ b/src/main/recorder.ts
@@ -112,7 +112,7 @@ const glob = require('glob');
     
         // Guard against multiple buffer timers. 
         if (this._bufferRestartIntervalID) {
-            console.error("[Recorder] Already have a buffer interval.")
+            console.error("[Recorder] Already has a buffer interval.")
             return;
         }
 
--- a/src/main/recorder.ts
+++ b/src/main/recorder.ts
@@ -1,7 +1,6 @@
 import { Metadata } from './logutils';
 import { writeMetadataFile, runSizeMonitor, getNewestVideo, deleteVideo, cutVideo, addColor } from './util';
 import { mainWindow }  from './main';
-import { app } from 'electron';
 import path from 'path';
 import { AppStatus } from './types';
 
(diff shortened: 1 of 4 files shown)


tdaugaard (approved):
Looks fine to me!

src/main/recorder.ts
```diff
@@ -130,6 +134,8 @@ const glob = require('glob');
 
         console.log(addColor("[Recorder] Stop recording buffer", "cyan"));
         clearInterval(this._bufferRestartIntervalID);
+        this._bufferRestartIntervalID = undefined;
```
ericlytle, line 137: i know why you are doing this, but this really seems odd to me and kind of a hacky way to accomplish what we are wanting. 
can we not rely on the outcome of clearInterval and use that?

if we cant, do we really need clearInterval call then if we are setting the value to undefined immediately following. 

all these comments are based on my 2 min google search for what clearInterval does. so please correct me if im way off

tdaugaard: It's not actuallt necessary to do this. 
image

tdaugaard: It's not necessary to clear the timer ID, no; it references nothing when it's been cleared, even if it still has a value.

aza547: Thanks. Agree I can just delete this line.

aza547: Hm actually I'm less sure on second thought. Are you suggesting I do:

`this._bufferRestartIntervalID = clearInterval(this._bufferRestartIntervalID);` 

I'm relying on this for the guard elsewhere

tdaugaard: I saw that you're testing it later for not being undefined. I wouldn't personally use a timer variable as a guard . I think I've done something similar elsewhere, where after clearing the interval I clear the guard variable.

I guess you _could_ double-use it by setting it to undefined - it doesn't hurt in that case.

aza547: OK, going to keep it as it was initially then. Prefer that I think than having another variable.

PR #132, opened 2022-09-20
Fix timer for Iron Docks

Emme:
Sadly it's still only 30 min


tdaugaard (approved):
That is true, apparently. I had taken the timers from https://mythicpl.us/, but they are wrong it seems.

aza547:
Can we get a changelog entry here? User facing change.

aza547:
Just gonna add the changelog on main and merge this on second thoughts, one less open PR to look at!

PR #135, opened 2022-09-21
Add a button to manually stop recording

tdaugaard:
Sometimes, the app doesn't stop recording on its own for various reasons so this change introduces a way to stop recording manually.

Whenever the app is recording and the "REC" icon is flashing, clicking on that icon will change it to a stop icon and clicking it will ask for confirmation to do so.

In the event of a force stop, the following will happen to attempt to clean up:

  - Current recording `duration` is set to `<current time> - videoStartDate`
  - Combatant map and player is cleared
  - Any active challenge mode object is cleared
  - Any active test is cleared
  - Recorder will be stopped with whatever current metadata is present

`runRecordingTest` has been modified as such:

  - It now allows for testing the above functionality by holding down Ctrl and Alt while clicking the test button. This causes the test to never end on its own and requires the use of the force stop button.

  - Timestamp for log lines in this function wil be set on-the-fly to match current date and time + an optional adjustment in seconds to avoid confusing the app when the start date is too far in the past.

App CSS has been changed slightly to accomodate the new `<ConfirmationDialog/>` component:

  - `h2` is no longer styled for `.MuiDialogTitle-root` elements
  - `.MuiDialogTitle-root` elements no longer has the styling from `.MuiTypography-root`.

This was to make it looks not great, but also not terrible.


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Manual stop recording button, on the 'recording' status icon.
```
aza547, line 10: https://github.com/aza547/wow-recorder/issues/142 raised so you can link it to an issue.

src/main/logutils.ts
```diff
@@ -880,15 +901,27 @@ const runRecordingTest = () => {
...
-    const testArenaSpellLine = "8/3 22:09:59.365  SPELL_AURA_APPLIED,Player-1084-08A89569,\"Alexsmite-TarrenMill\",0x511,0x0,Player-1084-08A89569,\"Alexsmite-TarrenMill\",0x511,0x0,110310,\"Dampening\",0x1,DEBUFF";
-    const testArenaStopLine = "8/3 22:12:14.889  ARENA_MATCH_END,0,8,1673,1668";
+    // This inserts a test date so that the recorder doesn't confuse itself with
```
aza547, line 904: Comment a bit misleading - I think this actually just gets a date object relative to the current time + `seconds `seconds?

src/main/logutils.ts
```diff
@@ -880,15 +901,27 @@ const runRecordingTest = () => {
...
+    const testArenaSpellLine = startDate + "  SPELL_AURA_APPLIED,Player-1084-08A89569,\"Alexsmite-TarrenMill\",0x511,0x0,Player-1084-08A89569,\"Alexsmite-TarrenMill\",0x511,0x0,110310,\"Dampening\",0x1,DEBUFF";
+    const testArenaStopLine = getAdjustedDate(2 * 60) + "  ARENA_MATCH_END,0,8,1673,1668";
```
aza547, line 915: endDate variable as per startDate probably a bit cleaner to read than this

src/main/logutils.ts
```diff
@@ -818,6 +818,24 @@ const watchLogs = (logdir: any) => {
     return [name, realm];
 }
 
+/**
```
aza547, line 821: Feels like this should be in recorder.ts, but I think it's a bit tricky because of the metadata being a variable in logutils so not proposing you do anything about it now. Something to keep in mind for us in the future.

aza547 (changes requested):
Love it. Minor comments only. Had a test too, and seems great.

src/main/logutils.ts
tdaugaard: There's actually a lot of things I wanted to move around but due to where some data is located (like `metadata`) it isn't feasible for now - _rest assured_ I will restructure more stuff along the way ;)

PR #136, opened 2022-09-21
Added async load of videos for responsiveness

tdaugaard:
Application is somewhat slow to start up, especially with storage on mechanical drives, if there are many videos.

This change simply makes the loading of videos asynchronous.


aza547:
Also seems like a great change. No major comments - hopefully me hurrying reviews a bit more isn't going to bite me 😄

PR #138, opened 2022-09-21
Refactor: Log parsing performance

tdaugaard:
### Important!

This PR **requires** merging of #126 first, as it it built on top of that.

Either merge this in its entirety and close #126 unmerged, or merge #126 first, then this one.

This PR might clash with some of the more recent PRs from me, but I will make sure to fix any conflicts once the PRs start being merged.

The changes for this PR starts from and including commit 3e74a8fa

### Support for multiple log files

These changes enable the monitoring and processing of combat events from mulitple combat logs, which enables support for the various flavours of WoW Classic. This is achieved by monitoring several directories and whenever one of the combat logs receive data, the processing handle will only process data from that log file until a timeout is reached, if data stops coming in.

If data stops being received and the `DataTimeout` event is triggered, a cleanup process can be performed to reset the state of the recording/recorder.

### Merge `LogLine` and `splitLogLine`

Refactored `LogLine` and `splitLogLine()` for performance reasons, to avoid parsing arguments of a combat log line that we're not interested in.

Many combat log events have arguments that we don't use and there's no point in parsing them all.

This change primarily ensures that we parse on-demand.

Notable changes:

- `LogLine.args[n]` has been replaced by `LogLine.arg(n)`

- `LogLine` now takes the original combat line as its only constructor argument.

- `handleLogLine()` no longer calls `splitLogLine()` twice since that isn't necessary anymore.

### Refactor/rewrite the combat parser logic

The combat log parser, watcher, tail'er, and what have we has been refactored and some rewritten to support monitoring multiple log files simultaneously.

- `LogLine.parseLogArg()` has had its `continue` statements in the `switch (c)` block replaced by `break`, followed by a single `continue` at the end because TypeScript considered `continue` as a fall-through.

- `LogLine` has a new method, `addArg()` that simply facilitates adding a parsed argument to `this.args`.

New `CombatLogParser` class that replaces all previous logic related to the parsing of, discovery of, and tailing of combat log files that was in `logutils.ts`.

This class is instantiated with a single option, a data timeout. This timeout is used to fire an event if no data has been received from the active combat log, which can be used to stop recording or inform the user.

The class extends `EventEmitter` and therefore emits combat log events as JavaScript events as we know them best. All combat log handlers are now attached as event listeners using `.on('....', callback)` as seen in `logutils.ts`. All event listerners will receive the full `LogLine` as well as a 2nd argument, `flavour`, which will indicate which flavour of WoW the combatlog was parsed from.

Whenever data is received on a specific combat log, that log will have exclusive data processing for up to `dataTimeout` seconds such that the recorder will not receive events from two combat logs from differente flavours of WoW at the same time.


aza547:
#126 is merged, is this ready for review now?

tdaugaard:
Closing because the branch is a bit old; reposting a new PR rebased on main.

PR #145, opened 2022-09-21
Remember the selected category across restarts

tdaugaard:
Whenever a new category is selected, it'll be saved to the `selected-category` persistent setting and restored when the app restarts.


aza547:
Conflicts, otherwise looks good!

Issue #134, opened 2022-09-21
A Javascript error occurred


ddechant:
I get the follow error after running the application while in a raid instance. Logging was already enabled.

![image](https://user-images.githubusercontent.com/263218/191399700-163742d8-c6cf-4e6a-a075-1f39ecca460b.png)

tdaugaard:
Would you happen to have the combat log that caused this and would you be willing to share it either here or in a DM on Discord so I can replay it and see if I can reproduce the error?

ddechant:
Sent you the combat logs in discord.

tdaugaard:
Ah, I see! Thanks a lot - a player died before the raid encounter started and because of that, it was trying to access a timestamp of then the video started recording, but that hadn't been set yet - only set when something starts.

The fix is of course to only handle unit deaths when something is in progress.

aza547:
Fixed in next release.

PR #147, opened 2022-09-22
Fix UNIT_DIED handling before recording starts

tdaugaard:
In the case of raids, it's not unusual for a player to die _before_ an encounter starts for various reasons (e.g. trash) but as we don't have a `videoStartDate` before the recording/encounter starts, this caused the application to crash due to trying to access `videStartDate.getTime()`.

This fix simply guards `handleUnitDiedLine()` in such a way where if we are not recording or if we don't have a valid `videoStartdate`, we'll ignore `UNIT_DIED`.

Fixes issue #134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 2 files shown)


tdaugaard:
This one is quite important for the people recording raids as it will crash on _any_ `UNIT_DIED` event outside of an encounter.

PR #148, opened 2022-09-22
Name video files more sensibly

### LINKED ISSUE #52: Name videos more sensibly
Never worked out how to change the output file name in OSN / OBS. 

Would be good to for a bunch of reasons.

tdaugaard:
This adds an easy and scalable way to find a raid name from an encounter ID; `raidInstances` is an array of raids with zone ID, name, and a list of encounters (by encounter ID).

Hashmap `instanceEncountersById` has been added to have a single point of truth regarding encounter IDs to encounter name for all instances that support this (dungeons, raids).

A new file has been added, `helpers.ts`, which should be kept FREE from any filesystem related modules (as these cannot be imported in the renderer process) and some of the helper functions from both `utils.ts` as well as `renderutils.ts` has been moved here, as they were needed in multiple places.

This file utilizes the format of `export const methodName = () => {};` as it avoids having a long list of exported names at the bottom, and makes it easy to identify which methods are exported and which aren't.

Furthermore, it vastly improves renaming symbols across the entire codebase because it isn't just changed to `export { newName as oldName }`.

The `Recorder` class will now generate a suitable filename for the final video file based on the given metadata to name the final video files a bit more sensibly.

The format that was chosen results in file names like:

  - `2022-09-22 15-36-33 - {2v2} Enigma Crucible (Win)`
  - `2022-09-22 16-15-54 - {Raids} Sepulcher of the First Ones, Lihuvim (Wipe)`
  - `2022-09-22 15-44-19 - {Mythic+} Mechagon Junkyard +18 (+1)`

So, basically: `{timestamp} - {category} {Zone/Instance}` + some category specific information as seen above, whenever possible.

Replaced all occurances of `{ [id: number]: string; }` with `NumberKeyToStringValueMapType` in `constants.ts`.

Closes #52

### DIFF
--- a/src/main/helpers.ts
+++ b/src/main/helpers.ts
@@ -1,0 +1,116 @@
+import { dungeonsByMapId, instanceDifficulty, InstanceDifficultyType, instanceEncountersById, instanceNamesByZoneId, raidInstances, VideoCategory, zones } from "./constants";
+import { Metadata } from "./logutils";
+import { RaidInstanceType } from "./types";
+
+/**
+ * Get a result text appropriate for the video category that signifies a
+ * win or a loss, of some sort.
+ */
+export const getVideoResultText = (category: VideoCategory, isGoodResult: boolean): string => {
+
+    // Not sure how we can decide who won or lost yet.
+    // Combat log doesn't make it obvious.
+    if (category == VideoCategory.Battlegrounds || category == VideoCategory.SoloShuffle) {
+        return "";
+    }
+
+    switch (category) {
+        case VideoCategory.MythicPlus:
+            return isGoodResult ? "Time" : "Depl";
+
+        case VideoCategory.Raids:
+            return isGoodResult ? "Kill" : "Wipe";
+
+        default:
+            return isGoodResult ? "Win" : "Loss";
+    }
+};
+
+/**
+ * Get the name of a dungeon by its map ID
+ */
+export const getDungeonByMapId = (mapId?: number): string => {
+    if (mapId && dungeonsByMapId.hasOwnProperty(mapId)) {
+        return dungeonsByMapId[mapId];
+    }
+
+    return 'Unknown Dungeon';
+};
+
+/**
+ * Get the name of a boss encounter based on its encounter ID
+ */
+export const getEncounterNameById = (encounterId? : number): string => {
+    if (encounterId && instanceEncountersById.hasOwnProperty(encounterId)) {
+        return instanceEncountersById[encounterId];
+    }
+
+    return 'Unknown Boss';
+};
+
+/**
+ * Get the name of a zone in WoW based on its zone ID
+ */
+export const getInstanceNameByZoneId = (zoneId?: number): string => {
+    if (zoneId && instanceNamesByZoneId.hasOwnProperty(zoneId)) {
+        return instanceNamesByZoneId[zoneId];
+    }
+
+    return 'Unknown Zone';
+};
+
+/**
+ * Get the difficulty of an instance based on its difficulty ID, as found in
+ * `ENCOUNTER_START` log lines.
+ */
+export const getInstanceDifficulty = (difficultyId: number): InstanceDifficultyType | null => {
+    if (instanceDifficulty.hasOwnProperty(difficultyId)) {
+        return instanceDifficulty[difficultyId];
+    }
+
+    return null;
+};
+
+/**
+ * Get the zone name.
+ */
+export const getVideoZone = (metadata: Metadata) => {
+    const zoneID = metadata.zoneID;
+    const encounterID = metadata.encounterID;
+
+    switch (metadata.category) {
+        case VideoCategory.MythicPlus:
+            return getInstanceNameByZoneId(zoneID);
+
+        case VideoCategory.Raids:
+            return getRaidNameByEncounterId(encounterID);
+
+        default:
+            if (zoneID && zones.hasOwnProperty(zoneID)) {
+                return zones[zoneID];
+            }
+    }
+
+    return "Unknown Zone";
+}
+
+/**
+ * Get the raid name from the encounter ID.
+ */
+export const getRaidNameByEncounterId = (encounterID?: number) => {
+    const raid = getRaidByEncounterId(encounterID);
+    if (!raid) {
+        return 'Unknown Raid';
+    }
+
+    return raid.name;
+}
+
+/**
+ * Get the raid instance from an encounter ID.
+ */
+export const getRaidByEncounterId = (zoneID?: number): RaidInstanceType | undefined => {
+    const raid = raidInstances.filter(r => zoneID && r.encounters.hasOwnProperty(zoneID))
+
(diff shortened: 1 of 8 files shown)


src/main/helpers.ts
```diff
@@ -0,0 +1,116 @@
+import { dungeonsByMapId, instanceDifficulty, InstanceDifficultyType, instanceEncountersById, instanceNamesByZoneId, raidInstances, VideoCategory, zones } from "./constants";
+import { Metadata } from "./logutils";
+import { RaidInstanceType } from "./types";
+
```
aza547, line 4: Can we get a comment in here to be explicit about no filesystem stuff in this file.

src/main/helpers.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Get a result text appropriate for the video category that signifies a
+ * win or a loss, of some sort.
+ */
+export const getVideoResultText = (category: VideoCategory, isGoodResult: boolean): string => {
+
+    // Not sure how we can decide who won or lost yet.
```
aza547, line 11: This probably needs an update. "Non-trivial to determine who won a BG/SoloShuffle so just don't report it" -- something like that

src/main/recorder.ts
```diff
@@ -280,6 +289,52 @@ type RecorderOptionsType = {
...
+     * to identify the video files on the filesystem and be able to tell what it
+     * was about and the result of it.
+     */
+    getFinalVideoFilename (metadata: Metadata): string | undefined {
+        let outputFilename = '';
+        let zoneName: string;
+        const resultText = getVideoResultText(metadata.category, metadata.result);
+
+        switch (metadata.category) {
+            case VideoCategory.TwoVTwo:
```
aza547, line 307: Can we use default to avoid this pattern of multiple cases in a row

src/main/recorder.ts
```diff
@@ -280,6 +289,52 @@ type RecorderOptionsType = {
...
+                    outputFilename = `${getDungeonByMapId(cm.mapId)} +${cm.level} (${resultText})`;
+                }
+            break;
+        }
+
+        if (!outputFilename) {
+            return;
+        }
+
+        return `{${metadata.category}} ${outputFilename}`;
```
aza547, line 336: Not a big fan of curly braces making it to something user visible. Think it would be better to be:

```
2022-09-22 20-33-47 - 2v2 - Enigma Crucible (Win)
2022-09-22 20-30-39 - Mythic+ - Mechagon Junkyard +18 (+1)
```
Spaces in file names is weird to me as well but I think that's just business as usual on windows so I'll get over it!

src/main/util.ts
```diff
@@ -453,10 +421,16 @@ const runSizeMonitor = async (storageDir: string, maxStorageGB: number): Promise
...
+const cutVideo = async (
+        initialFile: string,
+        finalDir: string,
+        outputFilename: string | undefined,
+        desiredDuration: number
+    ): Promise<string> => {
     
     const videoFileName = path.basename(initialFile, '.mp4');
-    const finalVideoPath = path.join(finalDir, videoFileName + ".mp4");
+    const baseVideoFilename = sanitizeFilename(videoFileName + (outputFilename ? ' - ' + outputFilename : ''));
```
aza547, line 432: Doing this inline is a bit hard to read. Can we bring it out?

```
if (outputFilename) {
  outputFilename = " - " + outputFilename;
} else {
  outputFilename = "";
}

const baseVideoFilename = sanitizeFilename(videoFileName + outputFilename);
```

src/main/recorder.ts
```diff
@@ -280,6 +289,52 @@ type RecorderOptionsType = {
...
+        let zoneName: string;
+        const resultText = getVideoResultText(metadata.category, metadata.result);
+
+        switch (metadata.category) {
+            case VideoCategory.TwoVTwo:
+            case VideoCategory.ThreeVThree:
+            case VideoCategory.Skirmish:
+            case VideoCategory.SoloShuffle:
+                zoneName = getInstanceNameByZoneId(metadata.zoneID);
+                outputFilename = `${zoneName} (${resultText})`;
```
aza547, line 312: No resultText for SoloShuffles
Also BGs, but they are missed out of this function, I think by mistake?

src/main/recorder.ts
```diff
@@ -280,6 +289,52 @@ type RecorderOptionsType = {
...
+            case VideoCategory.Raids:
+                const encounterName = getEncounterNameById(metadata.encounterID);
+                zoneName = getRaidNameByEncounterId(metadata.encounterID);
+                outputFilename = `${zoneName}, ${encounterName} (${resultText})`;
+            break;
+
+            case VideoCategory.MythicPlus:
+                if (metadata.challengeMode) {
+                    const cm = metadata.challengeMode;
+                    const keystoneUpgradeLevel = ChallengeModeDungeon.calculateKeystoneUpgradeLevel(cm.allottedTime, cm.duration);;
```
aza547, line 324: This case could probably do with its own helper function `getChallengeModeResultText` or so, there is a bit much packed in here for my liking. VideoButton could also benefit from sharing it?

src/main/util.ts
```diff
@@ -442,6 +398,18 @@ const runSizeMonitor = async (storageDir: string, maxStorageGB: number): Promise
     writeMetadataFile(videoPath, metadata);
 }
 
+/**
```
aza547, line 401: Love this!

aza547 (changes requested):
Few minor comments, looks great.

src/main/recorder.ts
tdaugaard: Spaces in filenames is totally normal on Windows, yeah :D Curly braces look bad, agreed.

src/main/recorder.ts
tdaugaard: We can - I know I did that somewhere else, but for this one I just felt it made more sense to have the explicitly stated to make it quickly obvious what was being handled how, but ultimately I don't really care either way =)

src/main/recorder.ts
tdaugaard: Spaces in file names are here to stay :p Curly braces looks bad, I agree.

src/main/recorder.ts
tdaugaard: VideoButton wouldn't benefit from this, as it's only used for the filename, really. It's not so much result text (which we have already "_Depl_" and "_Time_") as it is a condensed string of information for the given video.

I can surely give it its own method!

src/main/recorder.ts
tdaugaard: Most likely because I know nothing about PvP. I made a change now where, if `resultText` is empty it won't add it to the filename - will that satisfy SoloShuffles and BGs?

aza547: Sounds good. The problem with solo shuffle is that there are 6 round and you can win any number of that. BGs there is just no indication in logs who wins. Currently in either case we don't display a result.

Issue #139, opened 2022-09-21
Spontaneously and quietly we stop asserting OBS signalling 


aza547:
![image](https://user-images.githubusercontent.com/32389325/191578769-b671fcb0-dcef-43d0-980f-0ce6fb4e4d68.png)
[WarcraftRecorder (1).log](https://github.com/aza547/wow-recorder/files/9619253/WarcraftRecorder.1.log)

Emme:
The OSN log:
[2022-09-20 19-44-54.txt](https://github.com/aza547/wow-recorder/files/9619292/2022-09-20.19-44-54.txt)

aza547:
This is interesting:

```
[000:03:17:45.323.033.200][12688][Warning] [AMF] <Id: 11> No output Packet, encoder is overloaded!
```

Annoyingly I think the OSN timestamps are from start-up time, we can probably cross reference to see if it matches up.

aza547:
I think we handle the AMF encoder pretty badly actually, this might be a product of that. 

See obsRecorder.ts:

```
  if (selectedEncoder.toLowerCase().includes("amf")) {
    // For AMF encoders, can't set 'lossless' bitrate.
    // It interprets it as zero and fails to start.
    // See https://github.com/aza547/wow-recorder/issues/40.
    setSetting('Output', 'Recbitrate', 50000);
  }
```

aza547:
and #40 as per the comment

aza547:
Looks like an interesting read https://obsproject.com/forum/threads/recent-encoding-overloaded-error-with-amf-no-output-packet.124033/

aza547:
Maybe not on second thoughts as I think this was not over a long recording but between a buffer restart. Will come back to this.

tdaugaard:
Would it make sense to expose the available encoders in the Settings window to allow people to pick in case one isn't optimal?

aza547:
It's an interesting question - I don't know is the answer! 

On one hand, if things go wrong it would be great to say "try the other encoder" - on the other hand, it's probably confusing to most users. 

I think I want to expose this, but when I revamp the settings page and include an "advanced tab" or something.

aza547:
[...]

PR #149, opened 2022-09-22
Issue 139

aza547:
See #139.

The main goal here is to reject a promise (which we then throw an error on due to the listeners in main.ts) if OBS doesn't signal within 5 seconds.

### DIFF
--- a/src/main/obsRecorder.ts
+++ b/src/main/obsRecorder.ts
@@ -248,7 +248,12 @@ const start = async () => {
   console.log("[OBS] obsRecorder: start");
   osn.NodeObs.OBS_service_startRecording();
 
-  let signalInfo = await waitQueue.shift();
+  // Reject the promise if we don't get a response in time.
+  let signalInfo = await Promise.race([
+    waitQueue.shift(), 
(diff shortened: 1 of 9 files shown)


src/main/obsRecorder.ts
```diff
@@ -259,13 +264,28 @@ const stop = async () => {
   console.log("[OBS] obsRecorder: stop");
   osn.NodeObs.OBS_service_stopRecording();
 
-  let signalInfo = await waitQueue.shift();
+  // Reject the promise if we don't get a response in time.
```
tdaugaard, line 267: I think these 3 could probably be a `for` loop or something as they are identical, except for the log message
Perhaps:

```typescript
['stopping', 'stop', 'wrote'].forEach(signal => {
  // Reject the promise if we don't get a response in time.
  signalInfo = await Promise.race([
    waitQueue.shift(), 
    new Promise((_r, rej) => setTimeout(rej, 5000))
  ]);

  assertSignal(signalInfo, "recording", signal);
});
```

aza547: Have simplified this a lot, opted not for a loop in the end.

PR #150, opened 2022-09-22
Make buffer storage path configurable

### LINKED ISSUE #124: Make buffering dir configurable
Someone asked on reddit post for this to avoid bogging down their NIC as they are using a NAS device.

tdaugaard:
`bufferStoragePath` defaults to the value of `${storageDir}/.temp` in the Settings window, but this is only visual - it doesn't save that value anywhere. It's just to show people that it doesn't need to be set.

A guard has been put in place in `saveSettings()` to PREVENT Settings from being saved if `storagePath` and `bufferStoragePath` have the same value as this is an invalid configuration and would lead to LOSS of video files when `cleanupBuffer()` runs.

Merged `openStoragePathDialog()` and `openLogPathDialog()` into a new method, `openDirectorySelectorDialog()`, which is generic and takes a settings key for the path we're selecting.

Renamed `ConfirmationDialog` => `InformationDialog` and made it a generic dialog that can be used to display whatever type of information, confirmation, question, and what have we.

Closes #124


src/settings/Settings.tsx
```diff
@@ -123,22 +153,39 @@ export default function Settings() {
     ]
   };
 
+  const bufferStoragePathPlaceholder = state.bufferStoragePath ? state.bufferStoragePath : "(Default)";
```
aza547, line 156: Can we make this default to the real default path rather than "(Default)"?

```
const bufferStoragePathPlaceholder = state.bufferStoragePath ? state.bufferStoragePath : state.storagePath + "/.temp";
```

src/settings/Settings.tsx
```diff
@@ -123,22 +153,39 @@ export default function Settings() {
...
           <div className="card-body">
             <div className="row gutters">
               <div className="col-xl-6 col-lg-6 col-md-6 col-sm-6 col-12">
                 <div className="form-group">
                   <label> Storage Path </label>
-                  <input type="text" className="form-control" id="storage-path" placeholder={state.storagePath} onClick={openStoragePathDialog}/>
+                  <input type="text" className="form-control" id="storage-path" placeholder={state.storagePath} onClick={() => openDirectorySelectorDialog('storagePath')}/>
+                </div>
+              </div>
+              <div className="col-xl-6 col-lg-6 col-md-6 col-sm-6 col-12">
```
aza547, line 179: I think this is too much power for a regular user to be able to set this - not suggesting you change this but I think it has potential to be confusing.

Will be good to move to an advanced settings tab. Hopefully I can find time to re-do the settings UI a bit this weekend.

src/main/main.ts
```diff
@@ -47,10 +47,12 @@ console.log("[Main] App starting: version", app.getVersion());
  */
 const loadRecorderOptions = (cfg: ElectronStore): RecorderOptionsType => {
   const storageDir = getPathConfigSafe(cfg, 'storage-path');
+  const bufferStorageDir = getPathConfigSafe(cfg, 'buffer-storage-path');
 
   const config = {
     storageDir: storageDir,
-    bufferStorageDir: path.join(storageDir, ".temp"),
+    // Default value for buffer storage is '{storageDir}/.temp'
```
aza547, line 54: comment is a bit redundant - code is self documenting 😉

src/renderer/InformationDialog.tsx
```diff
@@ -0,0 +1,79 @@
+import * as React from 'react';
```
aza547, line 1: This is really nice. Probably loads we can do with this.

aza547 (changes requested):
Had a problem while testing this that it took an app restart to apply the buffer config change. 

Wasn't immediately obvious why, did a bit of debugging and it did seem to be getting loaded correctly by `loadRecorderOptions`. 

I tried to use debug mode and weird it worked fine with that. Maybe something weird like a timing thing and we restart the recorder to quickly? I don't see how this can be new in this PR though. Might be worth testing on main on storagePath change.

src/settings/Settings.tsx
tdaugaard: I actually did that at first, but I don't like having to hard-code the _implied_ default on the settings page because it means it has to be updated more than one place should it ever change.

I guess I could make a helper function like `getDefaultBufferStoragePath()`  - I'd be fine with that.

src/settings/Settings.tsx
tdaugaard: I was under the impression that this was a thing we'd want now? You don't have to mege this now if you don't want this setting in the app at this point. If you'd rather wait until the Settings has been revamped with an _Advanced_ tab, that's fine by me - I can keep the branch alive and make a new PR when the time is right - what do think?

tdaugaard:
> I tried to use debug mode and weird it worked fine with that. Maybe something weird like a timing thing and we restart the recorder to quickly? I don't see how this can be new in this PR though. Might be worth testing on main on storagePath change.

Oh, really? I actually didn't notice! I'll do some more testing and see if I can find the culprit.

src/settings/Settings.tsx
aza547: Nah lets get it in, less PRs hanging over my head 😄

src/settings/Settings.tsx
aza547: Not totally sure I'm with you here. If you change as above, doesn't that pass down to the settings page?

aza547: Oh, think I understand - personally OK with it being in two places. If we ever update it and forget to change it here then worst case is that it will look a bit odd in the settings but no functional impact. Perhaps add a comment to both locations to say "also see here" to avoid that?

aza547: Alternatively what you are suggesting with a `getDefaultBufferStoragePath` func sounds great too.

src/main/main.ts
```diff
@@ -4,8 +4,20 @@
...
+  isConfigReady,
+  deleteVideo,
+  openSystemExplorer,
+  toggleVideoProtected,
+  fixPathWhenPackaged,
+  defaultMonitorIndex,
+  defaultMinEncounterDuration,
+  defaultAudioDevice,
+  getAvailableDisplays,
+} from './util';
```
tdaugaard, line 20: I split this up into multiple lines because it was getting way too long.

aza547:
We agreed to go with this despite the buffer path needing a restart of the monitor to take affect. I've opened an issue so we won't forget, but it's not a regression as it already occurs with the main storagePath.

PR #151, opened 2022-09-23
Feature: Better monitor selector

### LINKED ISSUE #81: Do something better in UI for monitor selection
image

tdaugaard:
### Ensure we save all settings correctly

`getAttribute()` returns `null`, specifically, when an element does not have the given attribute but the logic wasa simply testing the `value` as being 'truthy', which meant that `0` or even an empty string wouldn't be saved.

Monitor indexes are zero based and therefore we couldn't save monitor index `0`. We can now!

### Add methods for providing better monitor selection

Created a new method for retrieving the available monitors from Electron (which is the same as being provided from OBS, but with types!) and creates a new array for the frontend to display to make the selection easier.

### Add frontend selector for monitors

Replace the input text field with a dropdown selector box that displays the available monitors in a slightly better way which should hopefully make it easier to distinguish the individual monitors of ones system.

Example:

image

Closes #81

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -258,7 +258,7 @@ const createSettingsWindow = async () => {
 
   settingsWindow = new BrowserWindow({
     show: false,
-    width: 600,
+    width: 770,
     height: 500,
     resizable: (process.env.NODE_ENV === 'production') ? false : true,
     icon: getAssetPath('./icon/settings-icon.svg'),
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -5,7 +5,7 @@
  */
 import path from 'path';
 import { app, BrowserWindow, shell, ipcMain, dialog, Tray, Menu, net, screen, Display } from 'electron';
-import { resolveHtmlPath, loadAllVideos, isConfigReady, deleteVideo, openSystemExplorer, toggleVideoProtected, fixPathWhenPackaged, getPathConfigSafe, getNumberConfigSafe, defaultMonitorIndex, defaultMinEncounterDuration, getStringConfigSafe, defaultAudioDevice } from './util';
+import { resolveHtmlPath, loadAllVideos, isConfigReady, deleteVideo, openSystemExplorer, toggleVideoProtected, fixPathWhenPackaged, getPathConfigSafe, getNumberConfigSafe, defaultMonitorIndex, defaultMinEncounterDuration, getStringConfigSafe, defaultAudioDevice, getAvailableDisplays } from './util';
 import { watchLogs, pollWowProcess, runRecordingTest, forceStopRecording } from './logutils';
 import Store from 'electron-store';
 const obsRecorder = require('./obsRecorder');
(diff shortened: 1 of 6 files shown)


tdaugaard:
Improved it with some physical position stuff:
image

There's a bug regarding monitor index 0 that I'm working on.

tdaugaard:
Final result. Displays are numbered from 1 because `null`, `Nan`, `undefined`, _and_ `0` are all considered _false_.

image

src/main/types.ts
```diff
@@ -1,3 +1,5 @@
+import { Rectangle, Size } from "electron";
```
aza547, line 1: Rectangle unused?

src/main/main.ts
```diff
@@ -4,15 +4,14 @@
...
+import { app, BrowserWindow, shell, ipcMain, dialog, Tray, Menu, net, screen, Display } from 'electron';
+import { resolveHtmlPath, loadAllVideos, isConfigReady, deleteVideo, openSystemExplorer, toggleVideoProtected, fixPathWhenPackaged, getPathConfigSafe, getNumberConfigSafe, defaultMonitorIndex, defaultMinEncounterDuration, getStringConfigSafe, defaultAudioDevice, getAvailableDisplays } from './util';
 import { watchLogs, pollWowProcess, runRecordingTest, forceStopRecording } from './logutils';
 import Store from 'electron-store';
 const obsRecorder = require('./obsRecorder');
 import { Recorder, RecorderOptionsType } from './recorder';
 import { getAvailableAudioInputDevices, getAvailableAudioOutputDevices } from './obsAudioDeviceUtils';
-import { AppStatus, VideoPlayerSettings } from './types';
-import { net } from 'electron';
+import { AppStatus, OurDisplayType, VideoPlayerSettings } from './types';
```
aza547, line 14: OurDisplayType unused as far as I can tell

src/main/util.ts
```diff
@@ -581,6 +582,69 @@ const isNumberClose = (val: number, compare: number) => {
...
+                id: display.id,
+                index: displayIndex,
+                physicalPosition: getDisplayPhysicalPosition(numberOfMonitors, index),
+                primary: isPrimary,
+                displayFrequency: display.displayFrequency,
+                depthPerComponent: display.depthPerComponent,
+                size: display.size,
+                scaleFactor: display.scaleFactor,
+                aspectRatio: width / height,
+                physicalSize: {
```
aza547, line 638: This makes me a bit nervous. How confident are you that we're not regressing recent fixes in this area to people who are unlucky with the rounding of their resolutions?

aza547 (commented):
Looks great. Had a good test and didn't spot anything amiss.

src/main/util.ts
tdaugaard: Because this is the correct way to do this. 2560.5 (like mine) becomes 2560, which is accurate. I'd be happy to make a test of various resolutions and scaling factors and apply the logic to it with a table to show the results.

aza547: What makes floor the correct thing rather than ceil? 

What if someone someone's rounding gives then 2559.5 which becomes 2559?

tdaugaard: Because of the examples I've seen, it'll _never_ be lower than the actual resolution - however, I'm working on a fix that - instead of checking if a number "is close", it will simply find the closest resolution to whatever is given. That'll sort it out once and for all - and I can test it because the user who reported this issue posted his display settings.

tdaugaard: Also, I didn't remove the `isNumberClose()` logic, which is "the heart" of this - but really, I can remove `Math.floor()` - it doesn't really matter _to me_.

tdaugaard: Please see #154 before considering this.

tdaugaard: Regression crisis averted! Since #154 was merged, you may re-consider this PR.

PR #156, opened 2022-09-24
Refactor: Log parsing performance & multiple log paths

### LINKED ISSUE #155: Watchlogs should be async
I've noticed this myself that dragging the window around has a minor lag spike every second.

Pretty sure that's the timer we look for logs on and I suspect its synchronous. 

We should make it async to improve app responsiveness.

tdaugaard:
### Merge `LogLine` and `splitLogLine`

Refactored `LogLine` and `splitLogLine()` for performance reasons, to avoid parsing arguments of a combat log line that we're not interested in.

Many combat log events have arguments that we don't use and there's no point in parsing them all.

This change primarily ensures that we parse on-demand.

Notable changes:

- `LogLine.args[n]` has been replaced by `LogLine.arg(n)`

- `LogLine` now takes the original combat line as its only constructor argument.

- `handleLogLine()` no longer calls `splitLogLine()` twice since that isn't necessary anymore.

### Refactor/rewrite the combat parser logic

The combat log parser, watcher, tail'er, and what have we has been refactored and some rewritten to support monitoring multiple log files simultaneously.

- `LogLine.parseLogArg()` has had its `continue` statements in the `switch (c)` block replaced by `break`, followed by a single `continue` at the end because TypeScript considered `continue` as a fall-through.

- `LogLine` has a new method, `addArg()` that simply facilitates adding a parsed argument to `this.args`.

New `CombatLogParser` class that replaces all previous logic related to the parsing of, discovery of, and tailing of combat log files that was in `logutils.ts`.

This class is instantiated with two options:

- A data timeout: This timeout is used to fire an event if no data has been received from the active combat log, which can be used to stop recording or inform the user.
- A function that can find files.

The class extends `EventEmitter` and therefore emits combat log events as JavaScript events as we know them best. All combat log handlers are now attached as event listeners using `.on('....', callback)` as seen in `logutils.ts`. All event listerners will receive the full `LogLine` as well as a 2nd argument, `flavour`, which will indicate which flavour of WoW the combatlog was parsed from.

Whenever data is received on a specific combat log, that log will have exclusive data processing for up to `dataTimeout` seconds such that the recorder will not receive events from two combat logs from differente flavours of WoW at the same time.

### Support for multiple log files

These changes enable the monitoring and processing of combat events from mulitple combat logs, which enables support for the various flavours of WoW Classic. This is achieved by monitoring several directories and whenever one of the combat logs receive data, the processing handle will only process data from that log file until a timeout is reached, if data stops coming in.

If data stops being received and the `DataTimeout` event is triggered, a cleanup process is performed to reset the state of the recording/recorder.

Fixes #155


src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+            .split("\n");
+
+        return content.length > 1 ? content[1] : 'unknown';
+    }
+
+    /**
+     * Ensure only a single logfile is being watched once one of them starts
+     * receiving data.
+     *
+     * The lock will timeout after 30 seconds if no data has been received.
```
aza547, line 322: Interesting. I think I was expecting the lock to be released by some other timeout. Might discover this as I read on, but are we stopping the recording if a lock times out? We probably should be. Is 30s a reasonable time for that? We might need some real data here to be sure. I could imagine a scenario where some M+ group takes a toilet break (they are casuals for arguments sake) and their M+ combat log lock is released.

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+        }, this._options.dataTimeout);
+    }
+
+    /**
+     * Find and return the most recent file that matches the combat log filename
+     * pattern.
+     */
+    private getLatestLog (pathSpec: string): string | undefined {
+        const globPath = path.join(pathSpec, 'WoWCombatLog*.txt');
+        const logs = glob.sync(globPath)
```
aza547, line 368: Did we not have a "get most recent file" function somewhere else now? Can we share this? 

Ah it's specific for mp4s, but perhaps we can adapt it to take an extension input? See util.ts `getSortedVideos`

If that's not trivial, then don't worry. 

I'd also love to make this async, see later comment and #155

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+        tailHandler
+            .on('line', (line: string) => {
+                if (!this.lockHandler(handler)) {
+                    return;
+                }
+
+                this.handleLogLine(handler.wowFlavour, line);
+            })
+            .on('error', (error: unknown) => {
+                console.error('[CombatLogParser] ERROR: ', error);
```
aza547, line 399: Don't need the "ERROR" in here as it's already an error level log - but worth some text giving the context of "tailing log file"

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    /**
+     * Watch a directory for combat logs and restart the monitoring
+     * mechanism if there's a new one detected.
+     */
+    private watchLogDirectory(path: string): void {
+        const handler = this._handlers[path];
+
+        handler.watchInterval = setInterval(() => {
```
aza547, line 412: This pre-existing logic causes problems. See #155 - I wonder if we can fix while we are here?

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+            const latestLogFile = this.getLatestLog(path);
+
+            // Handle the case where there is no logs in the WoW log directory.
+            if (!latestLogFile) {
+                return;
+            }
+
+            const logFileChanged = (latestLogFile !== handler.currentLogFile);
+
+            if (!handler.currentLogFile || logFileChanged) {
```
aza547, line 422: Don't think you need the (!handler.currentLogFile) part of this check - it's already caught by the logFileChanged variable

src/main/logutils.ts
```diff
@@ -22,66 +17,51 @@ let testRunning: boolean = false;
...
+ * Log line handlers attached to their respective combat log event types.
+ * Any combat log event is valid here.
+ *
+ * Callback signature is:
+ *
+ * (line: LogLine, flavour: string)
+ *
+ * 'flavour' is the WoW flavour like 'wow_classic', 'wow' (retail)
+ */
+combatLogParser
```
aza547, line 39: beautiful

src/main/combatLogParser.ts
aza547: Actually maybe this is just an outdated comment and the 30s should be the timeout that the constructor takes.

aza547: In which case this is likely fine. On second read I think it is.

src/main/logutils.ts
```diff
@@ -22,66 +17,51 @@ let testRunning: boolean = false;
...
-
-        if (day) dateObj.setDate(day);
-        if (month) dateObj.setMonth(month - 1);
-        dateObj.setHours(hours);
-        dateObj.setMinutes(mins);
-        dateObj.setSeconds(secs);
-        dateObj.setMilliseconds(msec);
-
-        return dateObj;
+    if (activeChallengeMode) {
```
aza547, line 60: Do we want other content types in here? I think we certainly want BGs and possibly others. BGs are particularly hard to deal with as we rely on ZONE_CHANGE events only, and the timeout could be big value there.

src/main/logutils.ts
```diff
@@ -411,22 +231,22 @@ function handleChallengeModeStartLine (line: LogLine): void {
...
         console.error(`[ChallengeMode] Invalid/unsupported mapId for Challenge Mode dungeon: ${mapId} ('${zoneName}')`)
     }
 
-    const dungeonAffixes = line.args[5].map((v: string) => parseInt(v, 10));
+    const dungeonAffixes = line.arg(5).map((v: string) => parseInt(v, 10));
 
     activeChallengeMode = new ChallengeModeDungeon(
         dungeonTimersByMapId[mapId], // Dungeon timers
-        parseInt(line.args[2], 10),  // zoneId
+        parseInt(line.arg(2), 10),  // zoneId
```
aza547, line 247: comment alignment

src/main/logutils.ts
```diff
@@ -821,16 +619,33 @@ const forceStopRecording = () => {
...
+    // Regardless of what happens above these lines, _ensure_ that these variables
+    // are cleared.
     activeChallengeMode = undefined;
     testRunning = false;
 
     recorder.stop(metadata, 0);
 }
 
+/**
+ * Watch the logs. Check every second for a new file,
```
aza547, line 640: Comment has a implementation details that probably don't need to be here.

src/main/main.ts
```diff
@@ -84,8 +84,12 @@ const loadRecorderOptions = (cfg: ElectronStore): RecorderOptionsType => {
  *   - (dev)  "C:\Users\alexa\AppData\Roaming\Electron\config.json"
  */
 const cfg = new Store();
-let baseLogPath: string = getPathConfigSafe(cfg, 'log-path');
 let recorderOptions: RecorderOptionsType = loadRecorderOptions(cfg);
+let baseLogPaths: string[] = [
+  getPathConfigSafe(cfg, 'log-path'),
+  getPathConfigSafe(cfg, 'log-path-classic'),
+  getPathConfigSafe(cfg, 'log-path-classic-era'),
```
aza547, line 91: is the 3rd for season of mastery? Hadn't expected that.

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+    };
+
+    /**
+     * Start watching a path, if it's a valid WoW logs directory.
+     * This is checked via `getWowFlavour()` which looks for the file '../.flavour.info'
+     * relative to the log directory, which all flavours of WoW have.
+     */
+    watchPath(pathSpec: string): void {
+        pathSpec = path.resolve(pathSpec);
+        if (pathSpec in this._handlers) {
```
aza547, line 241: Why do this and not simply return if we are already watching it?

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+
+    constructor(options: CombatLogParserOptionsType) {
+        super();
+
+        this._options = options;
+    };
+
+    /**
+     * Start watching a path, if it's a valid WoW logs directory.
+     * This is checked via `getWowFlavour()` which looks for the file '../.flavour.info'
```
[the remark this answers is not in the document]
aza547: Can we make it obvious here that this is totally expected to watch multiple paths at once?

aza547 (changes requested):
This is really great code. I appreciate all the hard work that has gone into this!

Minor comments.

src/main/combatLogParser.ts
aza547: Regarding not being async, this actually gets a lot worse with this PR as we open the possibility of the program running 3 "list files" operation per second instead of 1, so the "mildly annoying" jumping we see now could easily become "moderately annoying"

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+        // Split the line by any delimiter that isn't a number,
+        // convert them to actual numbers and reverse the array.
+        const timeParts = this.timestamp
+            .split(/[^0-9]/, 6)
+            .map(v => parseInt(v, 10))
+            .reverse();
+        const [msec, secs, mins, hours, day, month] = timeParts;
+        const dateObj = new Date();
+
+        if (month) dateObj.setMonth(month - 1);
```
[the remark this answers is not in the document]
tdaugaard: Because some of the test files were missing the date and day parts and this would make it fail - that is also why the array is reversed because the _end_ (time stamp) is always there.

So it's basically just a guard against those files where the date/time looks like this:

`20:13:42.126`

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+    /**
+     * Parse the timestamp from a log line and create a Date value from it
+     */
+    date(): Date {
+        // Split the line by any delimiter that isn't a number,
+        // convert them to actual numbers and reverse the array.
+        const timeParts = this.timestamp
+            .split(/[^0-9]/, 6)
+            .map(v => parseInt(v, 10))
+            .reverse();
```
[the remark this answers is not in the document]
tdaugaard: As commented above, because some of the test files were missing the date and day parts and this would make it fail - that is also why the array is reversed because the end (time stamp) is always there.

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+            throw `Unexpected EOL. There are ${openListCount} open list(s).`
+        }
+    }
+
+    /**
+     * Add an argument to the list
+     */
+    private addArg(value: any): void {
+        this._args.push(value);
+        this._argsListLen = this._args.length;
```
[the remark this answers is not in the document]
tdaugaard: Because it _ensures_ that we'll never try to make the array appear longer than it is, if an error was introduced.

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+
+        this._options = options;
+    };
+
+    /**
+     * Start watching a path, if it's a valid WoW logs directory.
+     * This is checked via `getWowFlavour()` which looks for the file '../.flavour.info'
+     * relative to the log directory, which all flavours of WoW have.
+     */
+    watchPath(pathSpec: string): void {
```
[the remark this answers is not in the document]
tdaugaard: `path` clashes with the `path` module - `pathSpec` is short for `pathSpecification`. that is a given path which hasn't been verified/checked/normalized.

src/main/logutils.ts
tdaugaard: I don't know if we do - I just knew that Keystone Dungeons need this. I hope that those of you who do PvP can chime in with how we could use this to handle other content types here?

src/main/logutils.ts
tdaugaard: Wasn't me - but I'll removed it 😂

src/main/combatLogParser.ts
tdaugaard: That's a damn good question to which I don't have a useful answer. 🤷‍♂️

src/main/combatLogParser.ts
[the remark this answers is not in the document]
aza547: What test files? Can't we just fix them?

src/main/combatLogParser.ts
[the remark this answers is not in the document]
aza547: ah- fine!

src/main/combatLogParser.ts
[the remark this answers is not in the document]
tdaugaard: I do believe it was actually the hardcoded test lines for the recording test in `runRecordingTest()`. Could definitely just fix them. I will fix it.

src/main/logutils.ts
aza547: For now can we just add BGs this? 

```
if (activeChallengeMode || isBG)
...
```

src/main/combatLogParser.ts
tdaugaard: The CombatLogParser class does not concert itself with what happens when the timeout reaches, except it "unlocks" the combat log such that we'll be able to receive data from any other watches logs.

This in turn fires an event (DataTimeout) that logutils catches and can act on, to stop recording for example.

The 30 seconds is a remnant - the timeout is actually set in the constructor and currently at 2 minutes. This could be made an advanced Setting so people can control it. 2 minutes might be a bit too short sometimes.

src/main/combatLogParser.ts
```diff
@@ -0,0 +1,434 @@
...
+        };
+
+        console.log(`[CombatLogParser] Start watching '${pathSpec}' for '${wowFlavour}'`, this._handlers[pathSpec]);
+        this.watchLogDirectory(pathSpec);
+    }
+
+    /**
+     * Unwatch a previously watched path.
+     */
+    unwatchPath(pathSpec: string): void {
```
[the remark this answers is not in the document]
tdaugaard: `unwatchPath()` is public and _no longer_ called from `watchPath`, so it'll stay here now that it's gone from the other one.

src/main/combatLogParser.ts
tdaugaard: We could definitely merge those two. `getSortedVideos(dir: string)` could become `getSortedFiles(dir: string, pattern: string)`. Good idea!

Re: async, yes - as much as possible things should be async. Definitely agree.

tdaugaard: The only reason to keep it in CombatLogParser, seperate, is that it's a self-contained unit.

I _could_ make an option in the ctor that would take `globFn` or something to give it a method for finding files. 🤔 Will do!

tdaugaard: This has been fixed!

src/main/combatLogParser.ts
tdaugaard: This has been fixed in 6633c90 due to it becoming async because of the change to `getSortedVideos()` becoming `getSortedFiles()`.

src/main/combatLogParser.ts
tdaugaard: Yeah, that makes sense, nice find!

PR #157, opened 2022-09-24
Draft: Update settings page

aza547:


src/renderer/rendererutils.ts
```diff
@@ -20,5 +20,5 @@ const getVideoResult = (video: any): boolean => {
 
 export {
     getFormattedDuration,
-    getVideoResult,
+    getVideoResult
```
tdaugaard, line 23: Pretty please leave trailing commas, so the next addition doesn't show the `getVideoResult` line as having changed when in fact it wasn't :p

src/settings/AudioSettings.tsx
```diff
@@ -0,0 +1,115 @@
...
+      new ObsAudioDevice('none', '(None: no sound will be recorded)'),
+      new ObsAudioDevice('all', '(All)'),
+      ...audioDevices.output,
+    ]
+  };
+
+  /**
+   * setSetting, why not just use react state hook?
+   */
+  const setSetting = (stateKey: StateToSettingKeyMapKey, value: any) => {
```
tdaugaard, line 39: Perhaps this should be split out and be shared amongs the tabs?

src/settings/ContentSettings.tsx
```diff
@@ -0,0 +1,29 @@
...
+import FormLabel from '@mui/material/FormLabel';
+
+export default function ContentSettings() {
+
+  return (
+    <div>
+      <FormLabel id="radios" sx={{color: "white"}}>PvE</FormLabel>
+      <Divider color="black" />
+      <FormGroup sx={{ width: '48ch', padding: 1 }} >
+        <FormControlLabel control={<Checkbox defaultChecked style = {{color: "#bb4220", padding: 5}} />} label="Raids" style = {{color: "white"}} />
```
tdaugaard, line 15: Sounds like we need `categories` to be labeled according to the type of activity, PvE or PvP and allow them to be dynamically shown here.

src/renderer/rendererutils.ts
aza547: That's great advice actually. I've never considered this!

PR #161, opened 2022-09-25
Update settings page

aza547:
TODO:
- Make content type config actually do something.
- Make it clear to users that classic isn't supported yet.
- All sorts of testing.


src/main/configService.ts
```diff
@@ -6,15 +6,24 @@ import { EventEmitter } from "stream";
...
     startUp?: boolean,
+    recordRetail: boolean,
+    recordClassic: boolean,
+    recordRaids: boolean,
+    recordDungeons: boolean,
+    recordTwoVTwo: boolean,
+    recordThreeVThree: boolean,
+    recordSkirmish: boolean,
+    recordSoloShuffle: boolean,
+    recordBattlegrounds: boolean,
```
tdaugaard, line 26: I'm thinking we make this an array like `recordCategories` which just has the labels like "2v2", "Mythic+" but it's not super important.

aza547:
Also can we expose the schema help text in the UI, save people getting config wrong

aza547:
Also some validation on log path. Confirm it contains world of warcraft or something atleast. People keep getting that wrong.

aza547:
Ah you're right about the flavor stuff of course. I care less about this if we are going to give the exclamation icon if they get it wrong, that's way better than we do currently.

aza547:
I've done some of the todos:

Make content type config actually do something.
- This was a bit hacky but I'm not totally sure how to do what I want with the types. 
- Hoping you can educate me @tdaugaard. 

Make it clear to users that classic isn't supported yet.
- I've hidden config for classic/retail now by commenting it out. 
- Same for classic log path.

PR #171, opened 2022-09-27
Fix unable to test record with 2v2 disabled

tdaugaard:
When the content type '2v2' is disabled, the test recording won't run because it runs a test on '2v2'.

This has been fixed by allowing a recording to take place even if we're not allowed to record the test content type, as long as `testRunning` is true.

Refactored `categoryRecordConfigMapping` into
`categoryRecordingSettings`, since we already had a similar object for 'overrun' (`videoOverrunPerCategory`). These have been merged and renamed.

In logutils, the logic regading the allowance of recording content types has been centralised into `allowRecordCategory()`, which also takes `testRunning` into account.

This method is used in `startRecording()` so there's a central place where this is checked and blocked, if needed.

Cleaned up some of the `console.log()` messages and removed the TODO regarding afk/hearth out as that has been fixed with the DataTimeout mechanism.

### DIFF
--- a/src/main/configService.ts
+++ b/src/main/configService.ts
@@ -4,7 +4,7 @@ import path from "path";
 import { EventEmitter } from "stream";
 import { CombatLogParser } from "./combatLogParser";
 
-type ConfigurationSchema = {
+export type ConfigurationSchema = {
     storagePath: string,
     bufferStoragePath?: string,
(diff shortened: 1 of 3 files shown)


aza547:
Looks good. Grabbing this and doing some testing before a minor version release.

PR #172, opened 2022-09-27
Improve logic around waiting for OBS to close temporary video file

tdaugaard:
OBS won't close the temporary video file until _after_ the signal `wrote` has been seen from OBS, but it isn't observed and isn't reliable either and instead of using `setTimeout()` we can test for the file being open by other applications and simply wait for that, which will take a shorter time and also be more reliable since we're not "hoping" it's there after 2 seconds, which could present some problems on slower disks.

## IMPORTANT

The new node module `@ronomon/opened` that has been added requires you to run the following to recompile the native C-bindings that it uses:

```
.\node_modules\.bin\electron-rebuild
```

This change introduces the following major changes:

### In `recorder.ts`:

- `finalizeVideo()` no longer uses `setTimeout()` to hope that the video file is released by OBS. It doesn't wait at all, anymore, but simply awaits `util.cutVideo()` to complete.

  It also doesn't call `updateStatus` anymore, since we're not _actually_ ready for recording again until _after_ `stop()` finishes entirely, after running `cleanupBuffer()` and `startBuffer()`.

  This logic has been moved to `stop()` where it belongs.

### In `errors.ts`:

- New file that contains classes inheriting Error(), such that we can add more useful information to our errors and differentiate them in `catch()` at some point.

### In `util.ts`:

- `waitFile()` has been created which uses a new module `@ronomon/opened` which can check whether a file is open for exclusive write access/locked in another aplication.

- `ffmpegProbe()` has been created to serve as a Promise version of `ffmpeg.ffprobe()` to make `cutVideo()` look much cleaner.

- `cutVideo()` has been _majorly_ overhauled and now uses `waitFile()` to ensure we don't attempt to operate on a file still locked by OBS. If we don't have a result after 10 seconds, we'll throw an error and assume something went wrong.

  A new guard against in invalid `duration` has been put in place to avoid duration being a null value, which causes ffmpeg to error.

  `throw new Error()` has been replaced by `FfmpegError()` and `FfmpegProbeError()` such that we can pass additional information to it and have more information at the point where we - eventually - will catch and handle them.


aza547:
Can you clarify on the requirement to build native bindings? When do we need to run that? Every build? Once per checkout?

tdaugaard:
> Can you clarify on the requirement to build native bindings? When do we need to run that? Every build? Once per checkout?

So far I've only had to do it once per repo clone, but I haven't tested it thorougly across branch switching, npm install's and what have we.

tdaugaard:
Well, that is it seems that if I switch to another branch that _doesn't_ have this package and run `npm install` it loses the package and the compiled bindings, but when the `@ronomon/opened` package isn't removed, the building of the native bindings persists.

aza547:
OK, I'd like to avoid adding a step to our build instructions, seems like we should be able to include that step in the correct place in package.json to have npm do it for us?

tdaugaard:
> OK, I'd like to avoid adding a step to our build instructions, seems like we should be able to include that step in the correct place in package.json to have npm do it for us?

It is necessary to run it _once_ everytime you clone the repo fresh and run `npm install` for the first time and therefore I'd say it's also required when you build the app, if you build it via an automated pipeline.

If you're not interested in this, you can close the PR as the whole simplification and change revolves around the feature that this module provides :)

PR #173, opened 2022-09-27
Fix crash when force stopping a recording

tdaugaard:
`const videoStopDate = ..` in `forceStopRecording()` was shadowing the global variable with the same name that is the one that was supposed to have the value.

Removing `const` fixes this as it no longer declares the variable local to the function.


aza547:
That is annoying that VSCode / TS doesn't spot this. Good catch.

PR #174, opened 2022-09-27
Fix test recording not working for some locales

tdaugaard:
12-hour clock locales like `en-US` would cause the test recording to fail because the resulting date from the test combat log lines would be malformed (`9/27 2:50:54 PM` instead of `9/27 14:50:54`) and cause `videoStartDate` to be null.

Furthermore, `LogLine.date()` assumed the date would only have 5 items which would cause the malformed date to result in an array of:

`[9, 27, 2, 50, '54 PM']` instead of `[9, 27, 2, 50, 54, 'PM]` which would've worked fine since we only care about the first 5 items.

The fix:

- No limit on splitting the date in `LogLine.date()`

- Enforce a locale of `en-GB` on `toLocaleTimeString()` such that we're sure we always get a correct time.

- Output the log lines that we're passing to the combat log parser during a test for easier debugging in the future.

### DIFF
--- a/src/main/combatLogParser.ts
+++ b/src/main/combatLogParser.ts
@@ -89,7 +89,7 @@ class LogLine {
         // Split the line by any delimiter that isn't a number
         // and convert them to actual numbers.
         const timeParts = this.timestamp
-            .split(/[^0-9]/, 6)
+            .split(/[^0-9]/)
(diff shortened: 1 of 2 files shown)


aza547:
Love it!

PR #181, opened 2022-09-28
Feature: Configurable output resolution

tdaugaard:
Add `obsBaseResolution` and `obsOutputResolution` as configurable settings.

Only `obsOutputResolution` can be configured at the moment, and base resolution will always be set to the resolution of the selected monitor.

Both have been made configurable'ish because it is intended to be exposed at a later time with a different mechanic that isn't suitable for this change, and it was easier to add both now. In the future when Game Capture is implemented, we might want to make Base Resolution configurable, so that's why.

Please note, that the changes to output resolution does not always take effect before the application has been restarted. This is a known issue with the reconfiguration of OBS that applies to most of its settings that are saved in `basic.ini`, but _not_ those that are set in-memory like output sources and scenes.


aza547:
This looks great - not reviewed code yet but had a quick test and mostly worked- one issue that app crashes when doing first time config (I deleted my config-v2.json and tried to reconfigure and it closes when I press the settings). I think we've seen this issue recently, I'll try refresh my memory.

aza547:
> This looks great - not reviewed code yet but had a quick test and mostly worked- one issue that app crashes when doing first time config (I deleted my config-v2.json and tried to reconfigure and it closes when I press the settings). I think we've seen this issue recently, I'll try refresh my memory.

Ah yes, it's the same problem as we had with getting the audio devices before OBS has been configured. I think you want a guard around this before we have a recorder object. For example:

```
ipcMain.on('getAudioDevices', (event) => {
  // We can only get this information if the recorder (OBS) has been
  // initialized and that only happens when the storage directory has
  // been configured.
  if (!recorder) {
    event.returnValue = { input: [], output: [] };
    return;
  }

  event.returnValue = {
    input: getAvailableAudioInputDevices(),
    output: getAvailableAudioOutputDevices(),
  };
});
```

src/main/obsRecorder.ts
```diff
@@ -4,6 +4,7 @@ import { getAvailableAudioInputDevices, getAvailableAudioOutputDevices } from ".
 import { RecorderOptionsType } from "./recorder";
 import { OurDisplayType } from "./types";
 import { Size } from "electron";
+import { IScene } from "obs-studio-node";
```
aza547, line 7: I've seen this naming a lot but never understood it. What is the I prefix?

tdaugaard:
Could be, yeah. I have been working on another branch that would allow us to initialize OBS without starting the buffer recording _before_ we have any meaningful configuration that would allow us to gather these values and avoid the guards here, but that is obviously not part of this PR so I'll find a fix for this seperately.

aza547 (changes requested):
No criticisms of the code, just the one bug to fix and one question :)

src/main/obsRecorder.ts
tdaugaard: It's a prefix commonly used for interfaces, mostly to be able to visually distinguish them from classes and types.

tdaugaard:
Guard has been added against OBS not being initialized.

src/main/logutils.ts
```diff
@@ -669,6 +669,7 @@ const forceStopRecording = () => {
  * @returns {[boolean, boolean]} retailRunning, classicRunning
  */
 const checkWoWProcess = async (): Promise<[boolean, boolean]> => {
+    return [true, false];
```
aza547, line 672: it happened!

src/main/main.ts
```diff
@@ -410,6 +412,16 @@ ipcMain.on('settingsWindow', (event, args) => {
     event.returnValue = getAvailableDisplays();
     return;
   }
+
+  if (args[0] === 'getObsAvailableResolutions') {
+    if (!recorder) {
```
aza547, line 417: pedantic comment, but perhaps neater as:

```
if (args[0] === 'getObsAvailableResolutions') {
  event.returnValue = { 'Base': [], 'Output': [] };
  if (!recorder) return;
  event.returnValue = getObsResolutions();
}
```

src/settings/VideoSettings.tsx
```diff
@@ -22,6 +26,10 @@ export default function VideoSettings() {
     setConfig((prevConfig: any) => ({ ...prevConfig, [stateKey]: value }));
   };
 
+  if (!fpsOptions.includes(config.obsFPS)) {
```
aza547, line 29: Why is this required? Shouldn't we be getting a default from electron store if it's not set?

src/settings/VideoSettings.tsx
```diff
@@ -46,30 +57,70 @@ export default function VideoSettings() {
...
-            )}
-          </Select>
-        </FormControl>
-        <Tooltip title={configSchema["monitorIndex"].description} >
-          <IconButton>
-            <InfoIcon style={{ color: 'white' }}/>
-          </IconButton>
-        </Tooltip>
-      </Box>
+      <FormControl fullWidth>
```
aza547, line 60: Seem to have lost the MUI box element here. I think that was crucial for aligning things when I made the settings changes recently. Is that deliberate? Does it still look OK? Also lost some of the margin specification. Suspect a conflict resolution gone the wrong direction?

aza547 (commented):
Few more comments.

aza547:
Oh, and need a changelog entry for this please

src/main/logutils.ts
tdaugaard: HA HA HA. _Maaaan._ Fixing. 😂

src/main/main.ts
tdaugaard: I'm not a huge fan of non-braces if's, but I see you've done it in _many_ places so as you wish =)

src/settings/VideoSettings.tsx
tdaugaard: If someone edits their config by hand and types 120 or 144 or something, OBS will just use 60, but in order to show something meaningful in the dropdown. `FPSCommon` only supports a few common FPS values.

src/settings/VideoSettings.tsx
tdaugaard: That's not inconceivable. I'll have a look at it.

src/main/main.ts
tdaugaard: This was apparently not possible, as setting `event.returnValue` twice resulted in the second assignment being ignored and not returned. I suspect that behind the scens, any assignment to `returnValue` will instantly return the `ipc.sendSync()` call and thus render subsequent assignments pointless.

aza547: To be honest I'm not sure what I prefer. I spent the first year or two of my career just writing Java as a {} using langauge and the coding standards at work mandate

```
if ()
{
 ...
}
else 
{
  ...
}
```

formatting. Not going to get excited about whatever goes in here.

aza547: Interesting re the event.returnValue stuff though - I think I actually spotted that once too but obviously forgot. Fine as is!

aza547:
Great! Merging.

src/main/main.ts
tdaugaard: My person preference is that whatever makes the intent of the code easier to read is whatever I go with, but I can see how the following can look off, as the `else` could be easily missed.

```js
if (...) {
} else
if (...) {
}
```

PR #182, opened 2022-09-29
Feature: ConfigService singleton & config keys cleanup

tdaugaard:
### Refactor ConfigService() into a singleton

We only ever need a single instance of this class and it was causing excessive log output due to duplicate instantiation of the class.

It has been turned into a singleton which is now created/obtained by simply calling `ConfigService.getInstance()`.

### Move config schema and type to configSchema.ts

In order to avoid duplication of the configuration keys in `useSettings.ts`, the configuration schema and its associated type has been moved to a new file that can be imported in the frontend free from filesystem imports.

`configValues` in `useSettings.ts` has been moved outside of the `useState()` method as we only need to load them from the configuration store _once_ per config window creation as they cannot and should not change outside of the config window.

### Delete deprecated config keys from ElectronStore

Over time, when config keys change and become deprecated they will stay and linger in the configuration store and this change will remove them when the app starts.


tdaugaard:
Converted to draft. Found a possible bug where, after saving settings it'll forget about `selectedCategory` which makes it look weird when it starts because no tab is selected.

tdaugaard:
Bug has been found and fixed, PR ready for review.

src/main/configSchema.ts
```diff
@@ -0,0 +1,140 @@
...
+    bufferStoragePath?: string,
+    retailLogPath?: string,
+    classicLogPath?: string,
+    maxStorage: number,
+    monitorIndex: number,
+    selectedCategory: number,
+    audioInputDevice?: string,
+    audioOutputDevice?: string,
+    minEncounterDuration: number,
+    startUp?: boolean,
```
aza547, line 12: I think this doesn't need to be optional

src/main/configSchema.ts
```diff
@@ -0,0 +1,140 @@
+export type ConfigurationSchema = {
+    storagePath: string,
+    bufferStoragePath?: string,
+    retailLogPath?: string,
+    classicLogPath?: string,
+    maxStorage: number,
+    monitorIndex: number,
+    selectedCategory: number,
+    audioInputDevice?: string,
```
aza547, line 9: I think this doesn't need to be optional

src/main/configService.ts
```diff
@@ -167,13 +57,19 @@ export default class ConfigService extends EventEmitter {
...
                 const value = this.get(key);
-                console.log('[Config Service] Get from config store:', key, value);
                 event.returnValue = value;
             } else
             if (fn === 'set') {
                 const value = args[2];
                 this.set(key, value);
-                console.log('[Config Service] Set in config store:', key, value);
+
+                this._changedProperties[key] = value;
```
aza547, line 66: Why do we need this variable? I don't see any uses of it. It feels a bit hacky to have this tracked as such, I don't have any objection to the actual code here but if we're going to have something checking this we're opening up opportunity for timing problems. 

If we really need this, then I'd prefer a pubsub style mechanism that we store the "changed" property on some object, like this, and then dismiss it once we've processed it.

tdaugaard: I'll find a different way to achieve what I intended. The intended result was that we only log the things that have changed/been set - this could be achieved in the settings window instead, collecting the values that have changed and saving those in bulk instead. I'll work on that.

I agree this solution wasn't the prettiest.

src/main/configSchema.ts
tdaugaard: They were copy/pasted from `configService.ts`, so I didn't change them, but I can surely make them non-optional anyway :)

aza547:
Pushed some changes that have caused conflicts here but have tried to make sure I stuck to your plan so hoping they are easy to resolve. I needed to split the configSchema out into its own file for the settings help text stuff.

aza547:
I think there are a few changes still needed here before merge? Just to check I'm not blocking this moving.

tdaugaard:
Yeah, I'll let you know when it's ready, thanks for checking =)

src/main/configService.ts
tdaugaard: Better solution was found - it now has a `set_values` function that checks the object being passed against the current configuration before setting anything, then logs the changes afterwards.

tdaugaard:
@aza547 ready for review.

src/main/configSchema.ts
```diff
@@ -1,15 +1,15 @@
 export type ConfigurationSchema = {
     storagePath: string,
-    bufferStoragePath?: string,
-    retailLogPath?: string,
-    classicLogPath?: string,
+    bufferStoragePath: string,
+    retailLogPath: string,
+    classicLogPath: string,
```
aza547, line 5: Just to confirm my understanding here. Even if user doesn't set classicLogPath, it will end up in the config as an empty string, so this is fine?

tdaugaard: Correct. Empty strings should be considered an abxesnt value anyway for a path value.

src/main/configService.ts
```diff
@@ -33,17 +56,49 @@ export default class ConfigService extends EventEmitter {
...
+                        return;
+                    }
+
+                    this.set(key, value);
+                    this.emit('change', key, value);
+                    this.logConfigChanged({[key]: value});
+                    return;
+                }
+
+                case 'set_values': {
```
aza547, line 78: Why do we need this case now?

src/main/configService.ts
```diff
@@ -33,17 +56,49 @@ export default class ConfigService extends EventEmitter {
...
+            switch (args[0]) {
+                case 'get': {
+                    const value = this.get(args[1]);
+                    event.returnValue = value;
+                    return;
+                }
+
+                case 'set': {
+                    const [_, key, value] = args;
+                    if (this.configValueChanged(key, value)) {
```
aza547, line 68: Am I dumb or is this condition missing an !?

src/main/configService.ts
aza547: I think it's for multi-valued attrs, I don't think we have any of those? Maybe you've done this for future things?

src/settings/useSettings.ts
```diff
@@ -5,64 +5,42 @@ export function getConfigValue<T>(configKey: string): T {
...
-  'recordSkirmish',
-  'recordSoloShuffle',
-  'recordBattlegrounds',
-  'obsBaseResolution',
-  'obsOutputResolution',
-  'obsFPS',
-  'obsKBitRate',
-];
+export function setConfigValues(dict: { [key: string]: any }): void {
+  window.electron.ipcRenderer.sendMessage('config', ['set_values', dict]);
```
aza547, line 12: and that will be why we need set_values!

src/settings/Settings.tsx
```diff
@@ -69,14 +69,7 @@ export default function Settings() {
...
 
-    configSettings.forEach((key: string) => {
-      const configKey = (key as keyof typeof config);
-      setConfigValue(key, config[configKey]);
-
-      if (key === "startUp") {
-        ipc.sendMessage("settingsWindow", ["startup", config[configKey]]);
-      }
-    });
+    setConfigValues(config);
```
aza547, line 72: love it

aza547 (commented):
A few comments to address. Keen to get this in though

src/main/configService.ts
```diff
@@ -33,17 +56,49 @@ export default class ConfigService extends EventEmitter {
...
-                event.returnValue = value;
-            } else
-            if (fn === 'set') {
-                const value = args[2];
-                this.set(key, value);
-                console.log('[Config Service] Set in config store:', key, value);
+            switch (args[0]) {
+                case 'get': {
+                    const value = this.get(args[1]);
+                    event.returnValue = value;
```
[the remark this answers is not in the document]
tdaugaard: I totally agree! Please see line 37 - it logs this at the beginning:

```js
console.log('[Config Service] Using configuration', inspectObject(this._store.store));
```

src/main/configService.ts
tdaugaard: You're right. I don't know how I missed that. :|

src/main/configService.ts
tdaugaard: No, this is for setting an object of values directly; it's used in `src/settings/Settings.tsx` line 72, which calls `setConfigValues()` from `useSettings.ts`, which calls the IPC channel `config` with `set_values` to avoid setting them one by one and being unable to print the differences _after_ setting everything.

PR #188, opened 2022-10-02
Make detection of WoW running better

tdaugaard:
- Detect any known version of WoW:
    - Wow.exe
    - WowB.exe
    - WowT.exe
    - WowClassic.exe

- Any version of WoW will keep the app thinking WoW is running, that is if you start Wow.exe, then start WowClassic.exe, then close Wow.exe it will keep monitoring log files because Classic is still running.

### DIFF
--- a/src/main/logutils.ts
+++ b/src/main/logutils.ts
@@ -1,8 +1,8 @@
 /* eslint import/prefer-default-export: off, import/no-mutable-exports: off */
 import { Combatant } from './combatant';
 import { recorder }  from './main';
-import { battlegrounds, categoryRecordingSettings, dungeonEncounters, dungeonsByMapId, dungeonTimersByMapId, VideoCategory }  from './constants';
-import { PlayerDeathType, UnitFlags } from './types';
+import { battlegrounds, categoryRecordingSettings, dungeonEncounters, dungeonsByMapId, dungeonTimersByMapId, VideoCategory, wowExecutableFlavours, WoWProcessResultKey }  from './constants';
+import { IWoWProcessResult, PlayerDeathType, UnitFlags } from './types';
 import { ChallengeModeDungeon, ChallengeModeTimelineSegment, TimelineSegmentType } from './keystone';
 import { CombatLogParser, LogLine } from './combatLogParser';
 import { getSortedFiles } from './util';
@@ -89,8 +89,7 @@ combatLogParser.on('DataTimeout', (timeoutMs: number) => {
 /**
  * Is wow running? Starts false but we'll check immediately on start-up.
  */
-let isRetailRunning: boolean = false;
-let isClassicRunning: boolean = false;
+let wowProcessRunning: IWoWProcessResult | null = null;
 
 /**
  * Timers for poll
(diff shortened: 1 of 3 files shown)


src/main/logutils.ts
```diff
@@ -665,46 +669,49 @@ const forceStopRecording = () => {
...
+    if (!wowProcess && wowProcessRunning !== null) {
+        wowProcessStopped();
+    }
 }
 
 /**
  * pollWoWProcess
  */
 const pollWowProcess = () => {
-    pollWoWProcessLogic(true);
```
tdaugaard, line 705: What was the point of this logic of `pollWoWProcessLogic()` not starting or stopping if its argument was `false`?
I've removed it because it seemed pointless and the app responds as expected without it, but if there was a specific reason for it that I don't know about, I'll revert that change.

aza547: I'm not really sure actually. I think it can be removed. Vaguely remember being irritated by the log we made as "Wow has started" isn't really true if we've just started WR with wow already running. But as far as I can tell we still suffer for that and just the log is a little less specific with "wow is running".

src/main/logutils.ts
```diff
@@ -665,46 +669,49 @@ const forceStopRecording = () => {
...
-    const processChanged = (retailProcessChanged || classicProcessChanged);
-    if (!retailProcessChanged && !startup) return;
-    (retailFound) ? wowProcessStarted() : wowProcessStopped();
+const pollWoWProcessLogic = async () => {
+    const wowProcesses = await checkWoWProcess();
+    const wowProcess = wowProcesses.pop();
+
+    if (wowProcess && wowProcessRunning === null) {
+        wowProcessStarted(wowProcess);
+    } else
```
aza547, line 700: Looks a bit dodgy, extra new line?

src/main/logutils.ts
```diff
@@ -665,46 +669,49 @@ const forceStopRecording = () => {
...
-    const [retailFound, classicFound] = await checkWoWProcess();
-    const retailProcessChanged = (retailFound !== isRetailRunning);
-    // TODO classic support
-    const classicProcessChanged = (classicFound !== isClassicRunning);
-    const processChanged = (retailProcessChanged || classicProcessChanged);
-    if (!retailProcessChanged && !startup) return;
-    (retailFound) ? wowProcessStarted() : wowProcessStopped();
+const pollWoWProcessLogic = async () => {
+    const wowProcesses = await checkWoWProcess();
+    const wowProcess = wowProcesses.pop();
```
aza547, line 696: So we'll only spot one, but we don't do any custom logic now so this is fine? If all 4 are running we'll spot one. No objection just confirming.

aza547 (commented):
Minor comments, looks good!

src/main/logutils.ts
tdaugaard: Yeah I was considering changing the log message to "_Detected WoW (<flavour>)_"

src/main/logutils.ts
tdaugaard: I've done that other places too, it's just to avoid second `if` being indented unnecessarily.

src/main/logutils.ts
tdaugaard: Precisely!

src/main/logutils.ts
aza547: Sounds good to me!

src/main/logutils.ts
aza547: that being said it's so minor it's not worth worrying about

aza547:
I'll just merge and update the log text, I think that's all that is remaining here!

src/main/logutils.ts
tdaugaard: Oh. I was totally gonna change it, but I guess that's too late now 🙈

PR #190, opened 2022-10-04
Feature: Configurable OBS recording encoder

tdaugaard:
Adds configurable OBS Recording Encoder (`RecEncoder`) and bitrate (necessary for AMD AMF/x264).

Let's discuss it. Bitrate is required for AMD AMF and obs_x264 encoders but obviously changing between them, the bitrate might not mean the same. as in ... 50 kbit/s on AMD AMF might be better or worse than the same for obs_x264.

I made this change because a few people have suggested it.

Adresses issue #67. There are several ways of doing this, one is bitrate and the other is the simple version with "quality presets" which are slightly more difficult to implement, I believe.

### DIFF
--- a/src/main/configSchema.ts
+++ b/src/main/configSchema.ts
@@ -14,6 +14,8 @@ export type ConfigurationSchema = {
     obsBaseResolution: string,
     obsOutputResolution: string,
     obsFPS: number,
+    obsRecEncoder: string,
+    obsRecBitrate: number,
     recordRetail: boolean,
     recordClassic: boolean,
     recordRaids: boolean,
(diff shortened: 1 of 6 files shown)


tdaugaard:
@aza547 how do you feel about this?

aza547:
Agree this is definitely something we want. It'd be awesome to let people put their recording on their CPU should they want also by selecting a x264 encoder or so. 

The impression I got from reading some OBS forums was that constant bitrate is bad for loads of reasons. 
- For AMF encoders currently we use 50k constant bitrate.
- For everything else we use lossless constant bitrate (I think). 

> Bitrate is required for AMD AMF and obs_x264 encoders but obviously changing between them, the bitrate might not mean the same. as in ... 50 kbit/s on AMD AMF might be better or worse than the same for obs_x264.

Are we sure this is true? Have you got a reference?

> There are several ways of doing this, one is bitrate and the other is the simple version with "quality presets" which are slightly more difficult to implement, I believe.

I'm suprised this is more difficult. I was expecting this was the regular way and the way we should do it. When I was playing around with ffmpeg a year or so ago when prototyping some of the app function there were some command line args that you could just specific the quality between two numbers. That had great results at the time for file sizes too, massive reduction with no noticeable difference in quality.

aza547:
I've stolen some of your thunder here with https://github.com/aza547/wow-recorder/commit/d39db36baad1812c641b9d6b01239a6748e30989 - sorry!.

I'd still like to expose the encoder as a setting though, but I think a bit of this is now irrelevant (we don't need seperate handling for AMF/nvenc anymore - yay).

tdaugaard:
> I've stolen some of your thunder here with https://github.com/aza547/wow-recorder/commit/d39db36baad1812c641b9d6b01239a6748e30989 - sorry!.

Steal away! It's all open source :p Let's close this PR as it's no longer relevant in its current and intended state.

PR #196, opened 2022-10-09
Feature: Queue for finalizing videos

### LINKED ISSUE #194: If FFMPEG cuts video slowly things can fall over
**Describe the bug**
A user in discord reported this. In their logs, ffmpeg took 3 mins to cut the video. They pulled the boss before this time elapsed and the app fell over. 

**Expected behavior**
The ffmpeg cutting the video should be async in respect to the recorder process. Currently, the recording buffer does not restart until the cutting is done. That's bad logic. See the code in recorder.stop, particularly `await this.finalizeVideo(metadata, outputFilename);`. 

**Quoting from Discord**
> huh this is an interesting one and is definitely a bug in our code
the problem seems to be that it took your PC ~3 mins to cut one of your pantheon pulls to size
in that time, the app isn't ready to start another recording, but you pulled again in that time
so it fell over
as for why it took so long to cut the video to size, that's very strange and I'm not sure. Was your PC super busy? Anything notable going on?
Regardless of how long it takes, we shouldn't require it completes before the next pull so there is some code fixes I can make to help this.
The cutting of the video usually takes 2-3 seconds in normal circumstances so probably just never hit this before as no one manages to pull within that window.

```
ahk@DESKTOP-C6ME3UB:/mnt/c/Users/alexa/Downloads$ grep "ENCOUNTER_" "WoWCombatLog-100322_173442.txt"  | grep -i panth
10/3 18:17:17.722  ENCOUNTER_START,2544,"Prototype Pantheon",16,20,2481
10/3 18:21:51.041  ENCOUNTER_END,2544,"Prototype Pantheon",16,20,0,273279
10/3 18:23:54.541  ENCOUNTER_START,2544,"Prototype Pantheon",16,20,2481
10/3 18:27:53.051  ENCOUNTER_END,2544,"Prototype Pantheon",16,20,1,238500
```

```
[2022-10-03 18:22:06.060] [info]  [OBS] obsRecorder: stop
[2022-10-03 18:22:06.091] [debug] [OBS] Asserted OBS signal: stopping
[2022-10-03 18:22:06.638] [debug] [OBS] Asserted OBS signal: stop
[2022-10-03 18:22:06.684] [debug] [OBS] Asserted OBS signal: wrote
[2022-10-03 18:22:08.442] [info]  [Util] Ready to cut video.
[2022-10-03 18:22:08.443] [info]  [Util] Initial duration: 572.45 Desired duration: 288 Calculated start time: 284
[2022-10-03 18:22:13.390] [info]  [Config Service] Set in config store: selectedCategory 5
[2022-10-03 18:22:13.396] [info]  [Config Service] Set in config store: selectedCategory 5
[2022-10-03 18:23:54.543] [info]  [Logutils] Start recording a video for category: Raids
[2022-10-03 18:23:54.544] [info]  [Recorder] Start recording by cancelling buffer restart
[2022-10-03 18:25:04.535] [info]  [Util] FFmpeg cut video succeeded
[2022-10-03 18:25:04.548] [info]  [Recorder] Finalized video E:\wowrecordings\2022-10-03 18-12-33 - Raids Sepulcher of the First Ones, Pantheon (Wipe).mp4
[2022-10-03 18:25:04.549] [debug] [Size Monitor] Running (max size = 805.3 GB)
[2022-10-03 18:25:04.562] [info]  [Size Monitor] Unprotected file(s) considered 20, total size = 108.5 GB
[2022-10-03 18:25:04.563] [info]  [Util] Deleting: E:\wowrecordings\.temp\2022-10-03 18-04-20.mp4
[2022-10-03 18:25:09.554] [info]  [Recorder] Start recording buffer
[2022-10-03 18:25:09.555] [info]  [OBS] obsRecorder: start
[2022-10-03 18:25:09.839] [debug] [OBS] Asserted OBS signal: start
```

Things all go downhill from here, as we previously cancelled the recording buffer (which was actually a no-op), so we get a buffer restart inside a genuine recording. Full log is attached. 

[WarcraftRecorder-2022-10-03 (3).log](https://github.com/aza547/wow-recorder/files/9739311/WarcraftRecorder-2022-10-03.3.log)

tdaugaard:
Improve finalizing videos by making them asynchronous.

### Replace `getNewestVideo()` with `getObsLastRecording()`

This will more reliably get the full path and filename of the recording that we'll be processing afterwards.

### Refactor `finalizeVideo()`

Introduce an atomic queue for finalizing videos asynchronously so as not to block the recording process from starting new recordings while processing the previous recording.

This will ensure that we have a minimal delay between ending and starting a recording, even on slower storage devices.

- Rewrite the logic of discarding video content to simplify the conditions and to make better logging for various scensarios.

- Rename `finalizeVideo()` to `queueVideo()` to better reflect the new functionality of the method.

- Restart recording buffer as soon as possible instead of waiting 5 seconds, as the buffer video file will have been released when `queueVideo()` returns.

Partially fixes #194


package.json
```diff
@@ -247,6 +247,7 @@
     "@mui/icons-material": "^5.8.4",
     "@mui/lab": "^5.0.0-alpha.80",
     "@mui/material": "^5.6.4",
+    "atomic-queue": "^5.0.4",
```
aza547, line 250: I used waitQueue elsewhere, can we make these the same? I think they have the same purpose. See obsRecorder.ts.

src/main/recorder.ts
```diff
@@ -51,6 +59,51 @@ type RecorderOptionsType = {
...
+
+            // Run the size monitor to ensure we stay within size limit.
+            // Need some maths to convert GB to bytes
+            runSizeMonitor(this._options.storageDir, this._options.maxStorage)
+                .then(() => {
+                    if (mainWindow) mainWindow.webContents.send('refreshState');
+                });
+
+            // Clean-up the temporary recording directory.
+            this.cleanupBuffer(1);
```
aza547, line 95: Do we need this now you are calling tryUnlinkSync?

src/main/recorder.ts
```diff
@@ -192,58 +244,52 @@ type RecorderOptionsType = {
...
-            this.cleanupBuffer(1);
-
             // Refresh the GUI
             if (mainWindow) mainWindow.webContents.send('refreshState');
 
             // Restart the buffer recording ready for next game.
-            setTimeout(async () => {
-                this.startBuffer();
-            }, 5000)
+            this.startBuffer();
```
aza547, line 264: How will the status icon behave here? Perhaps we want a second status icon for saving, so we can also show we've gone back to a watch state?

src/main/recorder.ts
```diff
@@ -192,58 +244,52 @@ type RecorderOptionsType = {
...
+                    bufferFile,
+                    metadata,
+                };
+
+                console.log("[Recorder] Queuing video for processing", queueItem);
+
+                this._videoQueue.write(queueItem);
+                resolve();
+
                 if (mainWindow) mainWindow.webContents.send('updateStatus', AppStatus.ReadyToRecord);
```
aza547, line 289: I've no idea why this was already like this, but surely this should come before the resolve?

src/main/recorder.ts
```diff
@@ -51,6 +59,51 @@ type RecorderOptionsType = {
...
+
+    /**
+     * Setup events on the videoQueue
+     */
+    private _setupVideoProcessingQueue(): void {
+        this._videoQueue.on('error', (err: any) => {
+            console.error("[Recorder] Error occured during video processing", err);
+        });
+
+        this._videoQueue.on('idle', () => {
```
aza547, line 84: this is really nice

src/main/recorder.ts
tdaugaard: Technically I don't think so, no.

src/main/recorder.ts
tdaugaard: As it did before. I've retained the status updates as `queueVideo()` _does_ have a slight delay of a few seconds which will be marked as "_Saving_".

src/main/recorder.ts
tdaugaard: Yeah, I kind of like this a lot too.

package.json
tdaugaard: No, this is a different queue type. waitQueue is a simple queue that you're processing yourself for the assertment of OBS signals; atomic-queue is a queue that starts workers to process data written to the queue.

tdaugaard: They are both queues, yes, but they work differently and for different purposes.

src/main/obsRecorder.ts
```diff
@@ -382,11 +382,16 @@ const assertNextSignal = async (value: string) => {
   console.debug("[OBS] Asserted OBS signal:", value);
 }
 
+const getObsLastRecording = (): string => {
```
aza547, line 385: Worth a comment here to specify exactly what is returned? e.g. Relative path / absolute path as a string?

aza547: Maybe even worth validating with path.resolve?

src/main/recorder.ts
```diff
@@ -51,6 +59,48 @@ type RecorderOptionsType = {
...
+            console.log("[Recorder] Video processing queue empty, running clean up.")
+
+            // Run the size monitor to ensure we stay within size limit.
+            // Need some maths to convert GB to bytes
+            runSizeMonitor(this._options.storageDir, this._options.maxStorage)
```
aza547, line 89: Should this be in the finish event?

src/main/recorder.ts
```diff
@@ -192,58 +241,53 @@ type RecorderOptionsType = {
...
             const isLongEnough = (metadata.duration - overrun) >= this._options.minEncounterDuration;
 
-            if ((!isRaid || isLongEnough) && !discardVideo) {
-                // Cut the video to length and write its metadata JSON file.
-                // Await for this to finish before we return to waiting state.
-                await this.finalizeVideo(metadata, outputFilename);
-            } else {
+            if (isRaid && !isLongEnough) {
                 console.info("[Recorder] Raid encounter was too short, discarding");
+            } else if (discardVideo) {
```
aza547, line 246: Does this discardVideo plumbing from `endRecording()` has a purpose? Can't see it in the code. If not, maybe should rip it out?

src/main/recorder.ts
```diff
@@ -192,58 +241,53 @@ type RecorderOptionsType = {
...
+                console.info("[Recorder] Discarding video");
+            } else {
+                const bufferFile = obsRecorder.getObsLastRecording();
+                if (bufferFile) {
+                    await this.queueVideo(bufferFile, metadata);
```
aza547, line 251: Why do we need to await here?

src/main/recorder.ts
```diff
@@ -192,58 +241,53 @@ type RecorderOptionsType = {
...
-            this.cleanupBuffer(1);
-
             // Refresh the GUI
             if (mainWindow) mainWindow.webContents.send('refreshState');
 
             // Restart the buffer recording ready for next game.
-            setTimeout(async () => {
-                this.startBuffer();
-            }, 5000)
+            this.startBuffer();
```
aza547, line 261: This makes me nervous. I think I've spent several afternoons scratching my head around this and found that if we don't wait at least some time between stopping and starting OBS things go wrong. 

I think previously it was 2 seconds, for some reason I changed it to 5 seconds. But this makes it 0 seconds. I think the symptom was the signalling going wrong. Not suggesting you change this immediately, but we'll need to carefully test this.

Worst case we can revert this change and the PR still massively improves things in this area anyway.

src/main/recorder.ts
aza547: OK, but the saving icon will disappear before the video becomes available now. That's a bit unintuitive I think - I like the idea of having it as a second icon still.

src/main/recorder.ts
aza547: We are probably indirectly waiting at least some time given the earlier await on the adding to queueing that I commented on above. If we need this to function though, this might be a bit fragile.

src/main/recorder.ts
tdaugaard: Yes, it will. How would you like it to display the status?

src/main/obsRecorder.ts
tdaugaard: I knew you were going to comment on it! :D I thought it was self-explanatory enough, but I'll add a comment :D

src/main/recorder.ts
tdaugaard: I find it more suitable to run it once the queue is idle/empty, but I honestly don't care when it's run. :)

src/main/recorder.ts
tdaugaard: I guess it doesn't anymore - it used to, but I can't find any references that uses it anymore.

src/main/recorder.ts
tdaugaard: We don't.

src/main/recorder.ts
tdaugaard: I will revert this - I removed it because all the logic that was depending on it has been removed/deprecated.

However, restarting the buffer here isn't any different from when it restarts every 5 minutes of idling, is it?

src/main/recorder.ts
aza547: Only slight preference as this way we avoid some condition where the queue gets backed up and we temporarily exceed the storage limit. Not going to lose sleep over it either way.

src/main/recorder.ts
aza547: That's a good question, and probably one I can work out myself with some frontend tinkering. I'm happy to merge this and handle that myself.

src/main/recorder.ts
aza547: I don't care enough about this in retrospect to hold up the PR. Lets get this in.

src/main/recorder.ts
tdaugaard: I was thinking to expand the status thing at some point anyway. I'd like it to show some information about what it's recording and such, like perhaps "_Tazavesh: Streets +17_" or the name of a raid boss - perhaps we could show an icon somewhere when a saving job starts and remove it when the queue becomes idle.

aza547: Think there is potential for something really nice too. For now I'm probably going to make a quick fix to display the saving icon as a secondary but we should discuss a future enhancement!

PR #198, opened 2022-10-09
Move all logic from obsRecorder.ts to recorder.ts

aza547:
This file is a remnant of days past!


src/main/recorder.ts
```diff
@@ -596,6 +609,32 @@ type RecorderOptionsType = {
         const closestResolution = getClosestResolution(availableResolutions, res);
         this.setSettingOBS('Video', paramString, closestResolution);
     }
+
+    async assertNextSignal (value: string) {
```
tdaugaard, line 613: Don't forget to mark them `private` when they aren't supposed to be called externally.

aza547:
im just going to close this, there have been a few changes since and it's probably easier to start again, this was only 10 mins or so work

PR #202, opened 2022-10-11
Feature: Add Game Capture

### LINKED ISSUE #2: Should use OBS game capture not monitor capture
http://underpop.online.fr/f/ffmpeg/help/gdigrab.htm.gz

Specifically: "title=window_title"

tdaugaard:
Replace monitor capture with game capture

- Add Settings for selecting recording mode (Game or Monitor)

- Create `createMonitorCaptureSource()`

- Create `createGameCaptureSource()` to create and configure an input source of `game_capture` with the appropriate settings for capturing any window named 'World of Warcraft'. This will also capture WowClassic.exe, WowB.exe, etc.

- Create `watchVideoSourceSize()` to periodically watch the video source for a change in its dimensions and adjust the scaling factor accordingly to fit the best in the selected output resolution.

Closes #2

### DIFF
--- a/src/main/obsRecorder.ts
+++ b/src/main/obsRecorder.ts
@@ -2,26 +2,28 @@ import { fixPathWhenPackaged, getAvailableDisplays } from "./util";
 import WaitQueue from 'wait-queue';
 import { getAvailableAudioInputDevices, getAvailableAudioOutputDevices } from "./obsAudioDeviceUtils";
 import { RecorderOptionsType } from "./recorder";
-import { OurDisplayType } from "./types";
 import { Size } from "electron";
-import { IScene } from "obs-studio-node";
 import path from 'path';
+import { inspectObject } from "./helpers";
+import { ISceneItem, IScene, IInput } from "obs-studio-node";
+import { OurDisplayType } from "./types";
+import { ISource } from "obs-studio-node";
 const waitQueue = new WaitQueue<any>();
 const osn = require("obs-studio-node");
 const { v4: uuid } = require('uuid');
 
 let obsInitialized = false;
-let scene = null;
+// Timer for periodically checking the size of the video source
+let timerVideoSourceSize: any;
+// Previous size of the video source as checked by checkVideoSourceSize()
+let lastVideoSourceSize: Size;
 
 /*
 * Reconfigure the recorder without destroying it.
 */
 const reconfigure = (options: RecorderOptionsType) => {
-  const baseResolution = parseResolutionsString(options.obsBaseResolution);
-  const outputResolution = parseResolutionsString(options.obsOutputResolution);
-
   configureOBS(options);
-  scene = setupScene(options.monitorIndex, baseResolution, outputResolution);
+  const scene = setupScene(options);
   setupSources(scene, options.audioInputDeviceId, options.audioOutputDeviceId);
 }
 
@@ -116,6 +118,7 @@ const configureOBS = (options: RecorderOptionsType) => {
   console.debug('[OBS] OBS Configured');
 }
 
+
 /*
 * Get information about primary display
 * @param zero starting monitor index
@@ -158,12 +161,12 @@ const getClosestResolution = (resolutions: string[], target: Size): string => {
 };
 
 /*
-* Given a none-whole monitor resolution, find the closest one that 
+* Given a none-whole monitor resolution, find the closest one that
 * OBS supports and set the corospoding setting in Video.Untitled.{paramString}
-* 
+*
 * @remarks
-* Useful when windows scaling is not set to 100% (125%, 150%, etc) on higher resolution monitors, 
-* meaning electron screen.getAllDisplays() will return a none integer scaleFactor, causing 
+* Useful when windows scaling is not set to 100% (125%, 150%, etc) on higher resolution monitors,
+* meaning electron screen.getAllDisplays() will return a none integer scaleFactor, causing
 * the calucated monitor resolution to be none-whole.
 *
 * @throws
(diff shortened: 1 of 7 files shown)


tdaugaard:
@aza547 it'd be great if you have time to test this out at some point and give some feedback, if any. I will continue testing myself as well.

aza547:
This just worked out the box for me which was awesome.

aza547:
This seems way more robust than I expected. Changing from fullscreen to windowed mid recording and also resizing my window was all just handled for us. That's awesome. Top worry just now is does this work for everyone else.

tdaugaard:
> There was a brief blip on the recording during the switch but totally acceptable

Yes, to avoid checking the size too often it only checks every 5 seconds - I assume that when a recording starts, the window will already have the size it will have throughout the entire recording and that changing the size _during_ the recording is only something we're doing to test it.

tdaugaard:
> Top worry just now is does this work for everyone else.

I'd be happy to reintroduce the monitor selection and allow both monitor and game capture so people can chose what they want.

I had originally done that and changed `monitorIndex` to `inputSource` with values like `monitor_capture:0` (monitorIndex = 1) and `game_capture`.

Up to you.

tdaugaard:
@aza547 I have reintroduced monitor capture. Feature is now:

- Add Game Capture
- Game Capture is default
- Monitor selection is hidden when Game Capture is selected
- Base recording resolution is set as such:
    - For monitors: The monitor resolution
    - For Game Capture: The selected output resolution
- `watchVideoSourceSize()` is enabled for both capture modes and for monitors will allow the change in resolution of a monitor, or the fix of selecting a different monitor with a different resolution.

aza547:
Sounds great, will look as soon as I have some time to!

src/main/obsRecorder.ts
```diff
@@ -174,44 +177,132 @@ const setOBSVideoResolution = (res: Size, paramString: string) => {
...
+ */
+const createGameCaptureSource = (): IInput => {
+  const videoSource = osn.InputFactory.create('game_capture', 'Game Capture');
+  const settings = videoSource.settings;
+
+  settings['capture_cursor'] = true;
+  settings['capture_mode'] = 'window';
+  settings['allow_transparency'] = true;
+  settings['priority'] = 1; // Window title must match
+  settings['window'] = 'World of Warcraft:GxWindowClass:Wow.exe';
```
aza547, line 241: Are all wow versions the same here? Classic/PTR/Beta? How can we confirm?

src/settings/VideoSettings.tsx
```diff
@@ -18,6 +18,10 @@ const obsResolutions: any = ipc.sendSync('settingsWindow', ['getObsAvailableReso
 const { Base: baseResolutions, Output: outputResolutions } = obsResolutions;
 
 const fpsOptions = ['10', '20', '30', '60'];
+const obsCaptureModes = {
+  'game_capture': 'Game (best)',
```
aza547, line 22: Can we reword this user facing text to:

`Game Capture (Recommended)`
`Monitor Capture`

src/main/obsRecorder.ts
```diff
@@ -174,44 +177,132 @@ const setOBSVideoResolution = (res: Size, paramString: string) => {
...
 
-  // A scene is necessary here to properly scale captured screen size to output video size
-  const scene = osn.SceneFactory.create('test-scene');
+    default:
+      throw Error(`[OBS] Invalid capture mode: ${options.obsCaptureMode}`);
+  }
+
+  setOBSVideoResolution(baseResolution, 'Base');
+
+  const scene: IScene = osn.SceneFactory.create('test-scene');
```
aza547, line 219: probably should stop calling this test-scene now?

src/main/obsRecorder.ts
```diff
@@ -174,44 +177,132 @@ const setOBSVideoResolution = (res: Size, paramString: string) => {
...
   const sceneItem = scene.add(videoSource);
   sceneItem.scale = { x: 1.0, y: 1.0 };
 
+  console.log(`[OBS] Configured video input source with mode '${options.obsCaptureMode}'`, inspectObject(videoSource.settings))
+
+  watchVideoSourceSize(sceneItem, videoSource, baseResolution);
```
aza547, line 225: Do we need to call this for monitor capture? Seems like it might just be desired for game capture?

src/main/obsRecorder.ts
```diff
@@ -174,44 +177,132 @@ const setOBSVideoResolution = (res: Size, paramString: string) => {
...
+
+    console.log("[OBS] Adjusting scene item scale due to video input source size change", inspectObject(logDetails));
+  }, 5000);
+};
+
 /*
 * setupSources
 */
 const setupSources = (scene: any, audioInputDeviceId: string, audioOutputDeviceId: string ) => {
+  clearSources();
```
aza547, line 304: Not clear to me what the need for this change is?

aza547 (changes requested):
Looks great overall. Minor comments. You mentioned we don't need the `watchVideoSourceSize` fn for monitor capture but it seems it's still in the code path? 

Main concern is do we need some logic for non-retail game-capture. That being said I think it just works for me on beta. Are we sure all the other types are going to just work also?

aza547:
Tested this again this morning, seems to record nothing on classic. I suspect that's due to the window naming I commented on in the review. It records a black screen. Opening retail and then re-testing worked fine.

src/main/obsRecorder.ts
aza547: Might be spill over from audio sources PR ?

src/main/obsRecorder.ts
tdaugaard: `priority` = 1 means it only considers "World of Warcraft" and as the window title, and doesn't look at "Wow.exe". I tested it for Classic now and it works fine.

If Beta/PTR hs a different Window title then it can't work unless that is exposed as a setting which would make it much less automatic.

image

src/settings/VideoSettings.tsx
tdaugaard: We can do whatever you want! =)

src/main/obsRecorder.ts
tdaugaard: Technically, we don't but if the resolution changes of the monitor even if the base resolution doesn't change, it will adjust the scaling so it still looks fine. I don't mind either way.

src/main/obsRecorder.ts
tdaugaard: No, not a spill over - intentionally ported over. We need to clear the output sources before we reconfigure them in order to properly release and remove them. It has worked before without it because we didn't change the video source. It's basically just cleaning up.

tdaugaard:
> Tested this again this morning, seems to record nothing on classic. I suspect that's due to the window naming I commented on in the review. It records a black screen. Opening retail and then re-testing worked fine.

I tested it today on Classic (WotLK) and it worked the first time on the test run. 🤷‍♀️ The window has the same name as retail.

image

tdaugaard:
Just tested on Dragonflight PTR - works fine.

aza547:
[2022-10-15 17-34-38.txt](https://github.com/aza547/wow-recorder/files/9792617/2022-10-15.17-34-38.txt)
[WarcraftRecorder-2022-10-15.log](https://github.com/aza547/wow-recorder/files/9792618/WarcraftRecorder-2022-10-15.log)

tdaugaard:
It was determined to be a leaky obs.exe running in the background that had hooked into WowClassic.exe and prevented the app from hooking into it, causing a black/blank recording.

aza547:
Great, let's get this in. 

Tweaked the description a bit for the new config param.

PR #204, opened 2022-10-12
Only record audio when recording to buffer

### LINKED ISSUE #57: Microphone streaming causes windows to not sleep/hibernate
Microphone recording stops windows from windows from sleeping on its own. From the terminal command "powercfg -requests" with the application in the task bar (WOW not running):

```
C:\WINDOWS\system32>powercfg -requests
DISPLAY:
None.

SYSTEM:
[DRIVER] Steam Streaming Microphone (ROOT\SteamStreamingMicrophone\0000)
An audio stream is currently in use.
[DRIVER] USB Audio 2.0 (USB\VID_30BE&PID_0101&MI_00\6&3c99bba&0&0000)
An audio stream is currently in use.
[DRIVER] USB Audio Device (USB\VID_0D8C&PID_9600&MI_00\6&2f66cdae&0&0000)
An audio stream is currently in use.
[DRIVER] Legacy Kernel Caller

AWAYMODE:
None.

EXECUTION:
None.

PERFBOOST:
None.

ACTIVELOCKSCREEN:
None.
```

After closing the wow-recorder app:

```
C:\WINDOWS\system32>powercfg -requests
DISPLAY:
None.

SYSTEM:
None.

AWAYMODE:
None.

EXECUTION:
None.

PERFBOOST:
None.

ACTIVELOCKSCREEN:
None.
```

tdaugaard:
To avoid Windows not going to sleep because we keep a microphone open, this change introduces a mechanism for only enabling the sound sources when the buffer restarting is canceled, e.g. when a recording is 'started'.

This _can_ result in a few seconds of missing audio in the beginning of the recording, but I _feel_ that's a fair sacrifice.

It works like this:

Start:

- `recorder.start()`
- Cancel buffer restart
- Setup audio sources (`obsRecorder.setupAudioSources()`)

Stop:

- `recorder.stop()`
- `obsRecorder.stop()`
- Clear and release audio sources

Further, simplify the filtering and adding of audio sources.

Fixes #57

### DIFF
--- a/src/main/obsRecorder.ts
+++ b/src/main/obsRecorder.ts
@@ -28,7 +28,6 @@ const reconfigure = (options: RecorderOptionsType) => {
 
   configureOBS();
   scene = setupScene();
-  setupSources();
 }
 
 /**
--- a/src/main/obsRecorder.ts
+++ b/src/main/obsRecorder.ts
@@ -1,6 +1,6 @@
 import { fixPathWhenPackaged, getAvailableDisplays } from "./util";
 import WaitQueue from 'wait-queue';
-import { getAvailableAudioInputDevices, getAvailableAudioOutputDevices } from "./obsAudioDeviceUtils";
+import { getAvailableAudioInputDevices, getAvailableAudioOutputDevices, ObsAudioDevice } from "./obsAudioDeviceUtils";
 import { RecorderOptionsType } from "./recorder";
 import { Size } from "electron";
 import path from 'path';
@@ -16,14 +16,19 @@ let obsInitialized = false;
 let timerVideoSourceSize: any;
 // Previous size of the video source as checked by checkVideoSourceSize()
 let lastVideoSourceSize: Size;
+let scene: IScene;
+let recorderOptions: RecorderOptionsType;
+let currentObsTrack: number;
 
 /*
 * Reconfigure the recorder without destroying it.
 */
 const reconfigure = (options: RecorderOptionsType) => {
-  configureOBS(options);
-  const scene = setupScene(options);
-  setupSources(scene, options.audioInputDeviceId, options.audioOutputDeviceId);
+  recorderOptions = options;
+
+  configureOBS();
+  scene = setupScene();
+  setupSources();
 }
 
 /**
@@ -88,26 +93,26 @@ const initOBS = () => {
 /*
 * configureOBS
 */
-const configureOBS = (options: RecorderOptionsType) => {
+const configureOBS = () => {
   console.debug('[OBS] Configuring OBS');
   setSetting('Output', 'Mode', 'Advanced');
 
   setObsRecEncoder(options);
 
   // Set output path and video format.
-  setSetting('Output', 'RecFilePath', options.bufferStorageDir);
+  setSetting('Output', 'RecFilePath', recorderOptions.bufferStorageDir);
   setSetting('Output', 'RecFormat', 'mp4');
 
   // VBR is "Variable Bit Rate", read about it here:
   // https://blog.mobcrush.com/using-the-right-rate-control-in-obs-for-streaming-or-recording-4737e22895ed
   setSetting('Output', 'Recrate_control', 'VBR');
-  setSetting('Output', 'Recbitrate', options.obsKBitRate * 1024);
+  setSetting('Output', 'Recbitrate', recorderOptions.obsKBitRate * 1024);
 
   // Without this, we'll never exceed the default max which is 5000.
   setSetting('Output', 'Recmax_bitrate', 300000);
   
   // FPS for the output video file. 
-  setSetting('Video', 'FPSCommon', options.obsFPS);
+  setSetting('Video', 'FPSCommon', recorderOptions.obsFPS);
 
   console.debug('[OBS] OBS Configured');
 }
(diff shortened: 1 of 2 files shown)


tdaugaard:
I'd like some testing and feedback in this.

It _could_ be changed so that it toggle when a WoW process is/isn't detected if that's a better solution.

aza547:
Thanks for looking at this. At a glance, I don't really like this solution I'm afraid, I don't think I want to sacrifice the first few seconds of audio in encounters for this. 

> It could be changed so that it toggle when a WoW process is/isn't detected if that's a better solution.

I agree this is better.

tdaugaard:
Mechanism changed such that it adds sources when the buffer starts recording and clears them when it stops.

That means when someone has it set to start up with Windows, it won't add the sources before WoW is detected and therefore not block Windows from going to sleep.

Also when WoW is closed, it'll release and remove the sources to bring it back to the state it was before WoW started.

aza547:
Tested this again this morning, seems to record nothing on classic. I suspect that's due to the window naming I commented on in the review. It records a black screen. Opening retail and then re-testing worked fine.

aza547:
> Tested this again this morning, seems to record nothing on classic. I suspect that's due to the window naming I commented on in the review. It records a black screen. Opening retail and then re-testing worked fine.

This is obviously the wrong PR for this comment...

src/main/obsRecorder.ts
```diff
@@ -300,48 +304,53 @@ const watchVideoSourceSize = (sceneItem: ISceneItem, videoSource: IInput, baseRe
...
-  console.log("[OBS] Removing all output sources")
+const setRecTracks = (): void => {
+  if (currentObsTrack == 1) {
+    setSetting('Output', 'RecTracks', 0);
+    return;
+  }
 
-  // OBS allows a maximum of 64 output sources
+  // Bit mask of used tracks: 1111 to use first four (from available six)
+  setSetting('Output', 'RecTracks', parseInt('1'.repeat(currentObsTrack - 1), 2));
```
aza547, line 337: this is fairly tricky to follow. Can we split this into two lines?

```
const audioBitMask = parseInt('1'.repeat(currentObsTrack - 1), 2);
setSetting('Output', 'RecTracks', audioBitMask);
```

src/main/obsRecorder.ts
```diff
@@ -300,48 +304,53 @@ const watchVideoSourceSize = (sceneItem: ISceneItem, videoSource: IInput, baseRe
...
+};
 
 /**
- * Clear all sources from the global output of OBS
+ * Set the Output.RecTracks setting for OBS
  */
- const clearSources = (): void => {
-  console.log("[OBS] Removing all output sources")
+const setRecTracks = (): void => {
+  if (currentObsTrack == 1) {
```
aza547, line 331: `===` ?

src/main/obsRecorder.ts
```diff
@@ -352,17 +361,22 @@ const setupSources = (scene: any, audioInputDeviceId: string, audioOutputDeviceI
     }
   }
 
-  setSetting('Output', 'RecTracks', 0); // Bit mask of used tracks: 1111 to use first four (from available six)
+  currentObsTrack = 1;
```
aza547, line 364: Can we make this code clearer than this? I think currently set to 1 means no audio? I don't have a good understanding of the OBS tracks or how this represents them.

aza547 (commented):
Finding this one a bit hard to follow, specifically the bitmask logic to enable the tracks. 

Maybe just need to come back and re-read it which I'll do later but interested in your thoughts on if this could be handled more clearly in the code.

src/main/obsRecorder.ts
tdaugaard: This is literally just a copied line, it was already there to behin with :p

src/main/obsRecorder.ts
tdaugaard: 1 doesn't mean much by itself, really. In our context, however, and currently it does mean that because we add the input source as the first track (1) and audio sources as tracks 2, 3, 4, 5, etc.

'1' is just the "beginning" as it's a 1-based index.

src/main/obsRecorder.ts
tdaugaard: I did split it up, though, since you asked for it.

src/main/obsRecorder.ts
tdaugaard: Obsolete.

src/main/obsRecorder.ts
aza547: Ah, hadn't realised. Apologies!

src/main/obsRecorder.ts
```diff
@@ -16,14 +16,19 @@ let obsInitialized = false;
...
 */
 const reconfigure = (options: RecorderOptionsType) => {
-  configureOBS(options);
-  const scene = setupScene(options);
-  setupSources(scene, options.audioInputDeviceId, options.audioOutputDeviceId);
+  recorderOptions = options;
+
+  configureOBS();
+  scene = setupScene();
+  setupSources();
```
aza547, line 31: Why do we need this? I think this is the same call we make when we start recording, doesn't this make a reconfigure prevent sleeping?

aza547:
Oh and worth a changelog entry here - user facing change

src/main/obsRecorder.ts
tdaugaard: Good catch, yes I believe you're right.

aza547:
I cut 2.10.0 and scanned last night so going to release (only relevant for CL I think - should go in unreleased)

tdaugaard:
> Oh and worth a changelog entry here - user facing change

There was changelog entry already :)

PR #209, opened 2022-10-15
Feature: Configurable OBS encoder

### LINKED ISSUE #205: Expose encoder in settings
Some people have problems with AMD encoding due to what looks like driver bugs. 

If we expose software encoding options also it might help with this.

tdaugaard:
Add OBS recording encoder setting

- Add `obsRecEncoder` to config schema and "Advanced Settings"

- Defaults to 'auto', which keeps the old logic of taking the last encoder of the list of available encoders.

- Fix race condition in stop/reconfigure obsRecorder due to not waiting for `obsrecorder.stop()` before calling `obsRecorder.reconfigure()`.

- Add Settings UI elements and helper functions

- Add styling to avoid the dropdown selector having black border/text

Fixes #205


aza547:
Also looks great. No complaints. Merging.

PR #212, opened 2022-10-16
Feature: Validate WoW log directories

### LINKED ISSUE #211: Feature: Validate combat log paths to avoid mistakes
As title says, validate the directory the user selects to ensure it's a valid combat log directory.

tdaugaard:
Add validation for WoW combat log paths such that there will no longer be any confusion about what it is that needs to be selected.

Add a guide in `docs/` that thoroughly explains how to find the appropriate log directory.

Improved styling for the `<InformationDialog/>` component to bring it in line with the rest of the app.

Fixes #211

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- [Issue 211](https://github.com/aza547/wow-recorder/issues/211) - Validate combat log paths to avoid mistakes.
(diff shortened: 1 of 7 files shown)


aza547:
Love it. Only comment is you could also pass the flavour through `validateLogPath` to confirm they haven't mixed up their retail / classic paths. However, that's minor, so I'm gonna hit approve.

aza547 (approved):

aza547:
Let me know if you'd like me to merge or hold off

tdaugaard:
> Love it. Only comment is you could also pass the flavour through validateLogPath to confirm they haven't mixed up their retail / classic paths. However, that's minor, so I'm gonna hit approve.

I was considering that, but we can't as we've told people to use the paths for other flavours like Beta/PTR and such which would fail validation.

Issue #164, opened 2022-09-26
Remaining actions after settings revamp


aza547:
Out of time to make finishing touches on settings page and want to release so people can test fixes.  Copying notes from the PR. 

TODO

- Attempt to centralise modifyConfig methods in frontend code.
- Attempt to fix remaining typescript errors in frontend code.
- Any final clean-up we can do / move things to final locations.
- Investigate errors like Error occurred in handler for 'getVideoState': Error: ENOENT: no such file or directory, stat 'D:\wow-recorder-files\2022-09-19 12-28-17.mp4. I think this is happening when the sizemonitor and clean-up after a recording overlap.

aza547:
[tdaugaard](https://github.com/tdaugaard) [22 hours ago](https://github.com/aza547/wow-recorder/pull/161#discussion_r979463372)
I'm thinking we make this an array like recordCategories which just has the labels like "2v2", "Mythic+" but it's not super important.

aza547:
Also can we expose the schema help text in the UI, save people getting config wrong

aza547:
"Make content type config actually do something." 
This was a bit hacky but I'm not totally sure how to do what I want with the types.
Hoping you can educate me tdaugaard.
Talking about the arena bit here where I had to make a mapping.

See categoryRecordConfigMapping for this one.

aza547:
Double setting of cfg params in logs:
```
20:57:31.213 > [Config Service] Set in config store: selectedCategory 6
20:57:31.218 > [Config Service] Set in config store: selectedCategory 6
20:57:37.912 > [Config Service] Set in config store: selectedCategory 5
20:57:37.915 > [Config Service] Set in config store: selectedCategory 5
20:57:39.656 > [Config Service] Set in config store: selectedCategory 6
20:57:39.660 > [Config Service] Set in config store: selectedCategory 6
```

tdaugaard:
> Investigate errors like Error occurred in handler for 'getVideoState': Error: ENOENT: no such file or directory, stat 'D:\wow-recorder-files\2022-09-19 12-28-17.mp4. I think this is happening when the sizemonitor and clean-up after a recording overlap.

I think this may have been remedied by changing to use `getObsLastRecording()`?

aza547:
I think we're done here?

PR #214, opened 2022-10-17
Refactor modifyConfig()

tdaugaard:
- Give all settings fields a `name` property that matches their state/configuration key since it's included in `event.target.name` which gives us exactly the information we need for the new `modifyConfig()`.

- Add `onChange`, `config` properties to the settings components that provides it with the current configuration as well as a method for changing the config in a central place.

- Remove some extranous whitespace

- Add `FakeChangeEvent` class to mock an event value for when we're saving configuration values delayed from the original change event.

- Remove `ConfigContext`/`SettingsContext`

Adresses parts of #164

### DIFF
--- a/src/settings/Settings.tsx
+++ b/src/settings/Settings.tsx
@@ -9,8 +9,8 @@ import VideoSettings from './VideoSettings';
 import AudioSettings from './AudioSettings';
 import AdvancedSettings from './AdvancedSettings';
 import ContentSettings from './ContentSettings';
-import ConfigContext from "./ConfigContext";
 import useSettings, { setConfigValues } from "./useSettings";
+import { configSchema } from 'main/configSchema';
 
 const ipc = window.electron.ipcRenderer;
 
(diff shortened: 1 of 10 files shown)


src/settings/Settings.tsx
```diff
@@ -116,46 +134,44 @@ const useStyles = makeStyles()({
...
-      </Box>
-    </ConfigContext.Provider>
+        <Tab label="General" {...a11yProps(0)} sx = {{ ...categoryTabSx }} />
+        <Tab label="Content" {...a11yProps(1)} sx = {{ ...categoryTabSx }}/>
+        <Tab label="Video" {...a11yProps(2)} sx = {{ ...categoryTabSx }}/>
+        <Tab label="Audio" {...a11yProps(3)} sx = {{ ...categoryTabSx }}/>
+        <Tab label="Advanced" {...a11yProps(4)} sx = {{ ...categoryTabSx }}/>
+      </Tabs>
+      <TabPanel value={tabIndex} index={0}>
+        <GeneralSettings config={config} onChange={modifyConfig}/>
```
aza547, line 157: hah, this is what all the tutorials I read when implementing this initially were telling me to do, pass down the prop... and I didn't get it and implemented the configContext stuff. A good learning experience atleast.

aza547 (approved):
This looks great and a huge simplification on the mess of code I left with the initial implementation.

PR #215, opened 2022-10-22
Draft: Classic Arenas

aza547:
- This works to some extent. I'd like to clean it up before merging.
- I think some sort of refactor of the combatLog class is needed. 
- Specifically to create a classic and retail combatLog class.
- There is scope to write quite a bit more for classic as classic combat logs are woefully poor. There is no COMBATANT_DATA event, or event an ARENA_MATCH_START/STOP event. 
- Despite this I have ideas on how we can get most of the information we get on retail. But to do that I think we need a classicCombatLog class to hold the logic and logically separate it from retail. How much of that logic we actually right for initial support here is more flexible. I think the minimum is to count the player units present between the ZONE_CHANGE events to work out if it's 2v2, 3v3 etc -- that way we can slot it into the pre-existing tab and not need a "Classic Arenas" tab, that is a stopgap for now. 
- Also need better logic to cut the video. On retail the ARENA_MATCH_END seems to always hit at the right time, regardless of log flushing. That means we don't need to care about the end point in the video. It's simply the end. In classic, the ZONE_CHANGE event to signify arena match has ended does not have this nice behaviour.


tdaugaard:
> Specifically to create a classic and retail combatLog class.

I would avoid that _at all costs_. If there really are _too_ many changes needed that it can't realistically be implemented in the existing class, let's make a `CombatLogBase` class and a `CombatLogClassic`/`CombatLogRetail` that extends the base class, with as much common functionality in the base class.

aza547:
> > Specifically to create a classic and retail combatLog class.
> 
> I would avoid that _at all costs_. If there really are _too_ many changes needed that it can't realistically be implemented in the existing class, let's make a `CombatLogParserBase` class and a `CombatLogParserClassic`/`CombatLogParserRetail` that extends the base class, with as much common functionality in the base class.

I was experimenting with this last night and came to the same conclusion. Although I planned to call them "LogHandler[...]" so that we could continue to use the current CombatLogParser (which just processes lines into arguments) and have logic in the LogHandler classes to replace the `handleArenaStartLine` etc functions that exist just now.

tdaugaard:
> CombatLogParser (which just processes lines into arguments) and have logic in the LogHandler classes to replace the handleArenaStartLine etc functions that exist just now.

Oh, yes of course - so you'd want to break out the event handlers from logutils into handler classes for retail and classic?

That's not a terrible idea. Should be fairly simple, honestly.

aza547:
> > CombatLogParser (which just processes lines into arguments) and have logic in the LogHandler classes to replace the handleArenaStartLine etc functions that exist just now.
> 
> Oh, yes of course - so you'd want to break out the event handlers from logutils into handler classes for retail and classic?
> 
> That's not a terrible idea. Should be fairly simple, honestly.

Yeah exactly. I'll probably have a go at this in the next few days. Saves having a bunch of variables line `isClassic` all over the place in the handler methods.

aza547:
Closing this as it's being included in #217

PR #217, opened 2022-10-30
LogHandlers, Activitys, Classic Arenas, and lots of refactoring

aza547:
# Overview
I've been bashing away at this for a while. Lots of changes and this is still weeks away from merge probably. Fundamentally there is loads of restructuring here to attempt to make life easier for future development -- in particular the design of logutils.ts is lacking.

# LogHandler, RetailLogHandler and ClassicLogHandler
- The model is now: `LogHandler` for shared logic between retail and classic. Respective handler for unique logic. 
- These classes which should probably end up as singletons now listen to the CombatLogParser for events. 
- They create, add to and forget `Activitys` when they recognised. 
- They also start and stop the recorder. 

# Activity and Subclasses 
- Activity class is an Abstract class representing any content type.
- Subclasses such as `RaidEncounter` or `ChallengeModeDungeon` extend this with custom logic. 
- These classes are agnostic of game flavour. They should not contain any log parsing logic. 

# What's to come? 
- This structure has much better encapsulation and will allow for easier writing of UTs. 
- Removal of Metadata type entirely?  Not sure on this yet. 
- Classic arena support, as it should now be easy to write lots of ZONE_CHANGE custom code for classic. 
- TODOs are marked with `@@@`
- Logging is a bit much just now, need to tone it down. 
- Comments! 
- Backwards compatibility with old metadata files might be scuppered. TBD if I care to fix this.


aza547:
@tdaugaard I'd love to get your opinion on this if you have a chance.

tdaugaard:
> Removal of Metadata type entirely? Not sure on this yet.

What would be purpose of that be? `Metadata` type is in dire need of a cleanup - there's a lot of duplicate stuff, like `playerSpec`/`playerName`, etc. instead of just `player` with a `Player` object

> Backwards compatibility with old metadata files might be scuppered. TBD if I care to fix this.

Make metadata versioned, keep a "version" key in the metadata and rewrite/transform to current format if you encounter any old ones. I na few versions you could drop it entirely since most would've upgraded anyway.

src/activitys/Activity.ts
```diff
@@ -0,0 +1,169 @@
...
+        this.category = category;
+        this.playerDeaths = [];
+    }
+
+    end(endDate: Date, result: boolean) {
+        this.endDate = endDate;
+        this.result = result;
+    }
+
+    getZoneID() {
```
tdaugaard, line 33: This could be a getter like this, but it obviously doesn't have to be.

```
get zoneID(): number { return this.zoneID }
```

src/activitys/Activity.ts
```diff
@@ -0,0 +1,169 @@
...
+        return this.zoneID;
+    }
+
+    setZoneID(zoneID: number) {
```
tdaugaard, line 37: Same as above, but a setter.

src/activitys/Activity.ts
```diff
@@ -0,0 +1,169 @@
...
+        this.zoneID = zoneID; 
+    }
+
+    getDuration() {
+        if (!this.endDate) {
+            console.error("[Activity] Failed to get duration of in-progress activity");
```
tdaugaard, line 43: I don't think classes should log much. If you throw an exception, let the catcher handle the logging.

src/activitys/Activity.ts
```diff
@@ -0,0 +1,169 @@
...
+            throw new Error("[Activity] Failed to get duration of in-progress activity");
+        }
+
+        const duration = (this.endDate.getTime() - this.startDate.getTime()) / 1000;
```
tdaugaard, line 47: No reason to use a variable here really:

```ts
return (this.endDate.getTime() - this.startDate.getTime()) / 1000;
```

src/activitys/Activity.ts
```diff
@@ -0,0 +1,169 @@
...
+        return duration;
+    }
+
+    getCategory() {
```
tdaugaard, line 51: All these static get/set things could be get/setters

src/activitys/Activity.ts
```diff
@@ -0,0 +1,169 @@
...
+    addPlayerDeath(death: PlayerDeathType) {
+        this.playerDeaths.push(death);
+    }
+
+    getPlayerDeaths() {
+        return this.playerDeaths;
+    }
+
+    // overriden by subclasses
+    getMetadata(): Metadata {
```
tdaugaard, line 84: If it's overriden by the subclass, it should probably be an abstract method instead of having an implementation.

src/activitys/Activity.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    getPlayerGUID() {
+        return this.playerGUID; 
+    }
+
+    setPlayerGUID(guid: string) {
+        this.playerGUID = guid; 
+    }
+    
+    getPlayerName(): string {
```
tdaugaard, line 100: `getPlayerName`/`getPlayerRealm`/`getPlayerSpecId` should ideally just be `getPlayer` that return playerCombatant - no reason to have so much identical code for no benefit.

src/activitys/ArenaMatch.ts
```diff
@@ -0,0 +1,56 @@
...
+        };
+
+        const playerCombatant = this.getCombatant(this.playerGUID);
+
+        if (!playerCombatant) {
+            console.error("[RetailLogHandler] No player combatant so no results possible");
+            return false;
+        }
+
+        const teamID = playerCombatant.teamID;
```
tdaugaard, line 35: This reads much more cleanly:

```ts
return (playerCombatant.teamID === winningTeamID)
```

src/activitys/ArenaMatch.ts
```diff
@@ -0,0 +1,56 @@
...
+        const result = (teamID === winningTeamID);
+        return result;
+    }
+
+    getMetadata(): Metadata {
+        const metadata: Metadata = {
```
tdaugaard, line 41: Return it straight away instead of assigning it first.

src/activitys/Battleground.ts
```diff
@@ -0,0 +1,28 @@
...
+export default class Battleground extends Activity {
+    constructor(startDate: Date, 
+                category: VideoCategory, 
+                zoneID: number) 
+    {
+        super(startDate, category);
+        this.zoneID = zoneID;
+    }
+
+    getMetadata(): Metadata {
```
tdaugaard, line 17: Return it straight away instead of assigning to temp variable.

src/activitys/ChallengeModeDungeon.ts
```diff
@@ -0,0 +1,158 @@
...
+import { ChallengeModeTimelineSegment, TimelineSegmentType } from "../main/keystone";
+import Activity from "./Activity";
+
+export default class ChallengeModeDungeon extends Activity {
+    private mapID: number;
+    private level: number;
+    private affixes: number[];
+    private timings: number[];
+    private CMDuration: number = 0;
+    private chests?: number;
```
tdaugaard, line 12: Better to name this `upgradeLevel`. "chests" is not an official term for anything. =D

src/activitys/ChallengeModeDungeon.ts
```diff
@@ -0,0 +1,158 @@
+import { Metadata } from "main/types";
+import { dungeonTimersByMapId, VideoCategory } from "../main/constants";
+import { ChallengeModeTimelineSegment, TimelineSegmentType } from "../main/keystone";
+import Activity from "./Activity";
+
+export default class ChallengeModeDungeon extends Activity {
+    private mapID: number;
+    private level: number;
+    private affixes: number[];
+    private timings: number[];
```
tdaugaard, line 10: We don't actually need this anymore; we need to just remove that and calculate the upgradeLevel and save that with the metadata - that way all that logic I made some time ago regarding this will be moot and could be removed.

You don't have to do that now, though, I can do that at a later time when you're done with this.

src/activitys/ChallengeModeDungeon.ts
```diff
@@ -0,0 +1,158 @@
...
+        const lastTimelineSegment = this.getCurrentTimelineSegment();
+
+        if (lastTimelineSegment && lastTimelineSegment.length() < 10000) {
+            console.debug("[ChallengeModeDungeon] Removing last timeline segment because it's too short.");
+            this.removeLastTimelineSegment();
+        }
+
+        this.endDate = endDate;
+        this.CMDuration = CMDuration;
+        this.result = true; // @@@ is this OK or need to handle non completed better?
```
tdaugaard, line 43: Basically:

```ts
this.result = this.chests > 0
```

src/activitys/ChallengeModeDungeon.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Pop the last timeline segment and discard it
+     */
+    removeLastTimelineSegment() {
+        this.timelineSegments.pop();
+    }
+
+    getChests() {
```
tdaugaard, line 123: Could be getters instead of methods.

src/activitys/ChallengeModeDungeon.ts
```diff
@@ -0,0 +1,158 @@
...
+    getMapID() {
+        return this.mapID;
+    }
+
+    getAllTimelineSegment() {
+        return this.timelineSegments;
+    }
+
+    getMetadata(): Metadata {
+        const metadata: Metadata = {
```
tdaugaard, line 140: Return it straight away instead of assigning to a variable.

src/activitys/RaidEncounter.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Class representing a raid encounter.
+ */
+export default class RaidEncounter extends Activity {
+    private difficultyID: number = 0;
+    private encounterID: number = 0;
+
+    constructor(startDate: Date, 
+                category: VideoCategory, //wtf alex
```
tdaugaard, line 13: Yeah, wtf? :D Why wtf?

src/log_handling/RetailLogHandler.ts
```diff
@@ -0,0 +1,344 @@
...
+        challengeModeActivity.endChallengeMode(endDate, CMDuration);
+        this.endRecording(this.activity);
+    };
+
+    handleEncounterStartLine(line: LogLine) {
+        console.debug("[RetailLogHandler] Handling ENCOUNTER_START line:", line);
+
+        if (!this.activity) {
+            super.handleEncounterStartLine(line);
+        } else {
```
tdaugaard, line 134: Rewrite this to remove the else, insert a return on line 134 and outdent everything under it by one.

src/log_handling/RetailLogHandler.ts
```diff
@@ -0,0 +1,344 @@
...
+            const category = this.activity.getCategory();
+            const isChallengeMode = category === VideoCategory.MythicPlus;
+
+            if (isChallengeMode) {
```
tdaugaard, line 138: Rewrite to:

```ts
if (!isChallengeMode) {
    return;
}
```

and outdent everything under it by one.

src/log_handling/RetailLogHandler.ts
```diff
@@ -0,0 +1,344 @@
...
+            }
+        }
+    }
+
+    handleZoneChange(line: LogLine) {
+        console.info("[RetailLogHandler] Handling ZONE_CHANGE line:", line);
+        const zoneID = parseInt(line.arg(1), 10);
+        const isZoneBG = retailBattlegrounds.hasOwnProperty(zoneID);
+
+        if (this.activity)
```
tdaugaard, line 196: These two cases could probably be split into two private methods that `handleZonechange()` could call to keep the line count and complexity down.

src/log_handling/RetailLogHandler.ts
```diff
@@ -0,0 +1,344 @@
...
+                activeChallengeMode.addTimelineSegment(segment, eventDate);
+                console.debug(`[RetailLogHandler] Starting new boss encounter: ${dungeonEncounters[encounterID]}`)
+            }
+        }
+    }
+
+    handleEncounterEndLine(line: LogLine) {
+        console.debug("[RetailLogHandler] Handling ENCOUNTER_END line:", line);
+
+        if (this.activity) {
```
tdaugaard, line 159: Same as above, return and fail early and often.

src/log_handling/RetailLogHandler.ts
```diff
@@ -0,0 +1,344 @@
...
+        return relativeTime;
+    };
+
+    dataTimeout(ms: number) {
+        super.dataTimeout(ms);
+        const isDungeon = this.activity?.getCategory() === VideoCategory.MythicPlus;
+
+        if (isDungeon) {
+            this.forceStopRecording();
+            return;
```
tdaugaard, line 316: Unecessary `return`.

src/main/types.ts
```diff
@@ -157,6 +159,29 @@ interface ISettingsPanelProps {
...
+    difficultyID?: number;
+    playerName?: string;
+    playerRealm?: string;
+    playerSpecID?: number;
+    teamMMR?: number;
+    challengeMode?: ChallengeModeDungeon;
+    playerDeaths?: PlayerDeathType[];
+    chests?: number;
+    mapID?: number;
+    timeline?: ChallengeModeTimelineSegment[];
```
tdaugaard, line 181: This is already present in `ChallengeModeDungeon` - if you want to decouple it from that, you should refactor it out entirely and rename it, probably.

src/main/util.ts
```diff
@@ -136,10 +137,10 @@ const getMetadataForVideo = (video: string) => {
 /**
  * Writes video metadata asynchronously and returns a Promise
  */
- const writeMetadataFile = async (videoPath: string, metadata: any) => {
+ const writeMetadataFile = async (videoPath: string, metadata: Metadata) => {
```
tdaugaard, line 140: This needs to be `any` because this method is also called when you toggle "Protected" on a video and at that point the structure is NOT `Metadata`. Alternatively, and probably better, `Metadata` could be refactored to an interface `IVideoMetadata` which would then be OK since it just needs a compatible structure to pass the type check.

src/activitys/ChallengeModeDungeon.ts
```diff
@@ -0,0 +1,158 @@
...
+            name: "a dungeon", // Instance name (e.g. "Operation: Mechagon")
+            category: VideoCategory.MythicPlus,
+            zoneID: this.getZoneID(),
+            mapID: this.getMapID(),
```
tdaugaard, line 144: `Metadata` doesn't have `mapID`, nor `chests`.

src/renderer/VideoButton.tsx
```diff
@@ -168,32 +169,34 @@ export default function VideoButton(props: any) {
 
   const buttonClasses = ['videoButton'];
   let keystoneTimelineSegments = [];
-
+  
   if (isMythicPlus) {
+    console.log("ahk");
```
tdaugaard, line 174: ahk? ahk!

src/renderer/rendererutils.ts
```diff
@@ -1,9 +1,5 @@
 const getVideoResult = (video: any): boolean => {
```
tdaugaard, line 1: You should remove it entirely at this point when it only returns whatever it's passed.

tdaugaard:
I think I'm through it now - it's a very large PR with a lot of changes that's a bit difficult to completely follow what changes affects what and in which way.

aza547:
> I think I'm through it now - it's a very large PR with a lot of changes that's a bit difficult to completely follow what changes affects what and in which way.

Thanks for all the detail. I should have been clearer that I was just looking for a 'code structure' style review rather than the implementation, but your comments look valuable none the less!

src/activitys/ChallengeModeDungeon.ts
aza547: Not sure I'm with you here... it's possible to complete a key with 0 upgradeLevel right?

aza547: i.e. depleted but still killed all the bosses

src/activitys/RaidEncounter.ts
aza547: not sure why I didn't just fix this, probably safe to assume RaidEncounter is VideoCategory.Raid...

src/main/util.ts
aza547: Ah I think this bit me and caused lots of confusion. Thanks for pointing it out.

aza547: I'd like to improve this with the interface as part of this so keeping open

src/renderer/VideoButton.tsx
aza547: Printing your initials is better than "hello" right?

src/renderer/rendererutils.ts
aza547: hilarious - how did I not think of this

src/log_handling/RetailLogHandler.ts
aza547: Yeah agree I should do this. Pretty ugly in its current form.

aza547:
> > Removal of Metadata type entirely? Not sure on this yet.
> 
> What would be purpose of that be? `Metadata` type is in dire need of a cleanup - there's a lot of duplicate stuff, like `playerSpec`/`playerName`, etc. instead of just `player` with a `Player` object
> 
> > Backwards compatibility with old metadata files might be scuppered. TBD if I care to fix this.
> 
> Make metadata versioned, keep a "version" key in the metadata and rewrite/transform to current format if you encounter any old ones. I na few versions you could drop it entirely since most would've upgraded anyway.

I had in mind we could have a type for each activity rather than just a single type with loads of optional fields. 

Making metadata versioned sounds like pain I'd rather not go through but will have a think about it.

src/main/types.ts
aza547: my plan was to pull it out as including the entire ChallengeModeDungeon object in the metadata seems a bit odd I thought, open to being convinced either way

aza547: (especially as I'm changing it as part of this PR - and might change it in future too)

src/activitys/ChallengeModeDungeon.ts
tdaugaard: Yes, you're right - I misunderstood the intention - `this.result` replaces the old `Metadata.result` and `this.chests` is basically equivalent to the old `ChallengeModeDungeon.result`. No problem then :)

src/main/types.ts
tdaugaard: I think it's totally fine to pull it out of CM and let it be generic as I _did_ want to maybe use it for raids too to have boss phases specified, although after looking through the source code for WCL that is a really time consuming task as it's usually a case of boss HP, boss cast, or some other weird thing that separates phases.

tdaugaard:
> I had in mind we could have a type for each activity rather than just a single type with loads of optional fields.

That makes typing it very difficult, so unless you're okay with NOT having a metadata type/interface I guess it's not a problem.

Perhaps make a `IVideoMetadata` interface that keeps common properties that are required for all, as I'm sure there will be a common set of properties all videos have - I _believe_, but am not certain, that TypeScript will pass the type check if you type hint an interface and the given value _at least_ satisfies the interface, but could contain extra properties as well.

> Making metadata versioned sounds like pain I'd rather not go through but will have a think about it.

I agree. I guess it's not super important to keep old videos available through the interface - we're naming them appropriately now and the app _is_ supposed to be more of an instant replay more than an archival app anyway.

aza547:
I'm targeting next weekend to get this merged. 

Classic arena is mostly working with this. Some cool log parsing to make that happen:
- We don't get COMBATANT_INFO, but I'm piecing most of it together from the SPELL_AURA_APPLIED Lines.
- We don't get told the bracket, so we assume 2v2 and correct it at the end based on the number combatants.  
- We don't get result info, but I'm counting the friendly vs enemy deaths to decide. 

Probably will finish writing UTs after merge - I did a simple one as proof of concept.

aza547:
This is a massive PR now. I'm merging it because I can. 

I've run pretty extensive tests, it's all working as far as I know.

Still to address a few open issues above, but will do that on main.

PR #241, opened 2022-12-09
Unit tests

aza547:
Stuff to make the code more unit testable, not actually added real UTs beyond a POC yet.


aza547:
This is risky but whatever

PR #242, opened 2022-12-10
Autoplay videos after they are selected

MarkyMarkMcDonald:
This makes videos auto play after they're selected, but not on first boot or when changing categories

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,4 +41,4 @@ This project is free and open source. It is built on the [electron-react-boilerp
 
 # Contributing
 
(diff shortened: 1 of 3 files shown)


aza547:
Thanks - this is great. One comment is that I think we also want to reset autoplay to false when the GUI is kicked to refresh after a new video is recorded (else the behaviour is to replay the currently playing video). Not a big change so I'll just add it and merge :)

aza547 (approved):

PR #243, opened 2022-12-12
Change import to let TS pull in obs node types

prsauer:


aza547 (approved):
Thanks for this. Looks like it will make life way easier.

PR #290, opened 2023-01-18
251: Weird cropping of M+ dungeon names

jhwess:
* Adds css selector for child divs of .videoButton to show overflowing text as ellipsis.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [Issue 285](https://github.com/aza547/wow-recorder/issues/285) - Fix bug that prevented retail recording of retail war games. 
 - [Issue 288](https://github.com/aza547/wow-recorder/issues/288) - Fix bug that prevented changing the FPS setting.
 - [Issue 275](https://github.com/aza547/wow-recorder/issues/275) - Increase retail log timeout for better handling of M+.
+- [Issue 251](https://github.com/aza547/wow-recorder/issues/251) - Fix text overflow clipping in video selection buttons.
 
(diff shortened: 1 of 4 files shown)


jhwess:
image

aza547:
Thanks for this - it's great to have more people involved! 

A few thoughts: 
- Should remove code at ~335 in videoButton.tsx that slices the string as no longer required, that was my initial attempts to make this work.

- Can we apply the same thing to character names? I think they could be a little too long also. Think the issue had an example.

- Any other fields worth worrying about? Can't think of any off the top of my head.

- Worth including a changelog entry too as this is a user facing change.

Hopefully formatting isn't too garbled as I'm on mobile. Cheers!

jhwess:
No problem, hyped to help out!
* will do
* These changes already apply the styling to any div that is a child of any element with the "videoButton" class which should cover any other potentially-problematic fields here.
  * Tested this with player names by manually editing the html with the dev tools:
image
* Ah yeah I'll do that as well with the next commit

If we wanted to use a css preprocessor, we could extract this into a re-usable set of styles (IIRC they're called mixins in SCSS) because it's a pretty common use case.

jhwess:
Addressed PR feedback and resolved conflicts

aza547:
> No problem, hyped to help out!
> 
> * will do
> * These changes already apply the styling to any div that is a child of any element with the "videoButton" class which should cover any other potentially-problematic fields here.

Ah - duh - great, my mistake! 

>   
>   * Tested this with player names by manually editing the html with the dev tools:
>     image
> * Ah yeah I'll do that as well with the next commit

Looks great!

> 
> If we wanted to use a css preprocessor, we could extract this into a re-usable set of styles (IIRC they're called mixins in SCSS) because it's a pretty common use case.

Not something I've ever heard of - but keen to be educated :)

PR #305, opened 2023-01-23
282: Play mono microphones through stereo

jhwess:
* Add obsForceMono option to configSchema.
* Check obsForceMono in Recorder#addAudioSourcesOBS and sets the appropriate flag on the input device.
* Add switch input to AudioSettings for Force Mono (obsForceMono).
* Slight refactoring of AudioSettings.

### DIFF
--- a/src/main/Recorder.ts
+++ b/src/main/Recorder.ts
@@ -8,6 +8,7 @@ import {
   EOBSOutputSignal,
   ERecordingFormat,
   ERecordingState,
+  ESourceFlags,
 } from './obsEnums';
 import { deleteVideo, fixPathWhenPackaged, getSortedVideos } from './util';
 import { IOBSDevice, Metadata, RecStatus, TAudioSourceType } from './types';
@@ -493,6 +494,11 @@ export default class Recorder {
     this.audioInputDevices.forEach((device) => {
       const index = this.audioInputDevices.indexOf(device);
       const channel = this.audioInputChannels[index];
+
+      if (this.cfg.get('obsForceMono')) {
+        device.flags = ESourceFlags.ForceMono;
+      }
+
       this.addAudioSourceOBS(device, channel);
     });
 
--- a/src/main/configSchema.ts
+++ b/src/main/configSchema.ts
@@ -13,6 +13,7 @@ export type ConfigurationSchema = {
   startMinimized: boolean;
   obsOutputResolution: string;
   obsFPS: number;
+  obsForceMono: boolean;
   obsKBitRate: number;
   obsCaptureMode: string; // 'game_capture' or 'monitor_capture'
   obsRecEncoder: string;
@@ -108,6 +109,7 @@ export const configSchema = {
     type: 'string',
     default: '1920x1080',
   },
+
   obsFPS: {
     description:
       'The number of frames per second to record the video at. Lower FPS gives smaller video size, but also more choppy playback.',
@@ -115,6 +117,12 @@ export const configSchema = {
     default: 60,
     minimum: 15,
     maximum: 60,
+  },
+  obsForceMono: {
+    description:
+      'Whether to force the audio of your input device to mono. Enable if your microphone audio is only playing out of one stereo channel.',
+    type: 'boolean',
+    default: true,
   },
   obsKBitRate: {
     description:
--- a/src/settings/AudioSettings.tsx
+++ b/src/settings/AudioSettings.tsx
@@ -7,10 +7,16 @@ import Box from '@mui/material/Box';
 import InfoIcon from '@mui/icons-material/Info';
 import IconButton from '@mui/material/IconButton';
 import Tooltip from '@mui/material/Tooltip';
-import { FakeChangeEvent, IOBSDevice, ISettingsPanelProps } from 'main/types';
-import { Checkbox, Chip, ListItemText, OutlinedInput } from '@mui/material';
+import { IOBSDevice, ISettingsPanelProps } from 'main/types';
+import {
+  Checkbox,
+  Chip,
+  FormControlLabel,
+  ListItemText,
+  Switch,
+} from '@mui/material';
 import React from 'react';
-import { configSchema } from '../main/configSchema';
+import { configSchema, ConfigurationSchemaKey } from '../main/configSchema';
 
 const ipc = window.electron.ipcRenderer;
 
@@ -21,55 +27,12 @@ enum DeviceType {
 
 export default function GeneralSettings(props: ISettingsPanelProps) {
   const { config, onChange } = props;
-
-  let initialInputs: string[];
-
-  if (config.audioInputDevices) {
-    initialInputs = String(config.audioInputDevices).split(',');
-  } else {
-    initialInputs = [];
-  }
-
-  let initialOutputs: string[];
-
-  if (config.audioOutputDevices) {
-    initialOutputs = String(config.audioOutputDevices).split(',');
-  } else {
-    initialOutputs = [];
-  }
-
-  const [input, setInput] = React.useState<string[]>(initialInputs);
-  const [output, setOutput] = React.useState<string[]>(initialOutputs);
-
   const devices = ipc.sendSync('getAudioDevices', []);
-
   const inputDevices: IOBSDevice[] = devices.input;
   const outputDevices: IOBSDevice[] = devices.output;
-
   const availableAudioDevices = {
     input: inputDevices,
     output: outputDevices,
-  };
-
-  const handleMultiSelect = (
-    type: DeviceType,
-    event: SelectChangeEvent<string[]>
-  ) => {
-    const {
-      target: { value },
-    } = event;
-
-    if (type === DeviceType.INPUT) {
-      setInput(
-        // On autofill we get a stringified value.
-        typeof value === 'string' ? value.split(',') : value
-      );
-    } else {
-      setOutput(
-        // On autofill we get a stringified value.
-        typeof value === 'string' ? value.split(',') : value
-      );
-    }
   };
 
   const getDeviceDescription = (id: string) => {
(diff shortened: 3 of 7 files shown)


jhwess:
image

jhwess:
image

jhwess:
Force Mono off:
https://user-images.githubusercontent.com/13799608/213971904-7e4d089d-a98e-4d44-86ab-e5b9d6e3f0f4.mp4

jhwess:
Force Mono on:
https://user-images.githubusercontent.com/13799608/213971981-83982ee4-fe92-43dc-a42e-f856d3200ca1.mp4

src/main/configSchema.ts
```diff
@@ -116,6 +118,12 @@ export const configSchema = {
     minimum: 15,
     maximum: 60,
   },
+  obsForceMono: {
+    description:
+      'Whether to force the audio of your input device to mono. Enable if your microphone audio is only playing out of  one stereo channel.',
+    type: 'boolean',
+    default: false,
```
aza547, line 125: Should this be true? I can't imagine many people want their mic audio coming through only one speaker. What's the implication if it's already mono and we have obsForceMono enabled?

src/settings/AudioSettings.tsx
```diff
@@ -125,20 +134,41 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
...
     <Stack
       component="form"
       sx={{
-        '& > :not(style)': { m: 0, width: '50ch' },
+        '& > :not(style)': { width: '50ch' },
       }}
       noValidate
       autoComplete="off"
     >
+      <Box component="h3" sx={headingStyle}>
```
aza547, line 165: Minor gripe - I don't really like the look of this text (does it need to be here?)

src/settings/AudioSettings.tsx
```diff
@@ -200,10 +229,37 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
...
+            color: 'white',
+          }}
+        />
+        <Tooltip title={configSchema.obsForceMono.description}>
+          <IconButton>
+            <InfoIcon sx={{ color: 'white' }} />
+          </IconButton>
+        </Tooltip>
+      </Box>
+      <Box component="h3" sx={headingStyle}>
```
aza547, line 256: Ditto

src/settings/AudioSettings.tsx
```diff
@@ -200,10 +229,37 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
...
+        Output Settings
+      </Box>
       <Box component="span" sx={{ display: 'flex', alignItems: 'center' }}>
         <FormControl sx={{ my: 1 }}>
           <InputLabel id="demo-simple-select-label" sx={style}>
-            Out
+            Output Devices
```
aza547, line 262: I took the easy way out here a few weeks back and didn't solve a problem with the border overlap, and instead used some short text. 

I think the below looks a bit sad, so I'd prefer to keep this as "Out" (or better yet, solve the border problem). I spent an hour or so on it embarrassingly and didn't find the solution. 

image

Same comment on the input devices box.

src/settings/AudioSettings.tsx
```diff
@@ -200,10 +229,37 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
           </IconButton>
         </Tooltip>
       </Box>
+      <Box component="span" sx={{ display: 'flex', alignItems: 'center' }}>
+        <FormControlLabel
+          control={
+            <Switch
```
aza547, line 235: I really like this! I've used checkboxes on the other pages, I'm inclined to switch them all to "Switches", looks way better.

src/settings/AudioSettings.tsx
```diff
@@ -200,10 +229,37 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
...
+              sx={switchStyle}
+              checked={forceMono}
+              name={forceMonoConfigKey}
+              onChange={(isForceMono) => {
+                setForceMono(Boolean(isForceMono.target.checked));
+                onChange(isForceMono);
+              }}
+            />
+          }
+          label="Force Mono"
```
aza547, line 245: I'd maybe go with "Mono Input" rather than "Force Mono" here

src/main/configSchema.ts
```diff
@@ -116,6 +118,12 @@ export const configSchema = {
     minimum: 15,
     maximum: 60,
   },
+  obsForceMono: {
+    description:
+      'Whether to force the audio of your input device to mono. Enable if your microphone audio is only playing out of  one stereo channel.',
```
aza547, line 123: I think you can drop the "Whether to" here.
Also double space between "of  one"

src/settings/AudioSettings.tsx
```diff
@@ -125,20 +134,41 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
       color: '#bb4220',
     },
   };
+  const switchStyle = {
```
aza547, line 137: I'd love to pick your brain on how you came up with this sometime. This still looks like a foreign language to me. I've copied loads of similar things of SO and not understood any of it.

src/settings/AudioSettings.tsx
```diff
@@ -125,20 +134,41 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
...
+          backgroundColor: '#bb4220',
+          opacity: 1.0,
+        },
+      },
+      '&.Mui-disabled + .MuiSwitch-track': {
+        opacity: 0.5,
+      },
+    },
+  };
+  const headingStyle = {
```
aza547, line 151: I think this can go if we lose the headings.

src/settings/AudioSettings.tsx
```diff
@@ -98,18 +62,63 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
...
+  } else {
+    initialOutputs = [];
+  }
 
-    const knownOutputs = output.filter(isKnownDevice);
-    setOutput(knownOutputs);
-    onChange(new FakeChangeEvent('audioOutputDevices', knownOutputs));
-  }, []);
+  const forceMonoConfigKey: ConfigurationSchemaKey = 'obsForceMono';
+  const initialForceMono = config.obsForceMono ?? false;
```
aza547, line 100: Do we need ?? here? Config from the ConfigService should always default to something sensible I think, even if it's not present

src/main/Recorder.ts
```diff
@@ -484,6 +490,12 @@ export default class Recorder {
     this.audioInputDevices.forEach((device) => {
       const index = this.audioInputDevices.indexOf(device);
       const channel = this.audioInputChannels[index];
+
+      if (this.cfg.get('obsForceMono')) {
+        // value of osn.EsourceFlags.ForceMono
+        // eslint-disable-next-line no-bitwise
+        device.flags = 1 << 1;
```
aza547, line 497: I've had this problem with other enums from OSN. I think they are naively exporting enums on module.d.ts but that doesn't work AIUI (we chatted about this recently in the dev channel on discord). 

Can we just reproduce their enum in obsEnums and use that to avoid having this slightly arcane bitwise syntax? 

```
const enum ESourceFlags {
    Unbuffered = (1 << 0),
    ForceMono = (1 << 1)
}
```

&&

```
if (this.cfg.get('obsForceMono')) {
  device.flags = ESourceFlags.ForceMono;
}
```

Downside is they might change it upstream, but I'd prefer this regardless I think.

src/main/Recorder.ts
```diff
@@ -2,7 +2,13 @@ import { BrowserWindow } from 'electron';
 import path from 'path';
 import fs from 'fs';
 import * as osn from 'obs-studio-node';
-import { IInput, IScene, ISceneItem, ISource } from 'obs-studio-node';
+import {
+  ESourceFlags,
```
aza547, line 6: unused

src/main/configSchema.ts
```diff
@@ -108,6 +109,7 @@ export const configSchema = {
     type: 'string',
     default: '1920x1080',
   },
+
```
aza547, line 112: extra new line

aza547 (changes requested):
This looks awesome, and thanks for sharing the test videos! Apologies for some pedantic review comments, happy to accept after fix up.

src/settings/AudioSettings.tsx
```diff
@@ -21,57 +28,14 @@ enum DeviceType {
...
-    event: SelectChangeEvent<string[]>
-  ) => {
-    const {
-      target: { value },
-    } = event;
-
-    if (type === DeviceType.INPUT) {
-      setInput(
-        // On autofill we get a stringified value.
-        typeof value === 'string' ? value.split(',') : value
```
aza547, line 65: glad to lose this mess - thanks!

aza547:
Oh - and needs a changelog entry please :)

src/settings/AudioSettings.tsx
jhwess: Not really, I added the headers to differentiate the setting types but it's probably overkill for the amount of audio settings we currently have. I'll remove them and their styling variable.

src/settings/AudioSettings.tsx
jhwess: I'll take a crack at fixing the border issue. Looks like it could be resolved with css since it displays correctly when the input isn't active.

src/settings/AudioSettings.tsx
jhwess: No, good catch. I think I added it before I'd set the type of the config prop. Looks like I can remove some other similar conditionals for initializing values from config as well.

src/main/Recorder.ts
jhwess: Ok, I agree that's a better way to handle it. [This page](https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls) explains why it's required.
> Inlining enum values is straightforward at first, but comes with subtle implications. These pitfalls pertain to ambient const enums only (basically const enums in .d.ts files) and sharing them between projects, but if you are publishing or consuming .d.ts files, these pitfalls likely apply to you, because tsc --declaration transforms .ts files into .d.ts files.

Perhaps we could let OSN know they could [configure TS](https://www.typescriptlang.org/tsconfig#preserveConstEnums) in such a way where we could use them without duplicating them.

jhwess:
> Oh - and needs a changelog entry please :)

Oops, I knew I was forgetting something lol

src/settings/AudioSettings.tsx
jhwess: Yeah, agreed "Force Mono" isn't very clear.

src/settings/AudioSettings.tsx
jhwess: Sure! I was referencing some MUI docs to come up with it but as I understand it it's CSS passed as an object. There should be an easier way to do this through theming - I'll message you about that

src/settings/AudioSettings.tsx
```diff
@@ -253,25 +232,22 @@ export default function GeneralSettings(props: ISettingsPanelProps) {
...
             id="audio-output-device"
             multiple
             value={output}
-            label="Output"
+            label="Output Devices"
+            sx={selectStyle}
             onChange={(e) => {
               handleMultiSelect(DeviceType.OUTPUT, e);
             }}
-            input={<OutlinedInput label="Tag" />}
```
jhwess, line 219: I believe this was the culprit for the visual bug with the borders shown above. Not entirely sure why but I'm guessing it injects an html element somewhere that interferes with some css rules MUI uses for the select field
image

src/main/configSchema.ts
jhwess: I think one implication would be if the user has a stereo mic and we force it to mono, they would hear the output as mono and thus lose some quality in their recording. Seems like an edge case though and I agree setting it to default to true makes sense

src/settings/AudioSettings.tsx
jhwess: Was able to fix this visual bug - see [this comment](https://github.com/aza547/wow-recorder/pull/305#pullrequestreview-1276461478)

jhwess:
> Oh - and needs a changelog entry please :)

Added!

src/settings/AudioSettings.tsx
aza547: Awesome - nice one!

src/main/Recorder.ts
aza547: https://github.com/stream-labs/obs-studio-node/issues/1260

src/settings/AudioSettings.tsx
aza547: This one still needing addressed?

aza547:
I think one thing left to address - this looks great :)

jhwess:
> 

Still getting used to github's UI, just pushed up changes that address it

PR #233, opened 2022-11-23
WAL parsing proof of concept

prsauer:

PR #318, opened 2023-02-04
WAL Parser

aza547:
Basically the same as https://github.com/aza547/wow-recorder/pull/233, except on an up-to-date version of the WR codebase.


aza547:
Need to update test logs - WR tests are using an old version of COMBATANT_INFO which doesn't work with WAL.

PR #339, opened 2023-03-17
Use videoJS player

aza547:


aza547:
Still to make volume setting / mute setting persist like it used to

PR #349, opened 2023-03-31
UI Overhaul

aza547:


aza547:
Remaining ideas:
- Fix images for some categories. 
- Should we use thumbnail in video lists?
- Fix error messages.
- Add status popup tray. 
- Limit list items (and add a button to load more) for performance?

PR #357, opened 2023-04-04
Adjusting Video Marker Width for M+

ChueyB:
Marker segments duration increased based on metadata for M+


aza547:
Code changes look good. I'd love a screenshot of this, I had a look with my M+ test script and saw:

image

...but I think that's just a side effect of the test not having proper timestamps? 

I'd be nice to have the entire dungeon bar being either purple / orange as I think anything that's not a boss can be marked trash by default. Is that already happening outside of the test case? Could you share / check-in a JSON file from a real run along with this?

ChueyB:
electron_HkPki53AaB

For the most part, it is covering the entire bar. The only exception being if a boss is the final combat encounter then the remainder/overrun is the white/grey default as you can see at the end of that screenshot. I can always re-style the bar to the purple color if its M+ to cover the entire thing if that's the preference.

aza547 (approved):

PR #360, opened 2023-04-04
Added JKL Hotkeys

### LINKED ISSUE #334: Bind jkl to video controls like YouTube (back 10s, pause, and forward 10s respectively)
title

ChueyB:
J: Backwards 10
K: Play/Pause
L: Forward 10

Closes #334

### DIFF
--- a/src/renderer/Layout.tsx
+++ b/src/renderer/Layout.tsx
@@ -4,6 +4,7 @@ import { TNavigatorState, VideoPlayerSettings } from 'main/types';
 import { Button, List, ListItem, ListItemButton } from '@mui/material';
 import Player from 'video.js/dist/types/player';
 import { VideoJS } from './VideoJS';
+import 'videojs-hotkeys';
 import { VideoCategory } from '../types/VideoCategory';
 import VideoButton from './VideoButton';
 import { addVideoMarkers } from './rendererutils';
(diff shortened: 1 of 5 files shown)


src/renderer/Layout.tsx
```diff
@@ -128,6 +129,21 @@ const Layout: React.FC<IProps> = (props: IProps) => {
       fill: true,
       inactivityTimeout: 0,
       playbackRates: [0.25, 0.5, 1, 1.5, 2],
+      plugins: {
+        hotkeys: {
+          seekStep: 10,
+          enableModifiersForNumbers: false,
+          forwardKey(event: { which: number }) {
+            return event.which === 76 || event.which === 39;
```
aza547, line 137: I think the 74, 75 and 75 are just ASCII here. What's the 32, 37, 39 etc? 

Didn't see a reference for what this actually is on the videojs-hotkeys page, probably could do with a comment.

ChueyB:
Added a comment to the file. These are [keyCode events](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). 
keyCode

ChueyB:
Quick change of mind. Removed the comments and refactored to make it more intuitive and the result should not necessitate the need for comments

aza547 (approved):
Looks great!

PR #361, opened 2023-04-06
Bug Fix: Minimum Keystone Settings

ChueyB:
The current iteration allows you to press the down arrow to go into the negatives. Even though you can't save, this should prevent that from happening at all. 

Also changed default to 2 as that's where keystones start.


aza547 (approved):
Sweet!

PR #377, opened 2023-04-15
Information dialog, background component update

Ago95Dev:
Added sx prop at Line 231


aza547:
Got a screenshot of how this looks now? I got fed up of trying to find the app colour theme and so started [this](https://github.com/aza547/wow-recorder/blob/main/docs/Colors.md) to write down the commonly used colours). Might be best to use one of those?

Issue #395, opened 2023-05-01
Fix dragonflight s2 M+ timings


aza547:
I grabbed the base timer for timing/depleting but couldn't find the 2/3 chest timers. 

That needs to go in constants.ts when they are available.

PR #409, opened 2023-05-15
Update constants.ts to include correct DF S2 timers

fjavifabre:
Solves #395 

I updated constants to use proper timers for Season 2 Mythic plus dungeons

AFAIK they always follow the same rule for +2 +3 based on base timer (+1).

- +2 --> `0.8 * baseTimer`
- +3 --> `0.6 * baseTimer`

I used this old reddit [post](https://www.reddit.com/r/wow/comments/54okeh/questions_regarding_time_needed_to_get_2_or_3/) as reference first but checked again my own timed dungeons an it seems that they always follow that rule.

I would suggest to compute this +2 & +3 timers automatically to avoid mistakes in future seasons (even in current season if they decide to change the timers). I can create a different PR to change this in constants.ts if you don't mind.

I also fixed some mistakes in DF Season 1 timers (not sure if all old timers are OK, I only checked DF S1)


aza547:
Thanks, this looks good. No objection to computing the timers either.

aza547:
We do keep a changelog file that you could update for future commits, but I'll sort that for this one :)

PR #471, opened 2024-01-12
Docs improvement

### LINKED ISSUE #470: aspect ratio changes makes the video a little weird.
**Describe the bug**

I have an 8k display (5120x2160). When the video is taken, it squashes it horizontally, which makes everything look a bit odd.

**To Reproduce**
Just play the game on a very wide screen. Resulting videos are squished.

**Expected behavior**
You know.. not squished video.

**Screenshots**
Full screen video after it was taken:
image

Where there are black bars on the side.. there should be video.

justinabrahms:
Fixes #470


aza547:
Propose more generic advice here as setting up the scene is more than just the resolution.

aza547:
I'll close this and just do directly.

PR #486, opened 2024-04-04
State manager

aza547:


aza547:
Things still broken:
- Frontend state not getting pushed to cloud
- Need a solution for migrating users from old cloud json format to new

Issue #506, opened 2024-07-13
Delete button delay is a bit crap


aza547:
![image](https://github.com/user-attachments/assets/eeb5e148-48c1-44a2-b666-147368e2a459)

PR #507, opened 2024-07-26
fix: Delete videos on disk directly, mark for deletion on next refresh if that fails

StephixOne:
Attempts a fix for [Issue#506](https://github.com/aza547/wow-recorder/issues/506).

My Windows does not complain when deleting a file that may be open, even if I've played half of it in warcraftrecorder, it'll delete it just fine. As such, simply deleting the file seems to do the trick. I've also added a fallback to the old mechanism, of marking the files for deletion and letting the next refresh/restart take care of it, if the hard delete fails.

I've also reset the video player progress when a deletion happens, because being `x` seconds through a video, deleting it, then landing at `x` seconds in the one it drops you to felt a bit jarring. Probably not the most common use-case, but it shouldn't hurt.


src/main/Manager.ts
```diff
@@ -127,7 +128,8 @@ export default class Manager {
       valid: false,
       current: this.obsBaseCfg,
       get: (cfg: ConfigService) => getObsBaseConfig(cfg),
-      validate: async (config: ObsBaseConfig) => Manager.validateBaseCfg(config),
+      validate: async (config: ObsBaseConfig) =>
```
aza547, line 131: I think your prettier is somehow ignoring the `/* eslint-disable prettier/prettier */` above here?

aza547:
Looks great - more than happy with the compromise / fallback behaviour - no need to let perfect be the enemy of good.

Happy to merge / approve once the prettier / yarn stuff is fixed.

src/main/Manager.ts
StephixOne: Let me check on that.

package-lock.json
aza547, line 1: Still a huge diff here that I think you intended to revert?

StephixOne: My `npm` does that, as yarn uses its own lockfile, `yarn.lock`. What node version are you using?

aza547: I have a global npmrc messing with this. 

https://stackoverflow.com/questions/55384977/npm-install-removes-dev-true-from-several-packages-in-package-lock-json

Agree it's fine.

aza547:
Thanks for the fix!

PR #516, opened 2024-08-18
OSN 0.23.71 -> 0.24.43

aza547:
Supersedes https://github.com/aza547/wow-recorder/pull/469.

I tried to go to the latest 0.24.45, but it was buggy and spamming logs complaining about my sources missing properties. So we're on 0.24.43, the same as the stream labs desktop client.

This overhauls the Recorder class. 
- Reverts to the older API that is in use by the Streamlabs desktop client in an effort to be able to copy them more.
- Removes the Window selection; we now auto detect this.
- The interface to the user is unchanged other than this.
- Cleans up the Recorder code by moving lots of stuff into the constructor.
- Changing OBS signal handling to be more forgiving.

To do list...
- [ ] Maybe worth switching away from monitor index to monitor ID, people had annoyances in the past with windows switching monitors.
- [ ] Can we do better with the audio encoder? There is an open issue.
- [x] Testing of all sorts of audio devices; no testing done that we're actually recording stuff yet. 
- [x] Run all the recording python tests. 
- [x] Test random combos of stuff in the UI.


aza547:
Think this is good to go. Not done all the stuff in the todo list but some are WIBNIs.

Issue #457, opened 2023-11-30
Media keys will start video when app is minimized


vaisest:
**Describe the bug**

**To Reproduce**
1. Open app
2. Start playback
3. Close app (to tray)
4. Press media keys

**Expected behavior**
I would expect for nothing to happen when the app is minimized. I would actually even prefer for media keys to not do anything when the app is open.

aza547:
Fixed by https://github.com/aza547/wow-recorder/pull/528

PR #528, opened 2024-09-16
Prevent media keys from working while minimized or not focused

vaisest:
Fixes my old issue #457.

This doesn't prevent media keys from working while the window is focused like originally. It should also be noted that the original quirks are still there. For example the keys only work after playback is started once, because they still come from the media session, and aren't actual hotkeys like the letter `k`.

I wasn't able to stop the stop key from working. I don't really understand why, but the "stop" handler just doesn't do anything.

### DIFF
--- a/src/renderer/VideoPlayer.tsx
+++ b/src/renderer/VideoPlayer.tsx
@@ -287,6 +287,25 @@ export const VideoPlayer = (props: IProps) => {
     }
   };
 
+  // By default the window hijacks media keys even when
+  // the window isn't focused or it is minimized
+  // so we override the action handlers
+  useEffect(() => {
+    ipc.on('window-focus-status', (arg: unknown) => {
+      const focused = arg as boolean;
+      if (focused) {
+        // unset action handlers when focused, making them work like initially
+        navigator.mediaSession.setActionHandler('play', null);
+        navigator.mediaSession.setActionHandler('pause', null);
+      } else {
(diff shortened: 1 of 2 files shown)


src/renderer/VideoPlayer.tsx
```diff
@@ -287,6 +287,25 @@ export const VideoPlayer = (props: IProps) => {
     }
   };
 
+  // By default the window hijacks media keys even when
+  // the window isn't focused or it is minimized
+  // so we override the action handlers
+  useEffect(() => {
+    ipc.on('window-focus-status', (arg: unknown) => {
```
aza547, line 294: I can't quite remember if this is required or not, but I think you might need to return a function from this useEffect hook to remove the listener, like we do here: https://github.com/aza547/wow-recorder/blob/main/src/renderer/containers/ApplicationStatusCard/Status.tsx#L214? 

Might be this is fine without it, but have definitely seen warnings in the console in dev mode when I've forgot this sort of thing before. Happy enough if you think it's not required but worth checking. IIRC the warning comes after 10+ listeners are added for the same event, so I think if you just flick between a bunch of videos 10+ times you should either see it or not.

aza547 (commented):
Thanks for the fix, one concern but otherwise happy.

vaisest:
Added. Sorry for the long delay (I might have almost forgotten this)

aza547:
Thanks!

PR #546, opened 2024-10-18
Update constants.ts - sob timer TWW S1

laurencee:
Siege of Boralus timer fix for TWW S1


laurencee:
Didn't notice this one was also off in the last patch.

aza547 (approved):

aza547:
thanks again :)

PR #551, opened 2024-11-12
Enable nvenc hevc support and electron version bump

kacperkowalczyk:
This PR bumps electron version to v33.2.0 based on electron-react-boilerplate project since v24.4.0 was not supported for some time already.

Reenables hardware acceleration and adds nvidia hevc encoder to the allow list.

For the rendering issues that some users might experience with nvidia hardware after enabling hardware acceleration - maybe it would be better to include a link to a guide explaining how to fix it rather than disabling that feature outright, alternatively, this could go behind a setting in the app, but I don't know enough about JS to do a good job with it.

### DIFF
--- a/.erb/configs/webpack.config.main.dev.ts
+++ b/.erb/configs/webpack.config.main.dev.ts
@@ -1,0 +1,63 @@
+/**
+ * Webpack config for development electron main process
+ */
+
+import path from 'path';
+import webpack from 'webpack';
+import { BundleAnalyzerPlugin } from 'webpack-bundle-analyzer';
+import { merge } from 'webpack-merge';
+import checkNodeEnv from '../scripts/check-node-env';
+import baseConfig from './webpack.config.base';
+import webpackPaths from './webpack.paths';
+
+// When an ESLint server is running, we can't set the NODE_ENV so we'll check if it's
+// at the dev webpack config is not accidentally run in a production environment
+if (process.env.NODE_ENV === 'production') {
+  checkNodeEnv('development');
+}
+
+const configuration: webpack.Configuration = {
+  devtool: 'inline-source-map',
+
+  mode: 'development',
+
+  target: 'electron-main',
+
+  entry: {
+    main: path.join(webpackPaths.srcMainPath, 'main.ts'),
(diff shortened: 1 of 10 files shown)


aza547:
I think you're probably right about the rendering issue. Not too fussed if it doesn't look right so long as there is a fix.

As for the electron bump, had a quick glance and looked good. Would be nice to get onto a more current version so long as it doesn't cause problems.

Major concern here is h265 is a licensed codec, with royalties. I don't think WCR is exempt from that? Although I wonder how OBS handle this. A guess is that it might be free to use offline, but then I need to worry about cloud users.

aza547:
ChatGPT seemed to give a decent summary:

>Summary
OBS can legally include H.265 recording support without paying licensing fees because:

>- It relies on external libraries (e.g., FFmpeg) for H.265 encoding.
>- It’s open-source and non-commercial, which places it outside the focus of patent enforcement.
>- Licensing responsibility shifts to end users if they use H.265 commercially.

>However, if a business incorporates OBS with H.265 recording for a commercial video-sharing service, that business would need to obtain the necessary H.265 licenses.

So I think this is still a problem for Pro / cloud users.

aza547:
A possible solution to that is to just disable this for Pro users. That sucks a bit to be losing a feature on paying money, but probably OK. Also need a mechanism to prevent cloud upload of previously recorded cloud videos.

A shame as I think AV1 is available in the version of OBS we ship and doesn't have all these complications. I don't have a GPU that supports it though.

aza547:
No reply here but I think there are lots of useful bits in here for the electron bump that I would like to take even if the hevc stuff is left out. That looks like a cherry-pick of https://github.com/electron-react-boilerplate/electron-react-boilerplate/commit/e7a8d90a61d7ae0871852f26a501f806b7aa5fea.

aza547:
I took the electron version bump here:
https://github.com/aza547/wow-recorder/commit/10f33e2253cd2c5ebc13d6ebf2bcfb5ffbd01ba4

A bit nervous about the compatibility with OSN but not seen anything actually not work yet so figure it's probably fine. 

No reply from the author of this PR re the hevc stuff and I don't think there is an easy win here, so I'm closing.

PR #554, opened 2024-11-21
Fix CMAA 2 game capture

CxLynx:
Disables `allow_transparency` on the Game Capture source. I'm not aware of any uses for this feature when capturing gameplay and it really borks capture when CMAA 2 is enabled in WoW.
If it is needed for something then it should probably be exposed to the user somewhere (but that's outside the scope of this 1-line hack).


aza547:
Thanks for the fix, sounds great.

PR #555, opened 2024-11-23
Redesign Video List as Table

aza547:
Still need to...
- [x] Tidy up all the table components (I did the raid one already)
- [x] Add indicators for sorting directions
- [x] Replace greyed out disk / cloud buttons with corresponding upload/download buttons.
- [x] Consider making row expansion and selection the same concept. 
- [x] Test deleting / undefined behaviour.
- [x] Fix persistent progress carried between povs
- [x] Make viewpoints field show first player? and "+N more"?
- [x] Resolve inline comments on this PR.
- [x] Make double clicking on a row expand/collapse it.
- [x] Expanded state needs incremented on a new video showing up?
- [x] Switch over to using unique ID to hold playing video instead of the mess currently.


src/renderer/components/Tables/VideoSelectionTable.tsx
```diff
@@ -0,0 +1,569 @@
...
+        id: 'Details',
+        size: 50,
+        cell: populateDetailsCell,
+      },
+    ],
+    []
+  );
+
+  const { category, playingVideo } = appState;
+  let columns = raidColumns;
```
aza547, line 342: switch for this

src/renderer/components/Tables/VideoSelectionTable.tsx
```diff
@@ -0,0 +1,569 @@
...
+    stateManager,
+    persistentProgress,
+  } = props;
+
+  const [expanded, setExpanded] = useState<ExpandedState>({});
+
+  /**
+   * Reset expanded on changing category.
+   */
+  useEffect(() => {
```
aza547, line 90: seems like this would be better set in the select category handler, higher in the stack

aza547:
Probably done here and this is fine to go into main. 
@StephixOne flagging in-case if you fancy a look, welcome any comments. Np if not.

aza547:
Ah I forgot I made some inline comments. Will resolve those tomorrow I guess.

StephixOne:
Happy to have a better look at the weekend, busy week.

aza547:
> Happy to have a better look at the weekend, busy week.

Thanks! Might be an idea to just try it out also - builds on this branch should be good.

aza547:
Some weird overflow maybe on the expand button?

{88BA2C03-2A5E-41A0-9812-668A991DC080}

src/renderer/components/Tables/VideoSelectionTable.tsx
aza547: I left a comment. Not changing for now.

aza547:
Think I'm actually done now.

image

Not sure I'm totally happy with the contents of the expanded box but don't want to let perfect be the enemy of good. I think this is a big improvement.

aza547:
Going for the merge. Let me know if you get a chance to look and have any retrospective comments @StephixOne  :)

PR #562, opened 2024-12-07
M+ Timer Fix

007bbb:


aza547:
Thanks for the PR. I'm not sure this is correct however. 

See https://www.wowhead.com/guide/mythic-plus-dungeons/the-war-within-season-1/overview.

It looks like you have added 90s to all these timers, but there is already handling for Challenger's Peril:
https://github.com/aza547/wow-recorder/blob/main/src/activitys/ChallengeModeDungeon.ts#L85

aza547:
> Shouldn't the progress time increase by 15 seconds every time a player dies during the progress, but doesn't the code say 10 seconds?
I think what happens here is that when the "CHALLENGE_MODE_END" event fires, the game logs a "challenge mode time" which becomes the CMDuration variable in the code. This time includes a correction for deaths, which is +5s. 

However when Challenger's Peril is active, this needs to be 15s per death, which isn't included in the event in the log. Hence the correction of +10s (5 + 10 = 15). 

I'm not 100% sure this is correct though and probably needs some investigation to confirm the combat log behaviour.

aza547:
See https://github.com/aza547/wow-recorder/commit/f0e8dd186cad3af0c52be60d06212ea5f18cb7f8

PR #564, opened 2024-12-10
Added in difficulty IDs for Ruins of AQ SoD P6

speeeday:
Added the new difficulty ID (185) for Ruins of AQ

Temple of AQ / Nightmare reuse old difficulty IDs and work correctly.


aza547:
Many thanks!

Issue #332, opened 2023-03-04
Disable hardware acceleration


aza547:
People had problems with this before. I think I've also seen screenshots of the app looking a bit rubbish.

I think we can just disable it natively in the app

https://stackoverflow.com/questions/62678626/enabling-hardware-acceleration-in-electron-app

Also see https://www.reddit.com/r/techsupport/comments/1168fb3/some_programs_area_not_working_correctly_and_are/

aza547:
![image](https://user-images.githubusercontent.com/32389325/222924099-2080af5f-eef3-47ba-9156-c1e03a9e3f53.png)

aza547:
https://github.com/aza547/wow-recorder/commit/3f9037eb1027919d5bb977e05c46401c02575fe6

aza547:
I've had to undo this last commit to make h265 work.

PR #607, opened 2025-03-11
Enable AV1 nvidia hardware encoding support

kacperkowalczyk:
This PR enables AV1 encoding support using nvidia hardware encoder in series 4000 and later GPUs with custom CRF values (based on personal testing mostly).
On supported cards, benefits include:
- much smaller file size
- up to 8k60 resolution support (including the dreaded 5120x1440)
This also enables hardware rendering in the app. In my testing it works without it, but with high-resolution playback, it's using up to 35% of my CPU (ryzen 7 9800x3d), so I would strongly recommend it.

### DIFF
--- a/release/app/package-lock.json
+++ b/release/app/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "WarcraftRecorder",
-  "version": "6.8.1",
+  "version": "6.8.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "WarcraftRecorder",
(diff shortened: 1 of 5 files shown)


aza547:
This looks great. Only concern is the re-enabling of HW acceleration for all users. I'm not sure if we spoke about this before or if it was someone else, but I think that was initially turned off because some users had issues with it (https://github.com/aza547/wow-recorder/issues/332). Not convinced that's reason to change course here, and I think it was probably just a misconfiguration of something anyway.

aza547 (approved):

PR #636, opened 2025-04-19
DrawingOverlay local

007bbb:
I was curious about excalidraw so I tested it. I added it by referring to the documentation on localization and it seems to be normal.
Excalidraw [Utils](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/api/props/) [Props](https://docs.excalidraw.com/docs/@excalidraw/excalidraw/api/utils)


aza547:
Awesome - thank you!

aza547 (approved):

Issue #602, opened 2025-02-23
Resizing screen when player is maximized is a bit rubbish


aza547:
https://cdn.discordapp.com/attachments/1164125914590879844/1343244922874101771/WarcraftRecorder_zEbxciNAy4.gif?ex=67bc91d1&is=67bb4051&hm=b9c99581b1df09d2cacc7647342b9b98246d8ec89a45a56a877e6f8b9221d8da&~

as shown in clip, the controls go out of scope

PR #644, opened 2025-04-25
feat: add button to reset video to its original height

FallenDoll:
This PR introduces a new feature that allows users to reset the video player's height to its original value with a single click. The key components of this update include:

UI Integration: A new button is added to the video player's control panel. This button leverages a Material-Icons design to indicate a "reset" action visually.

Functionality: Clicking the button calls a handler that retrieves the video player's original height (or uses a safe fallback) and adjusts the display accordingly. This ensures users can quickly revert any height adjustments without manually resizing the player.

Localization Updates: The feature is fully localized by updating the translation files in English, Chinese, Korean, and German. This ensures that the button's tooltip or accessible label (e.g., "Reset video height" in English) is displayed appropriately based on the user's locale.

Code Modifications: Changes have been made to both CategoryPage.tsx and VideoPlayer.tsx to incorporate the new reset logic and integrate the UI element seamlessly into the current video controls.

This enhancement improves the overall user experience by providing a quick and intuitive way to restore the video to its intended dimensions. Feedback and suggestions are very welcome.

### DIFF
--- a/src/localisation/chineseSimplified.ts
+++ b/src/localisation/chineseSimplified.ts
@@ -357,6 +357,7 @@ const CHINESE_SIMPLIFIED: Translations = {
   [Phrase.Saving]: '保存',
   [Phrase.StartTyping]: '开始打字',
   [Phrase.ToggleDrawingMode]: '切换绘图模式',
+  [Phrase.ResetVideoHeight]: '重置视频高度',
   [Phrase.StarSelected]: '锁定选定的行',
   [Phrase.UnstarSelected]: '解锁选定的行',
   [Phrase.Selection]: '选择',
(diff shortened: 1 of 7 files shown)


aza547:
Hello, and thanks for the PR! I thought this was to address https://github.com/aza547/wow-recorder/issues/602 initially but actually I don't think it does, as in the event of that issue the button is off-screen and not clickable. 

I had a review of the code and I don't think the approach is quite right. Nor is the behaviour what I want. I made some comments on the usage of state in-line in the code. Happy to discuss that.

I think this does go some way to provide the behaviour I want here which is on resizing the app the video and the grip handle for dragging is never off-screen.  I hope you won't be discouraged from further contributions!

PR #669, opened 2025-05-22
Add vc_redist install checks

imarkvisser:
## Summary
This PR adds a check to determine if the Visual C++ redistributable is already installed before attempting to download and install it. This improves the setup experience and reduces unnecessary downloads/installs.

## Changes
- Modified `installer.nsh` to check Windows registry for existing VC++ Redistributable installation
- Only downloads and installs VC++ Redistributable if not already present on the system
- Added cleanup to remove the downloaded installer file after installation

## Technical Details
The installer now checks three registry locations for the Visual C++ Redistributable:
1. `HKLM\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64`
2. `HKLM\SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64` (Installed DWORD)
3. `HKLM\SOFTWARE\WOW6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\x64` (for 32-bit NSIS on 64-bit systems)

The check looks for version 14 or higher, which covers Visual Studio 2015-2022 redistributables.

### DIFF
--- a/installer.nsh
+++ b/installer.nsh
@@ -1,21 +1,68 @@
 !macro customInstall
-  NSISdl::download https://aka.ms/vs/17/release/vc_redist.x64.exe "$INSTDIR\vc_redist.x64.exe"  
+  # Check if Visual C++ Redistributable is already installed
+  # Check for the 64-bit version in the registry
   
-  ${If} ${FileExists} `$INSTDIR\vc_redist.x64.exe`
-    ExecWait '$INSTDIR\vc_redist.x64.exe /passive /norestart' $1
-
-    ${If} $1 != '0' 
-      ${If} $1 != '3010'
-        MessageBox MB_OK|MB_ICONEXCLAMATION 'WARNING: Warcraft Recorder was unable to install the latest Visual C++ Redistributable package from Microsoft.'
-      ${EndIf}
+  Var /GLOBAL VCRedistInstalled
+  StrCpy $VCRedistInstalled "0"
+  
+  # Check for Visual Studio 2015-2022 Redistributable (they use the same key)
+  # This checks for the minimum version required
+  ReadRegStr $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Version"
+  ${If} $0 != ""
+    # Version format is like "v14.29.30133.00"
+    # Extract major version number for comparison
+    StrCpy $1 $0 2 1  # Get 2 chars starting from position 1 (skips 'v')
+    
+    # Check if version is 14 or higher (VS 2015+)
+    IntCmp $1 14 0 0 version_ok
+    version_ok:
+      StrCpy $VCRedistInstalled "1"
+  ${EndIf}
+  
+  # Also check the registry location used by VS 2022
+  ${If} $VCRedistInstalled == "0"
+    ReadRegDWORD $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Installed"
+    ${If} $0 == "1"
+      StrCpy $VCRedistInstalled "1"
     ${EndIf}
+  ${EndIf}
+  
+  # Alternative check using WOW6432Node for 32-bit NSIS on 64-bit systems
+  ${If} $VCRedistInstalled == "0"
+    ReadRegDWORD $0 HKLM "SOFTWARE\WOW6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Installed"
+    ${If} $0 == "1"
+      StrCpy $VCRedistInstalled "1"
+    ${EndIf}
+  ${EndIf}
+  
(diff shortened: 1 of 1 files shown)


installer.nsh
```diff
@@ -1,21 +1,68 @@
...
+  Var /GLOBAL VCRedistInstalled
+  StrCpy $VCRedistInstalled "0"
+  
+  # Check for Visual Studio 2015-2022 Redistributable (they use the same key)
+  # This checks for the minimum version required
+  ReadRegStr $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Version"
+  ${If} $0 != ""
+    # Version format is like "v14.29.30133.00"
+    # Extract major version number for comparison
+    StrCpy $1 $0 2 1  # Get 2 chars starting from position 1 (skips 'v')
```
aza547, line 14: This seems rather fragile - although it's probably fine. Not a big fan of the 2 char assumption here. It's also just ignoring any non-major version changes. Although I guess that's probably fine.

installer.nsh
```diff
@@ -1,21 +1,68 @@
...
+    
+    # Check if version is 14 or higher (VS 2015+)
+    IntCmp $1 14 0 0 version_ok
+    version_ok:
+      StrCpy $VCRedistInstalled "1"
+  ${EndIf}
+  
+  # Also check the registry location used by VS 2022
```
aza547, line 22: This comment seems misleading - it's the same location as above but now you are checking the installed key.

installer.nsh
```diff
@@ -1,21 +1,68 @@
...
+  # Check for Visual Studio 2015-2022 Redistributable (they use the same key)
+  # This checks for the minimum version required
+  ReadRegStr $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Version"
+  ${If} $0 != ""
+    # Version format is like "v14.29.30133.00"
+    # Extract major version number for comparison
+    StrCpy $1 $0 2 1  # Get 2 chars starting from position 1 (skips 'v')
+    
+    # Check if version is 14 or higher (VS 2015+)
+    IntCmp $1 14 0 0 version_ok
```
aza547, line 17: /14.0/ is already in the path so what's the value in doing this check at all?

installer.nsh
```diff
@@ -1,21 +1,68 @@
...
+  
+  # Also check the registry location used by VS 2022
+  ${If} $VCRedistInstalled == "0"
+    ReadRegDWORD $0 HKLM "SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Installed"
+    ${If} $0 == "1"
+      StrCpy $VCRedistInstalled "1"
+    ${EndIf}
+  ${EndIf}
+  
+  # Alternative check using WOW6432Node for 32-bit NSIS on 64-bit systems
```
aza547, line 30: Is this relevant? WCR has always been 64 bit only. I suspect this is redundant.

installer.nsh
```diff
@@ -1,21 +1,68 @@
...
+      #     MessageBox MB_OK|MB_ICONEXCLAMATION 'You must restart your computer to complete the installation.'
+      # ${EndIf}
 
-  ${Else}
+    ${Else}
       MessageBox MB_OK|MB_ICONEXCLAMATION 'WARNING: Warcraft Recorder was unable to download the latest Visual C++ Redistributable package from Microsoft.'
+    ${EndIf}
+    
+  ${Else}
+    DetailPrint "Visual C++ Redistributable is already installed. Skipping..."
```
aza547, line 65: Curious how this shows up. I'll have a test.

installer.nsh
```diff
@@ -1,21 +1,68 @@
...
+  # Check for the 64-bit version in the registry
   
-  ${If} ${FileExists} `$INSTDIR\vc_redist.x64.exe`
-    ExecWait '$INSTDIR\vc_redist.x64.exe /passive /norestart' $1
-
-    ${If} $1 != '0' 
-      ${If} $1 != '3010'
-        MessageBox MB_OK|MB_ICONEXCLAMATION 'WARNING: Warcraft Recorder was unable to install the latest Visual C++ Redistributable package from Microsoft.'
-      ${EndIf}
+  Var /GLOBAL VCRedistInstalled
```
aza547, line 5: I doubt this needs to be global

aza547 (changes requested):
I think this, like the other PR you submitted, highlights something that can be improved, but I think once again it looks like this was AI generated and not quite right. I'm not an expert on NSIS or anything so once again happy to be corrected if I'm missing the point.

aza547:
I also think this has a cleaner fix of this: https://github.com/aza547/wow-recorder/commit/6dea2d2f2dcfe2f776b0e73cb8023850ec35778e

PR #676, opened 2025-05-28
Expose and use obs_output_force_stop

aza547:
[Forks OSN](https://github.com/aza547/obs-studio-node-wcr) and adds force stop function. 
- That gives is a route to recovery without restarting OBS.
- Lets us gracefully sleep without restarting OBS.


aza547:
https://github.com/streamlabs/obs-studio-node/compare/staging...aza547:obs-studio-node-wcr:staging

Issue #668, opened 2025-05-18
Zoom in hotkey broken


aza547:
![Image](https://github.com/user-attachments/assets/abafeda9-edb9-4614-ba0b-c6198d93f1ad)

lionskape:
@aza547 can you take a look on the dependency list, may be you know a proper candidate to start investigation?

lionskape:
I found a source issue in Electron: https://github.com/electron/electron/issues/29996

PR #691, opened 2025-07-01
Refactor menu builder, remove unused code.

lionskape:
Fixes https://github.com/aza547/wow-recorder/issues/668

This PR refactors the menu builder by removing legacy macOS and development-only menu code and streamlining the default menu template. It also adds explicit zoom controls for in-/out/ reset zoom operations. So, menu is not visible, only hotkey access is possible (same as a11y tools).

Removed old Darwin and development context-menu logic from MenuBuilder
Simplified MenuBuilder to always use buildDefaultTemplate
Introduced hidden zoom menu items for zoom in, zoom out, and reset zoom


aza547:
Thanks! Had a play and looks good, kind of forgot this menu code existed. Something I picked up from the boilerplate I gues.

PR #692, opened 2025-07-07
Adds Configurable Push-To-Talk Release Delay

### LINKED ISSUE #661: PTT delay on release to match discord
Image

NathanPrice:
Closes #661

### What
- Introduce's `pushToTalkReleaseDelay` (0–2000 ms) setting in the config.
- Add slider UI in Settings with live seconds/ms preview.
- Added a localized tooltip explaining what it does.
- In main process, schedule mute on key-up/mouse-up using the configured delay.

### Why
(see issue #661). This adds a small post-release buffer, matching Discord’s behavior.

### How to test
1. Enable Push-To-Talk in Settings.  
2. Adjust “Release Delay” slider and verify the label shows. Example: `1.25 s` or `250 ms`.  
3. Press & hold the PTT key: mic should unmute immediately.  
4. Release the key: mic should stay unmuted for the configured delay, then mute.  
5. Switch app language and confirm the tooltip text is localized.

---

> **Note:**  
> - I’m not 100% confident in some translation's I provided since I don't speak anything besides English. If those need tweaks, please adjust or advise.  
> - This is my first contribution here and it’s been a while since I used JavaScript/TypeScript. If anything looks off or I need to change anything, let me know and I’ll fix it.  
> - Verified behavior and made sure I didn't break anything.


src/renderer/AudioSourceControls.tsx
```diff
@@ -644,6 +650,49 @@ const AudioSourceControls = (props: IProps) => {
     </div>
   );
 
+  useEffect(() => {
```
aza547, line 653: Not sure I understand the need for this

src/renderer/AudioSourceControls.tsx
```diff
@@ -644,6 +650,49 @@ const AudioSourceControls = (props: IProps) => {
...
+  const commitReleaseDelay = (newValue: number[]) => {
+    const ms = newValue[0];
+    if (typeof ms !== 'number') return;
+    setConfig((prev) => ({ ...prev, pushToTalkReleaseDelay: ms }));
+  };
+
+  const getPushToTalkReleaseDelaySlider = () => (
+    <div className="flex flex-col w-[300px]">
+      <Label className="flex items-center">
+        Release Delay
```
aza547: This should use getLocalePhrase to handle translations

src/renderer/AudioSourceControls.tsx
```diff
@@ -644,6 +650,49 @@ const AudioSourceControls = (props: IProps) => {
...
+        <Slider
+          defaultValue={[localReleaseDelay]}
+          onValueChange={(vals) => setLocalReleaseDelay(vals[0])}
+          onValueCommit={commitReleaseDelay}
+          min={0}
+          max={2000}
+          step={1}
+          withTooltip={false}
+        />
+        <span className="whitespace-nowrap">
```
aza547: I was going to say it would fit better with the style of the rest of the app to not have the text to the right here but to have a slider label. But it seens the shadcn slider component doesn't support this. The MUI one (used elsewhere does). But I think it's fine as is given that. Although maybe there is a better tailwind text class to use as it looks a bit out of place.

Perhaps `text-sm text-foreground-lighter` ?

aza547 (changes requested):
Looks great! Some minor comments.

Could you also update the CHANGELOG.md file?

src/config/configSchema.ts
```diff
@@ -385,6 +386,13 @@ export const configSchema = {
     type: 'string',
     default: '',
   },
+  pushToTalkReleaseDelay: {
+    description: Phrase.PushToTalkReleaseDelayDescription,
+    type: 'integer',
+    default: 200,
```
aza547: Oh, I think the default should remain zero just so people don't notice anything on upgrade.

src/main/Recorder.ts
```diff
@@ -856,10 +861,51 @@ export default class Recorder extends EventEmitter {
...
-      uIOhook.on('keydown', (e) => pttHandler(() => this.unmuteInputDevices(), e));
-      uIOhook.on('keyup', (e) => pttHandler(() => this.muteInputDevices(), e));
-      uIOhook.on('mousedown', (e) => pttHandler(() => this.unmuteInputDevices(), e));
-      uIOhook.on('mouseup', (e) => pttHandler(() => this.muteInputDevices(), e));
+      const getPttDelay = () => this.cfg.get<number>('pushToTalkReleaseDelay') ?? 0;
+
+      uIOhook.on('keydown', (e) =>
+        pttHandler(() => {
+          if (this.releaseDelayTimer) {
+            clearTimeout(this.releaseDelayTimer);
```
aza547: I think you should wrap this timeout handling into the pttHandler function to avoid repeating it x4

src/renderer/AudioSourceControls.tsx
NathanPrice: Not sure what you mean by this. Could you elaborate? All of the other audio settings are in this file from what I saw and thought this needed to be added here as well.

src/renderer/AudioSourceControls.tsx
NathanPrice: Ya you're right. Not sure how I missed that. I went ahead and fixed this.

src/config/configSchema.ts
NathanPrice: Ya didn't think about that. Went ahead and changed the default to 0.

src/renderer/AudioSourceControls.tsx
NathanPrice: Wasn't aware of that class and you're right it does look out of place. I went ahead and added that to the slider and label and it looks a lot better. Couldn't figure out how to get the styling on the slider to match the volume sliders exactly but not sure if that matters much.  I also noticed the label I made for this (Release Delay) wasn't lining up with the other labels here because I had it styled differently which I fixed.

Something else I noticed, the push to talk keybind label text in English says the same thing as the label over the Push to Talk slider. Think that should be updated to say "Push to Talk Keybind" or something similar. Was this intentional or no? Not sure if this is the same way for the other languages.

NathanPrice:
Ok went ahead and fixed everything besides 2 things that you commented on. My machine was acting up so there might be some weird / duplicate commits (long day lol). I just need a few responses on the things I mentioned in your comments and to update the pttdelay thing you mentioned.

src/renderer/AudioSourceControls.tsx
aza547: That's probably an oversight - agree it should say "push to talk keybind". Feel free to fix if you fancy, but not the end of the world :)

src/renderer/AudioSourceControls.tsx
aza547: This useEffect is saying to update the `localReleaseDelay` stateful variable whenever the `config.pushToTalkReleaseDelay` field changes. But I don't think the latter can change without the former changing first given you're passing 


```
          value={[localReleaseDelay]}
          onValueChange={(vals) => setLocalReleaseDelay(vals[0])}
          onValueCommit={commitReleaseDelay}
```

Oh but actually I see it now. Your commitReleaseDelay is a slightly different pattern from the rest of the config handling here. The sliders I wrote are just setting the `config` variable directly which is the triggering a debounce timer to set the config in the backend after a small delay.

That means for my sliders, the config may be applied while they are being dragged. But for your slider it won't be applied until you commit. 

I think that's actually fine and probably better in this case. There are some cases elsewhere (such as dragging the overlay position slider) where it's desirable for the config to be applied while dragging, which is probably the source of this pattern in the first place. 

Anyway I'd probably have stuck with the same pattern myself, but as above, I think this is fine and you can ignore it.

aza547:
Think I have responded to everything and left open the comments that still need addressed. Thanks :)

src/main/Recorder.ts
NathanPrice: Went ahead and fixed this. I put the general info in the commit message but this is what I did for this. If this doesn't suite your taste let me know and I can go about this another way / put this back similar to how it was before.

I eliminated the duplicate code for this. Also moved all timeout logic into a single pushToTalkHandler method. This also fixed some code duplication since keydown/mousedown and keyup/mouseup had identical logic. 

I also made some Additional Improvements to this code I had here. 

State checking: Only unmute if muted, only mute if unmuted.
Better timer handling: Skip timer for 0 delay, handle negative values. Figured it' be good to have that in case something weird ever happens.
Proper cleanup: Made code remove specific listeners instead of all listeners.
Added uIOhook.start(): Ensures the hook is actually running instead of failing silently if it breaks somehow.

NathanPrice:
I went ahead and fixed the last thing I had still opened for this whenever you get a chance to look and made sure there wasn't anything else I still needed to address. Let me know if I still need to change or fix anything and I'll do it. I also fixed the label I mentioned for the push to talk keybind (it was using the same phrase as the push to talk slider).

src/main/Recorder.ts
```diff
@@ -849,64 +849,33 @@ export default class Recorder extends EventEmitter {
...
-        if (isPushToTalkHotkey(config, convertedEvent)) {
-          fn();
-        }
-      };
+      const events = ['keydown', 'keyup', 'mousedown', 'mouseup'];
 
-      /* eslint-disable prettier/prettier */
-      const getPttDelay = () => this.cfg.get<number>('pushToTalkReleaseDelay') ?? 0;
+      events.forEach((event) => {
+        uIOhook.removeAllListeners(event);
```
aza547: I don't think this is required as on line 767 we already do this
`uIOhook.removeAllListeners();`

src/main/Recorder.ts
```diff
@@ -849,64 +849,33 @@ export default class Recorder extends EventEmitter {
...
-      uIOhook.on('keydown', (e) =>
-        pttHandler(() => {
-          if (this.releaseDelayTimer) {
-            clearTimeout(this.releaseDelayTimer);
-            this.releaseDelayTimer = undefined;
-          }
-          this.unmuteInputDevices();
-        }, e)
+      uIOhook.on('keydown', (e: UiohookKeyboardEvent) =>
+        this.pushToTalkHandler(true, e, config),
```
aza547: The boolean arg you have added here is contained within the event already, you can just check that

```
export interface UiohookKeyboardEvent {
    type: EventType.EVENT_KEY_PRESSED | EventType.EVENT_KEY_RELEASED;
```

src/main/Recorder.ts
```diff
@@ -849,64 +849,33 @@ export default class Recorder extends EventEmitter {
...
-            this.releaseDelayTimer = undefined;
-          }, getPttDelay());
-        }, e)
+
+      uIOhook.on('mouseup', (e: UiohookMouseEvent) =>
+        this.pushToTalkHandler(false, e, config),
       );
-      /* eslint-enable prettier/prettier */
+
+      try {
```
aza547: Not sure I understand the need for this. 

`uIOhook.start();` is already called in main.ts

src/main/Recorder.ts
```diff
@@ -1710,6 +1725,42 @@ export default class Recorder extends EventEmitter {
...
+    const converted = convertUioHookEvent(event);
+    if (!isPushToTalkHotkey(audioConfig, converted)) return;
+
+    if (isPress) {
+      if (this.releaseDelayTimer) {
+        clearTimeout(this.releaseDelayTimer);
+        this.releaseDelayTimer = undefined;
+      }
+
+      if (this.inputDevicesMuted) {
```
aza547: `this.inputDevicesMuted` is checked in the unmute function so you don't need to check it here

src/main/Recorder.ts
```diff
@@ -1710,6 +1725,42 @@ export default class Recorder extends EventEmitter {
...
+        this.unmuteInputDevices();
+      }
+    } else {
```
aza547: suggest an early return to avoid a layer of indentation

src/main/Recorder.ts
```diff
@@ -1710,6 +1725,42 @@ export default class Recorder extends EventEmitter {
...
+      if (!this.inputDevicesMuted) {
+        const delay = Math.max(
+          0,
+          this.cfg.get<number>('pushToTalkReleaseDelay') ?? 0,
+        );
+
+        if (delay === 0) {
```
aza547: you can just shove this all in setTimeout without worrying about checking for zero. that will mean this gets executed on the next pass of the event loop instead of inline here, but i think that's fine and simplifies this logic a bit

src/main/Recorder.ts
```diff
@@ -1710,6 +1725,42 @@ export default class Recorder extends EventEmitter {
...
+      if (this.releaseDelayTimer) {
+        clearTimeout(this.releaseDelayTimer);
+        this.releaseDelayTimer = undefined;
+      }
+
+      if (this.inputDevicesMuted) {
+        this.unmuteInputDevices();
+      }
+    } else {
+      if (!this.inputDevicesMuted) {
```
aza547: ditto re earlier comment, you don't need to check this here - fine to add the no-op call the muteInputDevices to simplify the logic

src/main/Recorder.ts
NathanPrice: Fixed. Removed the redundant event specific listener removal code I was doing since I didn't see what you pointed out when I made this.

src/main/Recorder.ts
NathanPrice: Aah wasn't aware. Went ahead and updated this to what you were saying.

src/main/Recorder.ts
NathanPrice: Wasn't aware my bad. Went ahead and removed this. Not sure what I was thinking lol.

src/main/Recorder.ts
NathanPrice: Went ahead and refactored this to return early. Also removed the redundant state checks.

src/main/Recorder.ts
NathanPrice: True. I mentioned this in the previous comment but went ahead and added an early return here and simplified this.

src/main/Recorder.ts
NathanPrice: Went ahead and removed this check so on the next event loop so it'll do what you say here and make the code a bit cleaner as well.

src/main/Recorder.ts
NathanPrice: Mentioned earlier but went ahead and removed this check here and made this simpler.

PR #694, opened 2025-07-07
Add a region from combat log to the saved metadata.

lionskape:
This PR extends combatant metadata by extracting a region from the combat log and persisting it through the parsing pipeline.

- Augments `ambiguate` to return a third element for region
- Updates `LogHandler` and `types` to include `_region`
- Extends `Combatant` class with a `region` property, adjusts `isFullyDefined` and serialization

### Reviewed Changes

| File                       | Description                                    |
| -------------------------- | ---------------------------------------------- |
| src/parsing/logutils.ts    | Extracts `region` from the split and returns it|
| src/parsing/LogHandler.ts  | Destructures and assigns `combatant.region`     |
| src/main/types.ts          | Adds `_region` field to `RawCombatant`         |
| src/main/Combatant.ts      | Adds `region` getter/setter, updates checks and serialization |

### DIFF
--- a/src/main/Combatant.ts
+++ b/src/main/Combatant.ts
@@ -14,6 +14,8 @@ export default class Combatant {
 
   private _realm?: string;
 
+  private _region?: string;
+
   /**
    * Constructs a new Combatant.
    *
@@ -107,13 +109,30 @@ export default class Combatant {
     this._realm = value;
   }
 
+  /**
+   * Gets the region.
+   */
+  get region() {
+    return this._region;
+  }
+
+  /**
+   * Sets the region.
+   */
+  set region(value) {
+    this._region = value;
+  }
+
   isFullyDefined() {
     const hasGUID = this.teamID !== undefined;
     const hasName = this.name !== undefined;
     const hasRealm = this.realm !== undefined;
+    const hasRegion = this.region !== undefined;
     const hasSpecID = this.specID !== undefined;
     const hasTeamID = this.teamID !== undefined;
-    return hasGUID && hasName && hasRealm && hasSpecID && hasTeamID;
+    return (
+      hasGUID && hasName && hasRealm && hasRegion && hasSpecID && hasTeamID
+    );
   }
 
   getRaw(): RawCombatant {
@@ -123,6 +142,7 @@ export default class Combatant {
     if (this.specID !== undefined) rawCombatant._specID = this.specID;
     if (this.name !== undefined) rawCombatant._name = this.name;
     if (this.realm !== undefined) rawCombatant._realm = this.realm;
+    if (this.region !== undefined) rawCombatant._region = this.region;
 
     return rawCombatant;
   }
--- a/src/parsing/logutils.ts
+++ b/src/parsing/logutils.ts
@@ -24,5 +24,6 @@ export const ambiguate = (nameRealm: string): string[] => {
   const split = nameRealm.split('-');
   const name = split[0];
   const realm = split[1];
-  return [name, realm];
+  const region = split[3];
+  return [name, realm, region];
 };
--- a/src/main/types.ts
+++ b/src/main/types.ts
@@ -242,6 +242,7 @@ type RawCombatant = {
   _specID?: number;
   _name?: string;
   _realm?: string;
+  _region?: string;
 };
 
 /**
--- a/src/parsing/LogHandler.ts
+++ b/src/parsing/LogHandler.ts
@@ -427,7 +427,8 @@ export default abstract class LogHandler extends EventEmitter {
       return combatant;
     }
 
-    [combatant.name, combatant.realm] = ambiguate(srcNameRealm);
+    [combatant.name, combatant.realm, combatant.region] =
+      ambiguate(srcNameRealm);
     this.activity.addCombatant(combatant);
     return combatant;
   }


src/parsing/logutils.ts
```diff
@@ -24,5 +24,6 @@ export const ambiguate = (nameRealm: string): string[] => {
   const split = nameRealm.split('-');
   const name = split[0];
   const realm = split[1];
-  return [name, realm];
+  const region = split[3];
```
aza547, line 27: I don't see how this works? As far as I understand it the combat log doesn't provide the region. 

For example:  `Alexsmite-TwistingNether`  will end up giving you undefined as the region here.

aza547:
Most of this looks totally sensible, but I don't think the combat log gives region information, and so I don't see how this can work.

I had some thoughts on how this could work by:
- Assuming the most likely region given the Windows locale.
- Allowing that to be overridden in the config.

aza547:
Actually, even better, we can read the `retail/WTF/Config.wtf` to find it:

```
SET portal "EU"
``` 

That actually seems like a nice way to do this - i.e. read that file when WoW launched and apply that to all recordings.

lionskape:
There is ia sample of SPELL_AURA_APPLIED from my recent log:

`7/6/2025 23:22:01.5223  SPELL_AURA_APPLIED,Player-1084-0AF0A8F9,"Betterpants-TarrenMill-EU",0x514,0x80000000,Player-1084-0AF0A8F9,"Betterpants-TarrenMill-EU",0x514,0x80000000,47753,"Божественное покровительство",0x2,BUFF,534734`

And from SPELL_CAST_SUCCESS:

`7/6/2025 23:22:02.6643  SPELL_CAST_SUCCESS,Player-1084-0AF0A8F9,"Betterpants-TarrenMill-EU",0x514,0x80000000,Vehicle-0-1461-2769-2439-231075-00006ADA7D,"Хромовый король Галливикс",0x10a48,0x80000000,32379,"Слово Тьмы: Смерть",0x20,Player-1084-0AF0A8F9,0000000000000000,5755380,10770200,13305,105400,21809,603,253,0,0,2122440,2500000,12500,-506.02,-651.98,2409,5.9662,684`

As I understand this is a two sources truth for the combatant data in the LogHandlers (events which is used in code to parse it)
I believe it will work.

P.S. Reading the configs also seems doable, but not like a best way. Info is already provided in the logs (which folder is explicitly defined by user). By the way I believe config will always be like "..\WTF\Config.wtf" relatively to logs folder.

src/parsing/logutils.ts
aza547: OK - I was wrong - I can see this works now.

But this is shared code and I expect classic logs will not have the region.

It would also I think be nice to parse this into a set of known enums rather than strings. 

I expect eventually it will be nice to map "EU" -> "eu" etc to facilitate linking to WCL / RIO / whatever other things. e.g. `https://www.warcraftlogs.com/character/eu/twisting-nether/alexsmite`

src/main/Combatant.ts
```diff
@@ -107,13 +109,30 @@ export default class Combatant {
...
   isFullyDefined() {
     const hasGUID = this.teamID !== undefined;
     const hasName = this.name !== undefined;
     const hasRealm = this.realm !== undefined;
+    const hasRegion = this.region !== undefined;
     const hasSpecID = this.specID !== undefined;
     const hasTeamID = this.teamID !== undefined;
-    return hasGUID && hasName && hasRealm && hasSpecID && hasTeamID;
+    return (
+      hasGUID && hasName && hasRealm && hasRegion && hasSpecID && hasTeamID
```
aza547: I think `isFullyDefined` should probably not check the region given it may be undefined. I think this fn is used to reduce log parsing overhead particularly in classic or game modes where there is no COMBATANT_INFO.

aza547: Definitely worth a comment here to explain that choice though

src/parsing/logutils.ts
lionskape: Thank you for review!

Enum is a definitly something that should be used for a such things.

What do you think if I'll left it here in an original format (with a string type in RawCombatant), and transform later when will be constructing a link? So, during transformation I will use a strict types and enums for a such data.
The idea is to left a raw parsed data from Log as close to original as possible and adjust a strict types later, when working with this data.

I will also need to remap "TwistingNether" to the "twisting-nether", "Ревущийфьорд" to the "howling-fjord" and etc.
I've collected a list of all existing realms with it regions. Tehre is a sample:

(please, note that it is only a sample draft, which will be adjusted before I will push it and make a PR. Just to introduce an idea)

```typescript
type Locale = 'en' | 'de' | 'fr' | 'ru' | 'tw' | 'cn';

type Region = 'EU' | 'KR' | 'TW' | 'US';

// Locales object type
type RealmLocales = {
  [K in Locale]: string;
};

type Realm = {
  name: string;
  slug: string;
  locales: RealmLocales;
  region: Region
};

type Realms = Realm[];

export type { Locale, Region, RealmLocales, Realm, Realms };

export const regionToRegionSlug = {
  'EU': 'eu',
  'KR': 'kr',
  'TW': 'tw',
  'US': 'us',
} as const;

export const realms = [
  {
    name: 'Howling Fjord',
    slug: 'howling-fjord',
    locales: {
      en: 'Howling Fjord',
      de: 'Howling Fjord',
      fr: 'Howling Fjord',
      ru: 'Ревущий фьорд',
      tw: 'Howling Fjord',
      cn: '嚎风峡湾',
    },
    region: 'EU',
  },
  {
    name: 'Scarshield Legion',
      slug: 'scarshield-legion',
      locales: {
        en: 'Scarshield Legion',
        de: 'Scarshield Legion',
        fr: 'Légion du Bouclier balafré',
        ru: 'Scarshield Legion',
        tw: '裂盾軍團',
        cn: '裂盾军团',
    },
    region: 'EU',
  },
  {
    name: 'Aggra (Português)',
    slug: 'aggra-português',
    locales: {
      en: 'Aggra (Português)',
      de: 'Aggra (Portugiesisch)',
      fr: 'Aggra (Português)',
      ru: 'Aggra (Português)',
      tw: '阿格拉(葡萄牙語)',
      cn: 'Aggra (Português)',
    },
    region: 'EU',
  },
  // etc...
]
```

src/main/Combatant.ts
lionskape: Thank you!
I've made a changes that you requested.

Should I also fix `const hasGUID = this.teamID !== undefined;` in this PR, or make a separate for fix?
I believe it is a bug in this function and should be `const hasGUID = this.GUID !== undefined;`

lionskape:
@aza547 I've made some checks ():

Classic (MOP):
`7/8/2025 23:46:12.1113  SPELL_AURA_APPLIED,Player-4452-05D21632,"Йурчи-Хроми-EU",0x80000518,0x80000000,Creature-0-5572-860-19-53714-00006D82B0,"Тренировочный манекен",0x80000a28,0x80000000,107079,"Сотрясающая ладонь",0x1,DEBUFF

`

Classic Era (SoD):
`7/8/2025 23:47:49.0733  SPELL_CAST_SUCCESS,Player-5829-02F67AB4,"Драхарикслок-Исповедь(сезонный)-EU",0x511,0x0,Creature-0-5250-0-254-299-00006D82A8,"Молодой волк",0x10a28,0x0,686,"Стрела Тьмы",0x20,Player-5829-02F67AB4,0000000000000000,100,100,0,0,0,0,0,-1,0,0,0,-8895.46,-86.43,1429,0.0652,0`

I believe, now region is included in all of the recorded logs.

P.S. Sorry, I just created a 1st lvl characters, and did not find any way to get some "SPELL_AURA_APPLIED", but I believe it will be the same charater name format.

src/parsing/logutils.ts
aza547: Agree handling this elsewhere is sensible and fine to leave as a string here. Fits with the raw type. 

I'm nervous about how the realms will be handled. Maintaining a list of all realms, slugs and translations seems unrealistic. But we should discuss that elsewhere.

aza547:
Thanks - that's convincing. Looks good. I'll merge.

Issue #698, opened 2025-07-13
Add a maintainable lists of realms to the repository


lionskape:
**Is your feature request related to a problem? Please describe.**
The project currently lacks a maintainable, up-to-date cache of WoW realms fetched from Blizzard APIs. This is needed for reliable app functionality, automation, and future extensibility.

**Describe the solution you'd like**
- Implement a script to fetch the list of realms from Blizzard APIs and store the raw response as JSON.
- Save fetched lists as files in `data/realms-{namespace}.json`, with namespace values for classic, retail, and classic-era.
- Allow the script to be run locally or via GitHub Actions for automated monthly updates.
- Use the stored raw API response as a cache source for later features.
- Fetch a fresh API token at each script startup (using client credentials via OAuth).

**Describe alternatives you've considered**
- Manually maintaining a static list of realms in the repository.
- Fetching realms live at runtime (less reliable, not cacheable).
- Storing only selected fields instead of raw API response.

**Additional context**
- Authentication is done via a script startup request for an OAuth token:
    curl -u {client_id}:{client_secret} -d grant_type=client_credentials https://oauth.battle.net/token
    curl --header "Authorization: Bearer <access_token>" <REST API URL>
- Raw API responses are preferred for future flexibility.
- Automated monthly updates are preferred. Future improvements may include filtering or transforming cached data.

PR #699, opened 2025-07-19
Add Blizzard API integration for fetching realms

lionskape:
- Implemented namespaces for Retail API to categorize data.
- Created game version enumeration for Classic, ClassicEra, and Retail.
- Defined supported locales and their mappings.
- Established namespace categories for Static, Dynamic, and Profile.
- Developed URI builder for Battle.net API with regional support.
- Added dotenv package for environment variable management.
- Created scripts for fetching realms from Blizzard API, including error handling and caching.
- Implemented Zod schemas for validating API responses.
- Added integration and unit tests for OAuth token fetching and realm search functionality.
- Created a script to fetch and save realm data for all supported regions and game versions.

Solves #698


data/Classic/cache/realms.json
lionskape, line 1: Looks like this files is really huge, but absolutely the must to create a WCL links, and all other links (we need a "slug" from this files).
What do you think about it? Can you suggest any other proper way to store them?

So we can't read this in runtime of recorder app. There is a few reasons:
1. We must have a blizzard api tokens
2. This data can be received only with both Clinet ID (which is public) and Secret (which is private)
3. Blizzard api does not look stable enough for runtime calls. During the development I see some 500 errors from api.

One more option to have an "in-app cache" (or some other cache), because it is really rarely updated.

.env.example
```diff
@@ -0,0 +1,2 @@
+BLIZZARD_CLIENT_ID= # Create a client and paste a ClientID. Instruction: https://develop.battle.net/documentation/guides/getting-started
```
lionskape, line 1: So, creation of api keys is not very stable. I try 3 days in a row, before creation page becomes working.

data/Classic/cache/realms.json
aza547: Sorry it has taken me so long to think about this. A few ideas:
- Is a 1MB file really that bad? Have you noticed real performance impacts?
- If it is bundled maybe it can ship outwith the bundle and we can read it directly with some Node filesystem calls?
- Failing all that, I could host it on the WCR server. That would also give it a mechanism to not rely on app updates to stay up to date, and it could be fetched into memory by WCR on launch / once per day etc.

aza547:
Closing as this is stale

PR #764, opened 2025-11-02
Chat feature

aza547:
The general idea here:
- Each video has a chat.
- Chat supports timestamping videos.
- The use case is review of videos to aid progression - raid leader can comment on those and raiders can follow along.
- The addition of a chat panel also seems like a nice place to fit the viewpoint selection which I think could do with being moved.

image

To support it...
- UI updates
- Tag feature dies, tags are migrated to the chat.
- Websocket updates based on existing mechanism for new messages.
- Chats served by a new endpoint GET / POST to `/api/video/:video/chat`? 
- Stored in a new database (or a new table?) (sharding? probably fine without initially)
- Chats are not part of video metadata.
- Chats never stored locally. Only valid for Pro cloud users. Only valid on videos in the cloud. 
- Chats stored on the activity not an individual point of view. Must be shared between them live. 
- Mobile view still needs to look good.
- Custom usernames needs to be allowed. Authentication changes to come with that.
- Custom username will be displayed with the chat message. Not email address as users generally don't want to expose that.

Database probably looks like...

## Table: video_chats
| Column         | Type        | Constraints                  | Notes                           |
|----------------|-------------|-------------------------------|----------------------------------|
| id             | UUID        | PK                            | Unique chat message ID           |
| activity_start_time       | UUID        | FK -> videos(id)              | Associated video                 |
| activity_unique_hash        | UUID        | FK -> users(id)               | Sender                           |
| username       | TEXT        | NOT NULL                      | Custom display name              |
| message        | TEXT        | NOT NULL                      | Chat content                     |
| created_at     | TIMESTAMP   | DEFAULT now()                 |                  

- both the activity* fields are needed to determine the video (as per existing mechanism)
- id can be a uuid for the message
- loading the chats for a video can happen with a `SELECT * WHERE activity_start_time < ? AND activity_start_time > ? AND activity_unique_hash    = ?` - probably sorted on CREATED_AT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 ### Changed
+- Moved the viewpoints selection into an expandable drawer.
+
 ### Added
 - [PR 764](https://github.com/aza547/wow-recorder/pull/764) - Adds the video chat feature for Pro users.
 - Classic PTR "support". The app will accept Classic PTR paths.  As usual, PTR support is best effort.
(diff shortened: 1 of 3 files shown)


aza547:
Still need to handle database housekeeping and also solve chat for clips but this can merge.

PR #768, opened 2025-11-09
Implementing "Record classic ptr" option as a separate button

VolgaStack:
Added "Record classic ptr" option in the "Game" settings
Moved recording classic beta and classic ptr to that button. 
Added translations (only English)
Added window that starts with "WoWClassicT.exe" to list of allowed names that is used as a source for capture


aza547:
thanks for the PR - appreciate the effort, i'm concerned about making it look like classic PTR is a supported content type, it's really best effort in the hope it might just work , that said given this looks sensible maybe all that is needed is a disclaimer in the client

src/utils/configUtils.ts
```diff
@@ -380,6 +380,32 @@ const validateBaseConfig = async (config: BaseConfig) => {
...
+      path.basename(classicPtrLogPath) === 'Logs';
+
+    if (!validPath) {
+      console.error('[Util] Invalid classic PTR log path', classicPtrLogPath);
+      const error = getLocaleError(Phrase.InvalidClassicPtrLogPath);
+      throw new Error(error);
+    }
+  }
+
+  if (recordRetailPtr) {
```
aza547, line 396: I think this is duplicated from above

aza547 (approved):
One minor issue and also needs translations but otherwise looks great. I've kept you waiting long enough so I'll just merge and fix up :)

aza547:
https://github.com/aza547/wow-recorder/commit/01d8c175f5d88fa1c17e4b9906d42abe09618c10

PR #793, opened 2026-02-06
[Draft] Add Linux Support to Warcraft Recorder

sepdotgg:
# Summary

## If anyone wants to use this

I periodically publish AppImage's here: https://github.com/sepdotgg/wow-recorder/releases

## PR

**NOTE:** _DRAFT_ PR for comment and discussion, but is almost completely in a shippable state. The [noobs PR](https://github.com/aza547/noobs/pull/23) and [rust-ps PR](https://github.com/aza547/rust-ps/pull/2) would need to be merged first. In the case of noobs, it would need to both be merged and published to NPM. See the comments in the noobs PR for things remaining to do before noobs can ship.

This PR consists of the changes necessary to achieve Linux support in Warcraft Recorder. As is the case with my [noobs PR](https://github.com/aza547/noobs/pull/23), a detailed summary of changes is included below. I imagine this PR will receive heavy scrutiny (as it should). Additional commits will likely be needed before merging. There are definitely some TODOs in place at the bottom that will need addressing.

Feel free to [DM me](https://discordapp.com/users/239934148704337930) on Discord (@`sep.gg`) if anyone wants to touch base. I see there already an open PR in wow-recorder, so I'm sure there will be discussion and I'm open to hear anyone's ideas and arrive at the best solution.

## High Level Goals
 ... and my general approach to adding linux support wow-recorder
 
- First and foremost, don't break Windows.
- Maintain structural similarity and parity with Windows
  - While there may be alternative ways of doing things on non-Windows platforms, diverging too much impacts long-term maintainability because it becomes difficult to reason about.
- Linux users deserve the same features as Windows users
  - As far as I've tested, this achieves full parity from a functional standpoint with Windows and does not remove functionality, which includes things like per-process audio capture, app launch on startup, and encoders. Obviously, if I missed something, it will be addressed.
- Maintain 'single file' packaging for download. Unfortunately, this means vending libraries (in noobs), but is probably worth the few extra MB for reproducibility and ease of use for the user. I don't want to assume modern linux gamers are comfortable installing *-devl packages, libraries, or compiling. This will only become more true over the next few years. 

# Testing

- `test.py` executed for Retail/Classic/Era/PTR. Results all pass ([gist for ref](https://gist.github.com/sepdotgg/976be29d8e4eb067ba57fe91a739c9b7))
- Builds for both windows and linux [here](https://github.com/sepdotgg/wow-recorder/actions/runs/21744597981) -- assets from linux should be downloadable.
- Screenshots of key bits of UI which have variations on Linux: [here](https://gist.github.com/sepdotgg/976be29d8e4eb067ba57fe91a739c9b7?permalink_comment_id=5972852#gistcomment-5972852)
- This has been pretty heavily tested the last week+ in Retail wow on my Arch/CachyOS machine.
- I've also done sanity testing of core functions on fresh installs of Bazzite and Fedora.
  - I plan on also including sanity on Ubuntu/Mint [according to the Steam survey](https://store.steampowered.com/hwsurvey/Steam-Hardware-Software-Survey-Welcome-to-Steam)
- This has been tested on Windows to ensure no regression, but would appreciate more scrutiny.

# Important Bits

### Electron / Ffmpeg 

The most important callouts are the areas I had the most challenges with: Electron's sandbox and FFmpeg. I summarized these in my noobs PR, but its worth calling out here as well.

The prefix to this is that these issues are solved in two ways: 

  1. Vending compiled libraries in noobs (same as on Windows)
  2. Ensuring these libraries are loaded and take priority over Electron's library loading. This is not an entirely uncommon process for Chromium-based things and Ffmpeg. For example the Vivaldi browser [does this](https://github.com/ric2b/Vivaldi-browser/blob/master/platform_media/docs/linux.md#loading-alternative-libffmpegso-libraries)

This is achieved in "dev mode" by a `./run-dev-linux.sh` script which ensures the correct library paths are set up. It gets messy in package.json, so the script is far simpler. It is achieved in the packaged AppImage by an `afterPack.js` script which customizes the launcher script inside of the AppImage, and loads the vended libraries before launching Electron.

Side note: Long term, I think it would be beneficial to move noobs/obs functionality _entirely_ out of the Electron main process and communicate with it over some form of IPC, or even websocket. Ffmpeg + Electron seems to be a minefield, especially on linux where things like [seccomp](https://en.wikipedia.org/wiki/Seccomp) are also in play.

### Reviewer Notes

- One of the obvious things you'll notice is that `release/app/package.json` points to local noobs. This will be addressed before being merge-ready. It is in place now until noobs review is completed.
- The included github actions for build reference my WIP branch of noobs, in order to validate the full build process. This also gives you access to the Linux AppImage binaries and Electron latest-linux.xml. This will be adjusted for release process as needed before merge.

### Other Bits

- Pipewire-only capture, which requires [Wayland](https://en.wikipedia.org/wiki/Wayland_(protocol)) Trivial concern, considering effectively every modern linux distro, _especially_ gaming focused ones, ship with Wayland+Xwayland+Pipewire+PulseAudio out of the box.
- Game Capture and Monitor capture tabs are hidden on Linux. The Pipewire portal allows the user to select any window or any screen/monitor.
- MP3 sound files for manual recording are converted to OGG. This provides better cross-platform support across the board since OGG is well established on the web and web-based products like electron/chromium.
- Run at Start up is achieved by manually injecting a .desktop file into the user's autostart dir. This is kept updated if the app image binary path changes, and is only triggered when running in packaged mode.
- linux x86 musl compiled version of rust-ps vended. [PR](https://github.com/aza547/rust-ps/pull/2)
- Standardized (electron-native) file/folder opening.
- In general, the changes are straightforward pretty straightforward due to the full feature parity.

# Remaining Items / TODOS

- [ ] [noobs PR](https://github.com/aza547/noobs/pull/23) must be merged and npm published first.
- [ ] Revert `release/app/package.json` from `"noobs": "file:../../../noobs"` to a published npm version
- [X] [rust-ps PR](https://github.com/aza547/rust-ps/pulls) - Merged
- [ ] Update GitHub Actions workflow (`.github/workflows/node.js.yml`) to reference upstream `aza547/noobs`
- [X] Resolve `AudioSourceType` enum value change in `src/main/types.ts`
- [X] Add UI toggle to enable/disable H.265 live transcoding feature. (Linux Only)
- [X] Add UI configuration for H.265 transcode cache size (Linux Only)
- [X] Disable H.265 as a selectable encoder option on Linux
- [x] Investigate "Auto start" not working
- [X] "Windows Settings" label still appears on non-Windows platforms
- [x] Complete sanity testing on Ubuntu/Mint with a Wayland DE


release/app/package.json
```diff
@@ -15,7 +15,7 @@
     "@ffmpeg-installer/ffmpeg": "^1.1.0",
     "@ffprobe-installer/ffprobe": "^2.1.2",
     "atomic-queue": "^5.0.4",
-    "noobs": "^0.0.181",
+    "noobs": "file:../../../noobs",
```
aza547, line 16: I think this is hiding a linker issue, eventually should point to a npm package but when I point it to the tgz it fails to find libobs.so and `npm install` fails with `/usr/bin/ld: cannot find -lobs: No such file or directory`. The Makefile had all sorts of references to the folder where it was built which I was a bit confused by.

sepdotgg: This is a really good catch. I had expected that this would be fixed as soon as the new version was published and this line reverted, but turns out it's [intentional](https://github.com/npm/npm-packlist/blob/main/README.md?plain=1#L44-L47) that `npm pack` does not include symlinks. (libobs.so -> libobs.so.30, libobs.so gets dropped, so linking fails when pointing at the tgz). Should be a straightforward fix to point binding.gyp to the versioned file.

src/main/types.ts
```diff
@@ -290,9 +290,9 @@ interface IDevice {
 }
 
 enum AudioSourceType {
-  OUTPUT = 'wasapi_output_capture',
-  INPUT = 'wasapi_input_capture',
-  PROCESS = 'wasapi_process_output_capture',
+  OUTPUT = 'output',
+  INPUT = 'input',
+  PROCESS = 'process',
```
sepdotgg, line 312: Adding my own note here to resolve this before moving out of draft. This is possibly a breaking change for existing Windows installs.

stegls:
I've been using Warcraft Recorder on Windows for a while and wanted to try it on my Linux setup as well, so I tested it in the last weeks. I'm not a Pro user, so I couldn't test online/cloud features.

**Tested on:** CachyOS, KDE Plasma on Wayland, AMD Ryzen 7800x3D, AMD Radeon RX 9070 XT, OBS 32.1.1. Battle.net and WoW are installed via Steam (Proton Experimental). 

**Video settings** encoder: OBS H.264 (Software), recording at 2560x1440 60fps Ultra quality

**Content types tested:** Manual recording, Mythic+, Raids (LFR/Normal/Heroic/Mythic), PvP just in HC week (a few Arena 3v3, Battlegrounds).

**Results:** Works. Automatic recording detection was reliable across all tested content types – recording started and stopped correctly every time, no missed or wrong recordings. BUT an internet connection loss during a key was detected and marked as "Abandoned", even though I reconnected and we finished the key in-time.

**Recording metadata:** All metadata was accurate – dungeon names, key level, affixes, result (+1/+2/+3/Abandoned), raid encounters, result (Kill / HP% on wipe), pull count, difficulty, duration, date/time, and character were all correctly identified. Deaths (own and others) are correctly parsed from the combat log, and encounter timestamps are accurate.

**UI/Features:** Search and date filters work correctly. Drawing mode, didn't know that existed, so unsure if it works as it does on Windows. Opening the video folder work, Clips, Chatoverlay, Cursor capture work. Max disk storage is correct (oldest recordings are deleted when the limit is reached). Minimize on close and minimize to tray both work.

**Audio:** Audio recording works well. By default it captured all system audio (including background music/videos), but I changed it to capture only WoW (detects it automatically on start), a specific microphone, and Discord - which however is shown as "WEBRTC VoiceEngine" and needed some trail & error.

**Performane:** No crashes, no performance impact noticed.

**Notes:**
- "Auto start" did not work for me ( + label still says "Windows Settings").
- The OBS window/screen capture prompt reappears every time if WoW is launched *after* the recorder. If WoW is already running when the recorder starts, it's fine. Sometimes somewhat annoying.


Would be cool to see Linux support continued, hope this gets released!

funnynin:
Hey, just a heads up, the linux build artifact appears to have expired. Would also like to test (or use 😛) this on linux, so would appreciate if either this gets suddenly merged or more realistically if it's possible to rebuild or otherwise rehost that artifact.

Opie-eg:
> Hey, just a heads up, the linux build artifact appears to have expired. Would also like to test (or use 😛) this on linux, so would appreciate if either this gets suddenly merged or more realistically if it's possible to rebuild or otherwise rehost that artifact.

+1 to this

sepdotgg:
I merged from upstream main, bringing this up to 7.6.4. The latest AppImage can be found [here](https://github.com/sepdotgg/wow-recorder/releases/tag/7.4.6-0-linux). It's now on GitHub releases so my mistakes can be preserved for all history.

@stegls Thanks for the testing and feedback. While I didn't fix either of those two issues in these changes, I have a good idea of the problems.

- **AutoStart:** I'll have to double check on this. Ideally togglign that option should be populating a .desktop file in `.config/autostart/` which KDE Plasma should work fine with, but this was one of the earliest things I added so I might have broken it at some point along the way.
- **Screen Capture Portal:** I'm reasonably confident this is just a race between when warcraft-recorder detects the wow process is running and when the window manager actually has a window for it. wow-recorder immediately triggers a capture request with the PipeWire restore token (so it doesn't need to re-ask), but Wayland/Pipewire sees no such window yet, so it triggers a fresh portal request. Fix here is to probably add a small delay before telling obs to start recording when it detects wow going from a closed to open state.

funnynin:
Apologies, have never used the windows version (seems unlikely I ever will) so I am not entirely sure what is normal behaviour, so tentatively:

(CPU: 9800X3D, GPU 9070 XT, 7.0.3-cachyos, kde plasma 6.6.4, wayland)

- There is only one option under "Scene -> Video -> Video encoder", which is "OBS H.264 (Software)", despite the output in terminal including this:
```
21:11:37.962 › [Recorder] Encoders: [
  'obs_x264',
  'ffmpeg_vaapi',
  'ffmpeg_vaapi_tex',
  'av1_ffmpeg_vaapi',
  'av1_ffmpeg_vaapi_tex',
  'hevc_ffmpeg_vaapi',
  'hevc_ffmpeg_vaapi_tex'
]
```
(standard OBS studio is also able to use e.g. ffmpeg vaapi av1 and it works fine)

I was expecting to run into more issues which is why this is a single bullet point, but I think this is the main one, at least as far as initial testing goes. There is also some output along the lines of 
```
[h264 @ 0x115c07155800] Codec (h264) not on whitelist 'vorbis,libopus,flac,pcm_u8,pcm_s16le,pcm_s24le,pcm_s32le,pcm_f32le,mp3,pcm_s16be,pcm_s24be,pcm_mulaw,pcm_alaw,aac'
```
but I am not confident that this is actually an issue, or related to linux in particular. (noting as a 'just in case' thing).

sepdotgg:
VAAPI is linux only, so this is purely an artifact of the frontend assuming there will only be windows compatible encoders, even though obs is returning the full list.

I guess I never noticed this because I have nvidia hardware, so I also get hardware encoded NVIDIA H.264 and AV1 in addition to x264, so mine always matched what it did on windows.

Easy fix in the frontend, unfortunately I have integrated graphics disabled in bios, But I think one of my mini pcs has an AMD system that supports VAAPI that I can try to test with.

> [h264 @ 0x115c07155800] Codec (h264) not on whitelist 'vorbis,libopus,flac,pcm_u8,pcm_s16le,pcm_s24le,pcm_s32le,pcm_f32le,mp3,pcm_s16be,pcm_s24be,pcm_mulaw,pcm_alaw,aac'

I'm pretty sure this is harmless. As far as I can tell, this happens when ffmpeg is muxing, and instructed to only look at the audio data, but h264 is still part of the file. Unless the final videos aren't being created correctly, I think this is just verbose noise from ffmpeg.

---

EDIT: I built a new AppImage with what I think is the fix for VAAPI. Turns out I swiped the drive out of my AMD mini PC so I'll need to reinstall a distro there. Might not get to it for a few days, but you can test the appimage now if you want: https://github.com/sepdotgg/wow-recorder/actions/runs/25588853602

funnynin:
Yep, I'm seeing VAAPI AV1 on the frontend now. Unfortunately, selecting either hardware encoder causes warcraft recorder to instantly crash when it tries to start the replay buffer (in practice, this means crashing when opening wow, or immediately crashing if wow is already open when starting wcr), with these being the last couple lines of each log file:

WarcraftRecorder-2026-05-09.log:
```
[2026-05-09 12:15:43.042] [info]  [Recorder] Created audio source WCR Audio Source 4
[2026-05-09 12:15:43.043] [info]  [Recorder] Queued start buffer
[2026-05-09 12:15:43.043] [info]  [Recorder] Start
```

OBS-2026-05-09.log:
```
[2026-05-09 12:15:43.043] [INFO]  ObsInterface::startBuffering called
[2026-05-09 12:15:43.043] [INFO]  ---------------------------------
[2026-05-09 12:15:43.043] [INFO]  [FFmpeg VAAPI encoder: 'noobs_file_encoder'] settings:
[2026-05-09 12:15:43.043] [INFO]  	device:       /dev/dri/renderD128
[2026-05-09 12:15:43.043] [INFO]  	rate_control: CQP
[2026-05-09 12:15:43.043] [INFO]  	profile:      100
[2026-05-09 12:15:43.043] [INFO]  	level:        -99
[2026-05-09 12:15:43.043] [INFO]  	qp:           30
[2026-05-09 12:15:43.043] [INFO]  	bitrate:      0
[2026-05-09 12:15:43.043] [INFO]  	maxrate:      0
[2026-05-09 12:15:43.043] [INFO]  	keyint:       60
[2026-05-09 12:15:43.043] [INFO]  	width:        1920
[2026-05-09 12:15:43.043] [INFO]  	height:       1080
[2026-05-09 12:15:43.043] [INFO]  	b-frames:     0
[2026-05-09 12:15:43.043] [INFO]  	ffmpeg opts:  
[2026-05-09 12:15:43.043] [WARN]  [FFmpeg VAAPI encoder: 'noobs_file_encoder'] Failed to create VAAPI device context: Cannot allocate memory
[2026-05-09 12:15:43.043] [ERROR]  Failed to start buffering!
```

STDOUT/STDERR (i.e. terminal output):
```
12:24:39.035 › [Recorder] Queued start buffer
12:24:39.035 › [Recorder] Start
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to start buffering!
[1]    15044 IOT instruction (core dumped)  ./WarcraftRecorder-7.6.4.AppImage
```

The OBS log output seems the most relevant, but have included the rest in case it helps. (Again, this is the end of each file, so there is no further output.) Also, h264 (software) still works fine. I suppose this is potentially more of an indicator of a problem with the noobs PR rather than this one, but still.

Otherwise, can also confirm at this point that "Run on Startup" works, though given I'm using KDE, that's probs expected.
Speaking of "frontend artifacts" - re: the "Windows Settings" heading, maybe could be like "{OS} Settings" or literally "Operating System Settings"? Though this is obviously not a functional issue.



> I'm pretty sure this is harmless.

Me too, though it was showing up when _viewing_ videos in-app. Your comment about muxing made want to test clipping though, which also works fine and doesn't even show potentially erroneous output (just all the info about the cut and then "finished processing" etc.)

As an aside to all that, I would probably prefer that any automatic attempts to capture wow fail silently rather than asking for a new window selection, especially when launching wow. (As noted previously it will attempt too early in this case). Based on your phrasing, this sounds like a wayland/pipewire issue, (explains why this annoying behaviour is so common,) but if there's anything that can be done (at least that ~10s delay on wow startup if wcr was launched first,) that would be great.

sepdotgg:
1. VAAPI - I installed a Wayland based DE on my AMD mini PC and tested a new fix. Seems this was another case of Electron's ffmpeg competing with the bundled one. Added the utilities library needed for VAAPI to the preload script. New artifact that I tested is [here](https://github.com/sepdotgg/wow-recorder/actions/runs/25612702313)
2. I'll try to put a fix in for the pipewire race and test in soon.

funnynin:
Not much more to add right now other than it's working over here afaics. :+1: 

VAAPI catches intel dgpus and nvidia nouveau drivers too, right? (I can't really easily test these cases but if support for them is still broken I suppose the fix is likely to be similar.)

sepdotgg:
The H.265 that was just brought in from upstream is ... problematic, to say the least, and I'm not sure yet how to deal with it.

Because of patent licensing, _encoding_ to H.265 is trivial because NVIDIA/AMD have already paid the patent fees.

_Playing back_ that video in Electron/Chromium is a problem, because Chromium's `<video>` (at least on Linux), flat out does not have an h265 decoder, and it doesn't even _try to load it_ such that we could inject one the same way we inject other ffmpeg libraries.

VAAPI users _might_ be able to get it, but nvidia is probably a non-starter without hacks to pipe the video through a software decoder.

sepdotgg:
Latest changes:

- Merge from `upstream/main`
- Added some better logic (I think) for when to trigger adding the Pipewire/Window capture source to help mitigate unnecessary portals.
  - The window capture source should now _never_ be called unless wow is running
  - There is a 10s delay between when rust-ps detects wow has moved to a running state before the source is added to the scene. This gives the compositor a chance to actually spawn a window that piprewire knows about.
  
### H265
For H265, I'm taking advantage of the fact that we have the power of `ffmpeg` available. For videos which have h265/hevc as the encoder in metadata (thanks for adding this!), this will now attempt to do realtime transcoding and cache the transcoded videos and swap out the URL in the \<video\>. This uses an LRU cache tracked by a json file of the last time the video was played, and evicts when the transcoded video cache dir reaches 10GB (hardcoded for now).


If this is an option that can be moved forward with, I'm planning on adding configuration options in the app that a. toggles this feature on or off entirely and b. allows configuration of the cache size on disk.

I'm also planning on entirely disabling H265 as an encoder option on Linux. Transcoding basically every local video is unsustainable.

sepdotgg:
Misc callout since `libfuse2` was mentioned previously in this PR.

The latest version of `electron-builder` [now allows](https://github.com/electron-userland/electron-builder/pull/9558) statically linking fuse into the AppImage runtime, which eliminates that requirement. (As of 26.8.0+)

src/main/types.ts
sepdotgg: Addressed in: https://github.com/sepdotgg/wow-recorder/commit/a7fd3e9cc001f9a238a75b48aa0f75ef22792bee

sepdotgg:
Latest pushes:

**H265**
- There are now 2 new Linux-only options in Application Settings
  - Enable H265 transcoding (default on)
  - Transcoding cache size
- The transcoding cache dir is stored as a hidden folder under the main storage path much like the default .temp buffer folder is.
- When transcoding is disabled, any H265 video that is selected for playback will instead have a message displayed and giving the user a button to go to settings to turn it on. This should also work per-cell for multi-pov videos, but would appreciate some more testing around this since fully exercising it requires I have guild mates who a) keep their app updated and b) select h265 :(
- Disables H265 encoding options entirely on Linux.

**General**

- Renamed the 'Windows Settings' header to 'Application Settings' on all platforms. It's slightly redundant given that it's already on the 'Application' page, but it seems accurate to what is actually in that subsection.
- I looked more into the `Auto Start` issue reported. I honestly can't find a bug here. As far as I can tell, it conforms to the [FreeDesktop](https://specifications.freedesktop.org/autostart/latest/) spec for autostart. My best guess, not sure if it's right: 
  1. Launched app, enabled autostart
  2. Moved the `.AppImage` file location on disk. This means the `.desktop` is pointing to a missing file.
  3. Rebooted. Autostart didn't work.
  - Manually re-launched from the new `.AppImage` location should cause the app to automatically update the location in the .desktop file.

AppImage artifact [here](https://github.com/sepdotgg/wow-recorder/actions/runs/25839433478/artifacts/6986408301). I opted not to make a GH Release out of this one for now.

LithianFI:
I was testing this in keys today and had a GPU crash occur with journalctl referring to WarcraftRecorder:

May 14 23:09:07 Ashyn kernel: amdgpu 0000:03:00.0: Dumping IP State
May 14 23:09:07 Ashyn kernel: amdgpu 0000:03:00.0: Dumping IP State Completed
May 14 23:09:07 Ashyn kernel: amdgpu 0000:03:00.0: [drm] AMDGPU device coredump file has been created
May 14 23:09:07 Ashyn kernel: amdgpu 0000:03:00.0: [drm] Check your /sys/class/drm/card1/device/devcoredump/data
May 14 23:09:07 Ashyn kernel: amdgpu 0000:03:00.0: ring gfx_0.0.0 timeout, signaled seq=12782752, emitted seq=12782754
May 14 23:09:07 Ashyn kernel: amdgpu 0000:03:00.0:  Process warcraftrecorde pid 16410 thread warcraftre:cs0 pid 16506
May 14 23:09:07 Ashyn kernel: amdgpu 0000:03:00.0: Starting gfx_0.0.0 ring reset
May 14 23:09:07 Ashyn kernel: [drm:gfx_v11_0_bad_op_irq [amdgpu]] *ERROR* Illegal opcode in command stream
May 14 23:09:08 Ashyn baloo_file_extractor[23681]: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could 
not register app ID: App info not found for 'org.kde.baloo'")
May 14 23:09:09 Ashyn baloo_file_extractor[23697]: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could 
not register app ID: App info not found for 'org.kde.baloo'")
May 14 23:09:09 Ashyn kernel: amdgpu 0000:03:00.0: MES failed to respond to msg=RESET
May 14 23:09:09 Ashyn kernel: amdgpu 0000:03:00.0: failed to reset legacy queue
May 14 23:09:09 Ashyn kernel: amdgpu 0000:03:00.0: reset via MES failed and try pipe reset -110
May 14 23:09:09 Ashyn kernel: amdgpu 0000:03:00.0: The CPFW hasn't support pipe reset yet.
May 14 23:09:09 Ashyn kernel: amdgpu 0000:03:00.0: Ring gfx_0.0.0 reset failed
May 14 23:09:09 Ashyn kernel: amdgpu 0000:03:00.0: GPU reset begin!. Source:  1

sepdotgg:
From some digging through source code and API signatures, as far as I can tell, this crash is in Mesa driver code. My first thought was VAAPI related, but this doesn't seem like it's on the encoder path. It's also on the Mesa main command thread (rather than on Chromium's GPU thread), so that seems to point to obs -> opengl -> mesa. The problem is the journalctl gives basically no info other than the GPU rejected or hung on the opcode that mesa sent.

It's possible that there's some knobs that could be turned in libobs like colorspace that seem to go through these codepaths, but those would all be workarounds rather than bug fixes, and this is low level and generic enough that I'm honestly not even sure where to start. If it's regularly reproducible on certain hardware GPU/Mesa driver versions, or reproducible in stock obs-studio, that might be a start to identify a pattern.

sepdotgg:
From the last push:

- Minor changes to the way the filesystem validation is called to hoist the platform check one level above `getDriveFormat` to make it clear that it is being short-circuited on non-Windows platforms.
- Otherwise, it is a no-op on Linux. Even if a user has selected an exFAT/FAT path (god I hope not), it will technically work fine on Linux since `fs.watch` operates via [inotify](https://man7.org/linux/man-pages/man7/inotify.7.html), which works at the VFS layer rather than the filesystem driver layer.

Additionally:

- I updated the GitHub actions on my branch to also start building and publishing an (unsigned) Windows installer so that Windows regression testing of the PR is simpler. ([example](https://github.com/sepdotgg/wow-recorder/actions/runs/26736064883))

sepdotgg:
Latest merge is again built off of upstream/main, contains the latest obs-studio+ noobs builds, which means the latest release in my linux repo is techinically the unreleased 7.10.0: https://github.com/sepdotgg/wow-recorder/releases and contains the Instant Replay feature (among other changes).

sepdotgg:
Latest push:

- Same wow-recorder/noobs upstream code as before
- Fixes an issue due to the recent Electron 42 version bump that caused Electron to launch in Wayland mode automatically (the PID handoff to obs for preview rendering doesn't work with wayland), and forces it back to X11.
- The repackaged build `7.1.1.1-linux` ([Releases](https://github.com/sepdotgg/wow-recorder/releases)) contains a new version of noobs linked against libpci 3.12 to fix an incompatibility with SteamOS that a Discord user DM'd me about. Details [in the noobs PR](https://github.com/aza547/noobs/pull/23#issuecomment-5126053835)

PR #796, opened 2026-02-21
Warn When Advanced Combat Logging is Disabled

### LINKED ISSUE #695: Parse Config.wtf and check for advanced combat logging
Check for advanced combat logging.

The config.wtf file has :

SET advancedCombatLogging "1"

Would be nice to parse that and warn users if it is off. It trips a bunch of people up.

NathanPrice:
Closes #695

### What
- Reads each enabled flavor's `Config.wtf` file to check the `advancedCombatLogging` setting.
- Shows per-flavor warning text in the settings page under each game version that has it disabled.
- Watches `Config.wtf` files for changes so warnings update reactively without restarting the app.
- Added a localized warning phrase in all 4 languages (English, Korean, German, Chinese Simplified).
- Updated the changelog.

### Why
Users often forget to enable Advanced Combat Logging in WoW, which results in recordings with missing or incorrect combat data. This should hopefully make it easier for people to notice that.

### How to test
1. Enable recording for a WoW flavor (ex: Classic).
2. In WoW's `_classic_\WTF\Config.wtf`, set `advancedCombatLogging` to `"0"` (or remove the line).
3. Open the app, the settings page should show red warning text under that flavor.
4. Hover the status card, the flavor should have a ⚠ icon next to it.
5. Change `advancedCombatLogging` back to `"1"` in Config.wtf or change the setting in game. The warning should disappear without restarting.
6. Toggle a flavor off and back on. The warning state should persist correctly.
7. Switch app language and confirm the warning text is localized.


src/main/util.ts
```diff
@@ -412,6 +412,47 @@ const getWowFlavour = (pathSpec: string): string => {
...
+  const configWtfFile = getConfigWtfPath(logPath);
+
+  console.info(
+    '[Util] Checking advanced combat logging:',
+    logPath,
+    '->',
+    configWtfFile,
+  );
+
+  if (!fs.existsSync(configWtfFile)) {
```
aza547: Please make this function async - it's bad for performance of the app to use sync methods here - they block the javascript event loop and prevent other code being executed while waiting for disk IO.

src/main/util.ts
```diff
@@ -412,6 +412,47 @@ const getWowFlavour = (pathSpec: string): string => {
...
+    console.warn('[Util] Config.wtf not found at', configWtfFile);
+    return false;
+  }
+
+  const content = fs.readFileSync(configWtfFile).toString();
+
+  const match = content.match(/^SET advancedCombatLogging\s+"(\d+)"/m);
```
aza547, line 440: Regex and capture group is probably overkill here, I think you can just `content.includes("SET advancedCombatLogging 1")`? Minor though, fine to leave.

src/main/util.ts
```diff
@@ -412,6 +412,47 @@ const getWowFlavour = (pathSpec: string): string => {
...
+
+  if (!match) {
+    console.info('[Util] advancedCombatLogging not found in', configWtfFile);
+    return false;
+  }
+
+  const enabled = match[1] === '1';
+  console.info(
+    '[Util] advancedCombatLogging =',
+    match[1],
```
aza547: I don't think you need to log both of these variables here - it's 0 or 1 and you're also logging if it's enabled.

src/renderer/App.tsx
```diff
@@ -47,6 +48,8 @@ const WarcraftRecorder = () => {
   const [errorReports, setErrorReports] = useState<ErrorReport[]>([]);
   const updateNotified = useRef(false);
   const { toast } = useToast();
+  const [advancedLoggingStatus, setAdvancedLoggingStatus] =
+    useState<AdvancedLoggingStatus | null>(null);
```
aza547: You can avoid the null here and subsequent null checks in Status.tsx by initializing all fields to true.

src/renderer/containers/ApplicationStatusCard/Status.tsx
```diff
@@ -206,6 +214,17 @@ const Status = ({
                 {': '}
               </span>
               <code>{config.retailLogPath}</code>
+              {advancedLoggingStatus && !advancedLoggingStatus.retail && (
```
aza547: If you initialize this in App.tsx you can avoid `advancedLoggingStatus && ...` here.

src/renderer/containers/ApplicationStatusCard/Status.tsx
```diff
@@ -206,6 +214,17 @@ const Status = ({
...
+                <span
+                  className="text-destructive ml-1"
+                  title={getLocalePhrase(
+                    language,
+                    Phrase.AdvancedCombatLoggingDisabledWarning,
+                  )}
+                >
+                  {'\u26A0'}
```
aza547: This should probably use a lucide-react icon rather than a UTF8 character.
https://lucide.dev/icons/triangle-alert perhaps?
Also see the <Tooltip> component from shadcn that is probably better than the span here to be consistent with other components. 

```
<Tooltip content={...}>
  <TriangleAlert size={XX} />
</Tooltip>
```

src/renderer/FlavourSettings.tsx
```diff
@@ -179,6 +181,14 @@ const FlavourSettings: React.FC<IProps> = (props: IProps) => {
                 )}
               </span>
             )}
+            {advancedLoggingStatus && !advancedLoggingStatus.retail && (
```
aza547: Another place that could be a bit simpler without the null possibility

src/main/Manager.ts
```diff
@@ -280,6 +292,79 @@ export default class Manager {
...
+    if (this.cfg.get<boolean>('recordClassicPtr'))
+      logPaths.push(this.cfg.get<string>('classicPtrLogPath'));
+
+    const watched = new Set<string>();
+
+    for (const logPath of logPaths) {
+      if (!logPath) continue;
+
+      const configPath = getConfigWtfPath(logPath);
+      if (watched.has(configPath) || !fs.existsSync(configPath)) continue;
```
aza547: async please - not existsSync :) I think there is a util method for this - see util.ts / exists().

aza547 (commented):
A few minor bits to resolve. Thanks again for the PR. The approach is spot on.

src/main/main.ts
```diff
@@ -234,6 +234,8 @@ const createWindow = async () => {
       cloud.refreshStatus(),
       cloud.refreshVideos(),
     ]);
+
+    manager.checkAdvancedLogging();
```
aza547: Is there a reason this needs to be here? Isn't it enough that it is in reconfigureBase?

src/main/Manager.ts
```diff
@@ -280,6 +292,79 @@ export default class Manager {
...
+        checkAdvancedCombatLogging(this.cfg.get<string>('retailPtrLogPath')),
+      classicPtr:
+        !this.cfg.get<boolean>('recordClassicPtr') ||
+        checkAdvancedCombatLogging(this.cfg.get<string>('classicPtrLogPath')),
+    };
+
+    console.info('[Manager] Advanced logging status:', JSON.stringify(status));
+    send('updateAdvancedLoggingStatus', status);
+
+    this.watchConfigWtfFiles();
```
aza547: A bit confused by this. I think you probably don't want to call watchConfigWtfFiles here?

`checkAdvancedLogging` is called in your FsWatcher callback:

```
        const watcher = fs.watch(configPath, () => {
          console.info('[Manager] Config.wtf changed:', configPath);
          this.checkAdvancedLogging();
        });
```

So every time a Config.wtf changes you are going to hit this line, read the config a bunch and reconfigure your watcher. I think you only need to reconfigure your watcher when the log paths change in config (i.e. reconfigureBase).

aza547: In summary I think you need to:
- Remove checkAdvancedLogging() from main.ts
- Replace line Manager.ts:190 with `this.watchConfigWtfFiles();` 
- Delete line Manger.ts:321. 

Let me know if I'm missing something.

src/main/Manager.ts
```diff
@@ -280,6 +292,79 @@ export default class Manager {
...
+      logPaths.push(this.cfg.get<string>('eraLogPath'));
+    if (this.cfg.get<boolean>('recordRetailPtr'))
+      logPaths.push(this.cfg.get<string>('retailPtrLogPath'));
+    if (this.cfg.get<boolean>('recordClassicPtr'))
+      logPaths.push(this.cfg.get<string>('classicPtrLogPath'));
+
+    const watched = new Set<string>();
+
+    for (const logPath of logPaths) {
+      if (!logPath) continue;
```
aza547: If you only call this method after reconfigureBase I don't think you need to worry about this being falsy, or otherwise invalid, because we will have already it is valid.

src/main/Manager.ts
```diff
@@ -280,6 +292,79 @@ export default class Manager {
...
+
+  /**
+   * Watch Config.wtf files for changes so the advanced logging status
+   * updates reactively when the user toggles the setting in WoW.
+   */
+  private watchConfigWtfFiles() {
+    this.configWtfWatchers.forEach((w) => w.close());
+    this.configWtfWatchers = [];
+
+    const logPaths: string[] = [];
```
aza547: Why not make this a set to avoid the deduplication logic later?

src/main/Manager.ts
aza547: In-fact as earlier per other comment, I think you can rely on this existing. We check it in `validateBaseConfig` already

src/main/util.ts
NathanPrice: Done. This is now async.

src/main/util.ts
NathanPrice: Tried doing the thing you described here with includes but couldn't get it working. Not sure if I was doing something wrong or not but I just went ahead and left the regex in.

src/renderer/App.tsx
NathanPrice: Done. Went ahead and made them all initialized as true.

src/main/util.ts
NathanPrice: Don't remember why I did this but went ahead and took this out.

src/renderer/containers/ApplicationStatusCard/Status.tsx
NathanPrice: Wasn't aware. Took it out of here and moved the initialization to where you said.

src/renderer/containers/ApplicationStatusCard/Status.tsx
NathanPrice: Ya you're right. I went ahead and imported the icon for this from this. I also did the thing you mentioned earlier and moved the warning icon to be at the start instead of the end.

src/main/main.ts
NathanPrice: Did this the handle the case where a user refreshes the app with Ctrl+R. The setup in reconfigureBase only handles the check on initial startup. Maybe there's a better way to handle this but not sure. Figured I'd handle the edge case here regardless.

src/main/Manager.ts
NathanPrice: Did items 2 and 3 you mentioned here. The first thing you mentioned I answered on a previous comment you made about it.

src/main/Manager.ts
NathanPrice: Fixed the sync issue. Still need to check / look into the other thing you mentioned.

src/main/main.ts
aza547: Oh, fair enough. A minor grip about this is that it will do extra IO work on CTRL + R being hit rather than just re-pushing some already computed state. But I don't think that's the end of the world.

aza547: ...and that said it's not like the surrounding code is any better :)

src/main/Manager.ts
```diff
@@ -280,6 +301,85 @@ export default class Manager {
...
+      classicPtr:
+        !this.cfg.get<boolean>('recordClassicPtr') ||
+        (await checkAdvancedCombatLogging(
+          this.cfg.get<string>('classicPtrLogPath'),
+        )),
+    };
+
+    console.info('[Manager] Advanced logging status:', JSON.stringify(status));
+    send('updateAdvancedLoggingStatus', status);
+    await this.watchConfigWtfFiles();
```
aza547: Still confused by this.

src/main/Manager.ts
```diff
@@ -280,6 +301,85 @@ export default class Manager {
     this.redrawPreview();
   }
 
+  /**
+   * Check Config.wtf for each configured WoW flavour and warn the user
+   * if advanced combat logging is not enabled.
+   */
+  public async checkAdvancedLogging() {
+    const status: AdvancedLoggingStatus = {
```
aza547: If you made this a member variable you could have a separate function to push the state to the frontend on CTRL+R being hit.

aza547: I think that will also save a double call on start up here (in both main.ts and Manager.startup).

src/main/util.ts
aza547: OK, not a big deal.

src/main/Manager.ts
NathanPrice: My bad, that was some debug code I had when I was testing something. Went ahead and took it out.

src/main/Manager.ts
NathanPrice: Ya you're right. Went ahead and turned this into a member variable so it caches the result so it doesn't have to read the Config.wtf file again every time this happens.

NathanPrice:
Hi, just following up on this. Figured you're busy with stuff as well but wanted to make sure to see if what if I had here right now is fine or if you want any more changes. I have some more free time this weekend to look over and fix anything you don't like or is wrong.

aza547:
I'll look again this weekend. Thanks!

src/renderer/components/HoverCard/HoverCard.tsx
```diff
@@ -11,19 +11,21 @@ const HoverCardContent = React.forwardRef<
...
-    className={cn(
-      'z-50 w-64 font-sans rounded-md border-popover-border ring-1 ring-inset ring-popover-inset bg-popover p-4 text-popover-foreground shadow-md',
-      'outline-none data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0',
-      'data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2',
-      'data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
-      className,
-    )}
-    {...props}
-  />
+  <HoverCardPrimitive.Portal>
```
aza547, line 14: Portals confuse me, I think this is a pretty advanced technique. What is the reason for it? 

You have also changed the behaviour of the status card so that it's wider. I think that looks better, but the reason it's not been like that historically is because of the scene page and the native preview display.

Image

Notice how it clips. 

I've been meaning to do something about this for ages. The scene dropdowns enable a flag that hides the preview content. That could also be applied to the status card. 

Image

See `setPreviewEnabled`. Might need some plumbing to make it available to the status card component.

src/renderer/containers/ApplicationStatusCard/Status.tsx
```diff
@@ -326,7 +405,7 @@ const Status = ({
           </div>
         </div>
         {!!statusDescription && (
-          <HoverCardContent className="w-[260px] mx-4">
+          <HoverCardContent className="w-fit mx-4">
```
aza547, line 415: I guess this is the change of the status card width.

src/localisation/english.ts
```diff
@@ -506,6 +506,8 @@ const ENGLISH: Translations = {
   [Phrase.StartManualRecordingTooltip]: 'Start a manual recording.',
   [Phrase.StopManualRecordingTooltip]: 'Stop the current manual recording.',
   [Phrase.RecordedAt]: 'Recorded at',
+  [Phrase.AdvancedCombatLoggingDisabledWarning]: 'Advanced Combat Logging is not enabled. Enable it in WoW: System > Network > Advanced Combat Logging.',
```
aza547: I found when testing that WoW needs to be closed to flush Config.wtf to disk. Should we providing some additional advice here? The warning won't disappear without that when checking the box in-game. That's a bit misleading and might confuse users. Maybe a compromise between having an essay on screen vs being clear how to resolve the issue is needed?

aza547:
I made a few tweaks to some ordering / logging / error handling. Could you review and apply please?

[changes.patch](https://github.com/user-attachments/files/26644218/changes.patch)

src/main/Manager.ts
```diff
@@ -136,6 +158,14 @@ export default class Manager {
     if (success) {
       this.setConfigValid();
       this.poller.start();
+      this.checkAdvancedLogging()
```
aza547: I think there is technically a race condition in here...

do the check, it passes
user disables advanced logging
start watching path
no warnings

very unlikely in reality i'm sure but see my patch for a fix

NathanPrice: Good catch, didn't think about or realize that. I applied the patch you gave in commit https://github.com/aza547/wow-recorder/commit/df72bc320fb63fa35b42ee53708c687445702b2b so watchConfigWtfFiles() now runs before the first checkAdvancedLogging() in both configureBase and reconfigureBase.

src/localisation/english.ts
NathanPrice: Ya true. Went ahead and updated the phrases I had to be a bit clearer on this.

src/renderer/components/HoverCard/HoverCard.tsx
NathanPrice: So, the reason I originally did the portal thing was the status card's inner wrapper (#status-card-inner) sat inside a relative z-10 stacking context. The popup was working intermittently before and stopped appearing once I added the advancedLoggingStatus prop. I had to research this a bit and look through the code but from what I could tell, portal renders the popup at the document root and prevents the clipping. Saw something similar in Popover.tsx and Dialog.tsx and just applied it to HoverCard. I went ahead and did what you said here though by lifting the previewEnabled state from SceneEditor into App so it could be shared with the application status card. Maybe there's a better way to do there's a better way to go about what I did here but I'm not 100% certain.

src/renderer/containers/ApplicationStatusCard/Status.tsx
NathanPrice: Ya adjusted the width a bit so it looked better but that introduced that clipping issue you mentioned before. Thought it was a bit harder to read squished down.

NathanPrice:
Went ahead and replied to everything and pushed my changes up and applied your patch let me know if there's any issues here.

src/renderer/components/HoverCard/HoverCard.tsx
aza547: Ah - I care a bit less if it's already in use elsewhere. Thanks. Still confused by portals in general but I won't hold this up.

aza547:
> Went ahead and replied to everything and pushed my changes up and applied your patch let me know if there's any issues here.

Great - only gripe here is the wording of the warning. I think users will be confused by "Config.wtf" unless they are really savvy. But I'll merge and consider that on main. Thanks for the effort, this is a nice improvement :)

Issue #810, opened 2026-04-12
Cloud status card mouseover should be wider and hide the scene preview


aza547:
...like the Recorder status card does.

aza547:
https://github.com/aza547/wow-recorder/pull/796 changed the recorder status card but not the cloud status card

NathanPrice:
Could try fixing this if needed but not sure if that code is in this repo as far as I know but could be wrong.

aza547:
Would be great if you could take a look :) It is in this repo, see CloudStatusCard.tsx / CloudStatus.tsx, iirc it's a copy paste of the AppStatus / AppStatus Card.

PR #811, opened 2026-04-13
Hide scene preview and widen hover card on cloud status card.

NathanPrice:
## Closes

- #810

## What

 - Wires the cloud status card's HoverCard to setPreviewEnabled so the scene preview hides while the popup is open. The recorder card already did this after #796; the cloud card got missed.
 - Drops the fixed w-[260px] on the cloud hover card in favour of w-fit so it sizes to its content.
 - Passes setPreviewEnabled through SideMenu > CloudStatusCard > CloudStatus. There's no new state. The lifted state from #796 already lives in App.tsx.

## Why

- #810 pointed out the cloud card's popup was narrower than the recorder one and didn't hide the scene preview.
- CloudStatus.tsx and CloudStatusCard.tsx are basically copies of the recorder files that I forgot to touch during #796.

## How to test

1. Open the app on the Scene Editor page with the WoW preview showing.
2. Hover the cloud status card and check:
    - Nothing gets clipped.
    - The scene preview disappears while the popup is open, and comes back when it closes.
    - The popup width fits its content rather than being stuck at ~260px.
3. Hover the recorder card to make sure its hide-preview behaviour still works.
4. Open the cloud popup, then slide over to the recorder card without closing it. The preview should stay hidden the whole time.

### DIFF
--- a/src/renderer/SideMenu.tsx
+++ b/src/renderer/SideMenu.tsx
@@ -263,7 +263,10 @@ const SideMenu = (props: IProps) => {
         advancedLoggingStatus={advancedLoggingStatus}
         setPreviewEnabled={setPreviewEnabled}
       />
-      <CloudStatusCard appState={appState} />
+      <CloudStatusCard
+        appState={appState}
+        setPreviewEnabled={setPreviewEnabled}
+      />
       <Separator className="mb-4" />
       <ScrollArea
         className="w-full h-[calc(100%-80px)]"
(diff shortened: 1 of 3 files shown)


NathanPrice:
Think I forgot to update the changelog file but if everything else is fine with this I can update it when I get some free time.

aza547:
thanks again!
fine to skip the changelog, not something users really need to know about :)

Issue #752, opened 2025-10-13
Marker alignment off on fullscreening


aza547:
https://discord.com/channels/1004860808737591326/1004860988081832196/1426630398997430313

PR #816, opened 2026-04-29
[codex] fix fullscreen progress markers

Nnoggie:
## Summary
- Refresh video progress marker layout when the progress slider changes size.
- Listen for fullscreen transitions and re-render after the fullscreen layout settles.
- Keep existing window resize behavior for normal resizing.

## Root Cause
The marker gradients are calculated from the slider width. Normal window resizing triggered a re-render, but fullscreen enter/exit did not reliably fire a resize at the right time, so gradients kept using stale width until the next interaction.

## User Impact
Encounter/round/death marker alignment now updates immediately when entering or exiting fullscreen.

## Validation
- `npm run build` passed.
- `npx eslint src\renderer\VideoPlayer.tsx` passed.
- `git diff --check` passed.
- `npm test` was attempted per `docs/CONTRIBUTING.md`, but failed before running tests on existing repo issues: Jest haste package-name collision between root and `release/app`, stale activity test constructor args, and unresolved `config/ConfigService` alias in tests.

### DIFF
--- a/src/renderer/VideoPlayer.tsx
+++ b/src/renderer/VideoPlayer.tsx
@@ -164,9 +164,9 @@ export const VideoPlayer = forwardRef<VideoPlayerRef, IProps>((props, ref) => {
   const [clipStartValue, setClipStartValue] = useState<number>(0);
   const [clipStopValue, setClipStopValue] = useState<number>(100);
 
-  // This exists to force a re-render on resizing of the window, so that
-  // the coloring of the progress slider remains correct across a resize.
-  const [, setWidth] = useState<number>(0);
+  // This exists to force a re-render when the progress slider changes size,
+  // so that marker gradients are recalculated against the current width.
+  const [, forceProgressSliderLayout] = useState<number>(0);
 
   // We show a progress spinner until the video is ready to play.
   const [spinner, setSpinner] = useState<boolean>(true);
@@ -762,6 +762,10 @@ export const VideoPlayer = forwardRef<VideoPlayerRef, IProps>((props, ref) => {
     setMuted(!muted);
   };
 
+  const refreshProgressSliderLayout = useCallback(() => {
+    forceProgressSliderLayout((value) => value + 1);
+  }, []);
+
   /**
    * Return an appropriate volume icon for the muted and volume state.
    */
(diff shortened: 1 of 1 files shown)


aza547:
This fixes https://github.com/aza547/wow-recorder/issues/752

src/renderer/VideoPlayer.tsx
```diff
@@ -762,6 +762,10 @@ export const VideoPlayer = forwardRef<VideoPlayerRef, IProps>((props, ref) => {
     setMuted(!muted);
   };
 
+  const refreshProgressSliderLayout = useCallback(() => {
```
aza547, line 765: This useCallback seems like AI being overly ambitious...

src/renderer/VideoPlayer.tsx
```diff
@@ -1300,17 +1304,41 @@ export const VideoPlayer = forwardRef<VideoPlayerRef, IProps>((props, ref) => {
...
   }, []);
 
-  // This hook updates some state to force a re-render on window resize,
-  // otherwise resizing the window (and hence the progress bar) causes
-  // all the makers to be offset until next render.
+  // This hook updates some state to force a re-render on progress slider resize,
+  // otherwise resizing the slider causes all the markers to be offset until next
+  // render. Fullscreen changes do not reliably fire a window resize at the
+  // right time, so observe the slider itself and listen for fullscreen changes.
   useLayoutEffect(() => {
```
aza547, line 1311: Do you need to track the screenfull stuff at all here? To me it looks like the bug is that we're tracking window size instead of progress slider size. I think this is simpler? Have not tested.

```
useLayoutEffect(() => {
  const slider = progressSlider.current;
  if (!slider) return;

  const resizeObserver = new ResizeObserver(() => {
    setWidth((v) => v + 1);
  });

  resizeObserver.observe(slider);

  return () => {
    resizeObserver.disconnect();
  };
}, [setWidth]);
```

aza547 (changes requested):

aza547:
I tested my suggestion and fixed this on main so closing this, thanks for the PR
https://github.com/aza547/wow-recorder/commit/9311c594133a4ddc553bb8f9642e591d1cf81cad

PR #817, opened 2026-04-30
[codex] Fix Mythic+ timer result handling

Nnoggie:
﻿## Summary

- Update Algeth'ar Academy's Mythic+ timer to 31:00, with matching +2/+3 thresholds.
- Preserve fractional combat-log challenge mode duration instead of rounding to whole seconds.
- Treat runs that finish before `timer + 1s` as timed, matching WoW's visible timer behavior.
- Remove the legacy Challenger's Peril timer adjustment from upgrade calculation.

## Context

The previous M+ result calculation could mark very tight timed keys as depleted because the combat-log duration was rounded before comparison. This keeps the raw fractional duration and applies the one-second timing window used by the game UI.

Challenger's Peril also no longer needs a timer-specific adjustment in recorder. Blizzard's [The War Within Season 2 Mythic+ Updates Ahead](https://worldofwarcraft.blizzard.com/en-us/news/24174877/the-war-within-season-2-mythic-updates-ahead) notes describe the Season 2 affix updates, including Challenger's Peril being replaced at +7 and the death-penalty behavior moving to Xal'atath's Guile at +12.

## Validation

- `npx jest src/__tests__/activitys/ChallengeModeDungeon.test.ts --runInBand --silent`
- `npx eslint src/activitys/ChallengeModeDungeon.ts src/parsing/RetailLogHandler.ts src/main/constants.ts src/__tests__/activitys/ChallengeModeDungeon.test.ts`

### DIFF
--- a/src/__tests__/activitys/ChallengeModeDungeon.test.ts
+++ b/src/__tests__/activitys/ChallengeModeDungeon.test.ts
@@ -133,7 +133,7 @@ test('Hard Depleted Challenge Mode', () => {
   expect(dungeon.upgradeLevel).toBe(0);
 });
 
-test('Peril Timed Challenge Mode', () => {
+test('Peril Does Not Extend Timer', () => {
   const startDate = new Date('2022-12-25T12:00:00');
   const endDate = getRelativeDate(startDate, 34 * 60); // Over the base time of 33 mins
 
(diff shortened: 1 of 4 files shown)


Nnoggie:
﻿Potential follow-up note:

Existing videos store `upgradeLevel` in metadata, so this PR affects new recordings unless old metadata is recalculated separately.

That could be useful as a separate feature later, for example:
- a manual "recalculate Mythic+ results" maintenance action
- an automatic metadata migration when app version changes and timer/result logic has changed
- a targeted repair pass for only Mythic+ videos whose stored timer rules are known to be stale

Leaving this out of scope here so this PR stays focused on fixing future result calculation.

aza547 (approved):
Thanks, looks sensible to me. I was expecting to see an update to how the +15s death penalty is applied too but actually I think that's built into the combat log arg. 

Not sure there is any point to the UTs that test Peril handling now (I don't think Peril can currently be applied at all?), but not doing any harm so whatever.

PR #818, opened 2026-05-02
macOS port (draft) — OSN backend, scene editor, signed CI build

yuripiratello:
## Status: draft / WIP — looking for direction before polishing

This is a **first-pass macOS port** of Warcraft Recorder, opened as a draft so you can see the shape of the work and steer the merge strategy before I clean it up. **Not asking for merge yet** — happy to split, rebase, drop scope, or rework anything.

The branch is currently powering a private beta on a fork (`v7.7.0-beta.1` released to a small group of M-series testers). Recording confirmed working end-to-end against a real Mythic+ run on macOS 26 Tahoe / M3 Pro.

## Scope at a glance

- **104 commits** across **94 files** (~14k insertions / 350 deletions).
- Core areas touched: `src/main/Recorder.ts`, `src/main/platform/recorder/*` (new `OsnBackend`), `src/parsing/CombatLogWatcher.ts`, `src/renderer/RecorderPreview.tsx`, build config, CI.
- New top-level files: `src/main/EditorService.ts`, `src/main/platform/recorder/{IRecorderBackend,NoobsBackend,OsnBackend,...}`, `vendor/node-window-rendering/`, `scripts/{relativise-osn-symlinks,sign-osn-binaries}.js`, `.github/workflows/release-mac-beta.yml`.

### 1. Platform interface extraction (Phase 0 — Windows-only refactor, zero behavior change)
- `Recorder.ts` no longer imports `noobs` directly. Recording goes through `IRecorderBackend`.
- Process polling, WoW path resolution, file reveal, ffmpeg path are all behind small adapter interfaces.
- Windows path (`NoobsBackend`) wraps the existing `noobs` calls 1:1.

### 2. macOS skeleton (Phase 1)
- Mac implementations of every adapter.
- `MacTccGate` surfaces Screen Recording / Microphone / Accessibility state to the renderer for a first-run permissions wizard.
- Window chrome, tray, menu mac branches.
- App boots on macOS, no recording yet.

### 3. macOS recording (Phase 2 — `OsnBackend`)
- `obs-studio-node` 0.26.17 wired through `IRecorderBackend`.
- `mac_screen_capture` window mode + Apple VideoToolbox encoders (`VT_H264`, `VT_HEVC`).
- Replay buffer + recording verified with a 25-min Mythic+ run, MKV → MP4 remux clean.
- Scene editor: OBS-drawn selection rectangle + drag/resize via `OBS_content_createSourcePreviewDisplay(sceneName)` — no OSN binding patches.
- Audio: SCK output + mic; per-app SCK partial.
- Combat log watcher gets a 2s polling fallback on Mac (FSEvents misses in-place log appends, breaking `CHALLENGE_MODE_START` detection).

### 4. Distribution (Phase 3)
- electron-builder mac targets (`dir`, `zip`, `dmg`).
- Custom `relativise-osn-symlinks.js` afterPack hook — fixes OSN's framework layout and replays symlinks that electron-builder dereferences during pack.
- `signIgnore` shrunk so electron-builder's deep-sign covers OSN/CEF natively (ran into Apple notarization rejecting custom afterSign re-signs).
- GitHub Actions workflow on `macos-14` runner — tag-triggered, signs + notarizes via App Store Connect API key, publishes draft GH Release.

## What needs to come out / change before merge

- [ ] `package.json > build.publish.owner` — currently `yuripiratello`, revert to `aza547`.
- [ ] `.github/workflows/release-mac-beta.yml` — drop the `if: github.repository == 'yuripiratello/wow-recorder'` guard once you decide whether to adopt the workflow.
- [ ] `release/app/package.json` version `7.7.0-beta.1` — your call on versioning. I bumped this on the fork only because the workflow needed a tag to fire.
- [ ] Decide if you want the `vendor/node-window-rendering/` prebuilt binary committed or built from source. I vendored it because the upstream npm package doesn't ship arm64 prebuilds and rebuilding from source is involved.
- [ ] `scripts/sign-osn-binaries.js` — currently unwired (was previously an `afterSign` hook, removed because it ran *after* notarization and invalidated Apple's stamp). Kept in `scripts/` for local dev signing. Probably safe to delete entirely.

## Things you'll want to know

- **Apple Silicon only** at this point. No Intel slice in the build. `obs-studio-node` ships an x64 tarball but I haven't verified it. Universal binary is feasible but I haven't done it.
- **No Game Capture on Mac** — macOS doesn't expose the equivalent of Windows' DirectX/Vulkan hook injection. Window Capture is the only mode.
- **No AV1** — Apple's media engine doesn't ship hardware AV1 encode.
- **Notarization** in CI requires App Store Connect API key secrets (`APPLE_API_KEY_BASE64`, `APPLE_API_KEY_ID`, `APPLE_API_ISSUER`). Without them, builds are signed but unnotarized — Tahoe Gatekeeper blocks the first launch.
- **Spec + plans** live under `docs/superpowers/specs/` and `docs/superpowers/plans/` if you want the design context. `docs/superpowers/notes/macos-port-beta1-status-2026-05-01.md` is the most current snapshot of what shipped vs. what's still pending.

## Testing done so far

- Manual smoke on macOS 26.3.1 Tahoe / M3 Pro:
  - App launches, permissions prompts fire, all three TCC grants accepted.
  - Window capture auto-attaches to WoW Retail.
  - Replay buffer + recording end-to-end through a Mythic+ run.
  - Encoder switch from x264 → VT_HEVC at runtime works.
- Unit + ESLint pass.
- Windows path **not regression-tested by me** — I don't have a Windows box. The Phase 0 refactor was designed to be a no-op on Windows but somebody with a Windows test rig should verify before merge.

## How I'd suggest reviewing this

If you're up for it — review by phase rather than commit-by-commit. The relevant docs:

1. `docs/superpowers/specs/2026-04-22-macos-port-design.md` — overall design.
2. `docs/superpowers/plans/2026-04-22-phase{0,1,2,2c}-*.md` — execution plans (mostly complete).
3. `docs/superpowers/plans/2026-04-30-mac-editor-service.md` — editor service, with a Resolution section explaining how it actually shipped vs. the original blocker hypothesis.
4. `docs/superpowers/notes/osn-macos-context.md` — context on why OSN init worked locally with binary patches but the real fix is signing + bundle context.

## Open questions for you

- Do you want this as one mega-PR, or split into a stack (Phase 0 first, then mac-specific phases)?
- Any concerns with `obs-studio-node` as a backend on Mac alongside `noobs` on Windows? It's a load-bearing dep and I'd want to know early if you'd prefer a different path.
- Anything in here you'd reject outright that I should drop now to make the PR smaller?

Happy to take this in any direction. Beer's on me if you ever want to chat about it.

— Yuri (yuripiratello)


aza547 (commented):
I had a quick skim of this today, here's some thoughts...

OSN instead of noobs. This seems to be the big one. There were lots of very good reasons for moving away from OSN and for the writing of noobs as a dedicated OBS library for WCR. I'm very unlikely to be convinced that reintroducing OSN is acceptable.

Some of this include...
- Constant disk wear from constant recordings
- Timing issues on video buffer cycling causing miscut videos 
- Relying on Streamlabs packaging of OSN
- Separate OBS process leaking on shutdown
- Different OBS versions shipped on Windows/Mac
- Not having control over this library for making fixes/enhancements as required

Mac signing. An apple developer account is in $100 a year which I need to be able to justify that there is enough worth in having Mac support. But beyond the real cost, the effort from me to set this up and maintain it is probably 10x that. 
- What does this require to maintain?
- Do I need an apple developer account?
- I don't own a Mac and probably won't be buying one to do this.
- I already have an EV code signing cert but I presume I can't use this because Apple?
- If this can happen entirely in the CI, then that'd be nice. 

A 14k line diff is simply not something I can feasibly review, consider, or maintain, or honestly even read without significant effort. Looking a bit closer maybe more than half of it is in your Claude plans. 

If you can remove those from the git tracking maybe I can make head or tail of this PR. I'm not going to have WCR become an AI driven project, or spend my limited time on reading AI output that I need to constantly assess if I can trust or not. If you want to use AI to help with PRs that's not a problem, but you can't merge this stuff or rely on it for communication.

In general the implementation looks OK beyond that, the splitting out into adapters seems like a good approach. A bit concerned about all the shell scripts that I would need to dig into a bit more to understand. 

The tests in this codebase are pretty useless on main, so you don't need to feel like you need unit tests to go along with this. I've mostly relied on the integration tests (the python stuff) and manual testing. Obviously unit tests are great to have but they are not a requirement here (and I'd rather solve that problem separately if at all).

You mentioned audio track limitations in discord, I presume you mean app capture. I think that's fine as a limitation for what it's worth. You also mentioned you were having trouble with the preview. Is that working now?

As I said earlier I've just had a skim, if this PR can become something I can feasibly review and consider I'd expect to have many more comments on the implementation. 

Also, would be great to hear about how testing of this is going. Did you find any users to help with that? Is it stable in your own testing? Presume you are gaming with your beta build now. Passing one of integration tests is one thing, or even recording individual dungeons but successfully recording entire raids etc is another.

yuripiratello:
Hey @aza547, thanks for taking the time to look through this even with the diff size — appreciate the honest read.

You're right on every point, let me work through them.

OSN. Fair. I knew the history of noobs going in but didn't think hard enough about why it existed. The objections you list (disk wear, output cycling, Streamlabs packaging dependency, separate process leaks, version skew, no upstream control) all apply just as much on Mac, and reintroducing OSN to the codebase isn't something I'd want to inflict on you to maintain. So that's a hard blocker for upstream.

The right long-term fix is a Mac equivalent of noobs — a small native binding around libobs that exposes the same API surface. That's a real chunk of work though (Objective-C++ around libobs, ScreenCaptureKit wiring, signal handling) and I think I can try to commit to it on a timeline that helps but it won't be too soon.

So a proposal, if you're open to it: let me close this PR and keep my fork running as a temporary Mac-only build while the proper noobs Mac port gets sorted out (whether by me chipping away at it, you, or someone else who shows up). I'd:
- Rename and rebrand it clearly as a fork (README that points back here as the source of truth and tells Windows users to use your release).
- Keep attribution + GPL-2.0 intact.
- Pull from upstream regularly so Windows users on my fork (if any wander in) don't fall behind.
- Send any non-Mac-specific fixes I find back as small focused PRs to your repo (no architectural noise).
- Take all the Mac support burden myself — Apple dev account, signing, notarization, CI, releases, user support.

That gets a working build to the Mac users who keep asking, doesn't add anything to your maintenance load, and the door stays open for a clean re-merge if the OSN dependency ever gets eliminated. Tell me if you'd rather I didn't do this and I'll just close it down.

Specific answers to what you asked:
- Apple developer account: yes, $100/yr. I cover it. Don't need anything from you.
- Mac to test on: not needed if it's my fork — I do the smoke testing.
- EV cert reuse: no, Apple has their own gated program (Developer ID), the EV cert can't sign for macOS.
- CI-driven: yes, fully — there's a .github/workflows/release-mac-beta.yml that runs on a macos-14 runner, signs from a .p12 in secrets, notarizes via App Store Connect API key, drops a draft release.
- 14k diff: a chunk of that is docs/superpowers/ — design notes I keep alongside the code while working. They shouldn't be in the shipping repo, and you're right to call it out. I'll scrub them out of git history on the fork before any further work.
- AI in PRs: understood. I use it as a tool but the accountability and communication needs to be mine. I'll keep that in mind going forward.
- Shell scripts: there are two (relativise-osn-symlinks.js, sign-osn-binaries.js). The second one I can drop entirely now — electron-builder's own deep-sign covers what it was doing. The first walks the OSN tree to fix relative symlinks the packager dereferences, and I'll add a clearer top-of-file comment explaining what's there and why.

Audio + preview status: yes, the limitation is per-app audio capture via SCK — I rolled it back because the create call hangs sync IPC during SCK init. Default desktop output + mic work. Preview is working end-to-end — drag, resize, OBS-drawn selection rectangle, transform handles. No OSN binding patches required for that, turned out the preview just had to be created via OBS_content_createSourcePreviewDisplay with a source name instead of the source-less createDisplay.

Testing reality check: honest answer — one Mythic+ run end-to-end (worked, MKV → MP4 cut clean). Not raids, not arena, not Solo Shuffle, not Battlegrounds, no other beta testers yet. The first beta tag was less than 48 hours ago. I'll be recruiting Mac-using WoW players in Discord this week and won't claim stability until I've got real-world hours behind it (full raid nights, M+ keys, arena, long uptime to shake out OSN leaks).

Let me know your thoughts on the fork plan. I'd rather get your blessing than just go ahead with it.

Issue #799, opened 2026-02-26
Audio devices get duplicated


aza547:
Title.

aza547:
[WarcraftRecorder-2026-02-23.log](https://github.com/user-attachments/files/25588610/WarcraftRecorder-2026-02-23.log)


[OBS-2026-02-22.zip](https://github.com/user-attachments/files/25588614/OBS-2026-02-22.zip)

aza547:
Hm I had a closer look at this. What I thought was a smoking gun in-fact is not. But a few realizations...

The obs_source_destroy method does not synchronously destroy a source.

```
/* defer source destroy */
os_task_queue_queue_task(obs->destruction_task_thread, (os_task_t)obs_source_destroy_defer, source);
```

So that explains the `Attempted to insert context with duplicate name` line here. Timing on deleting and recreating the source can mean that the earlier source is still alive. But WCR is still doing the right thing from there, and the initial source gets destroyed soon after (and it has already been removed from the scene anyway), and the dedup rename of the source is respected. 

```
console.info('[Recorder] Create audio source', src.id);
const name = noobs.CreateSource(src.id, src.type);
... use returned name not assumed name. 
```

So sadly the thing I thought was the smoking gun here is not it. Back to having no idea what is going on here and I'm not sure I can do anything without a reliable repro scenario.

aza547:
Here's some more logs that a re worth looking at:

[WarcraftRecorder-2026-05-21 (1).log](https://github.com/user-attachments/files/28140561/WarcraftRecorder-2026-05-21.1.log)
[OBS-2026-05-21 (2).log](https://github.com/user-attachments/files/28140562/OBS-2026-05-21.2.log)

Interesting in here is that it's not just a single extra device being created, we get up to "WCR Audio Source 1 3" so there most be something else going on here.

PR #823, opened 2026-05-22
Fix audio source cleanup after OBS source rename

Alkwer:
Related to #799

AI slop below but might be worth a shot 🙃.

OBS may rename a source when `CreateSource()` is called while another source with the requested name already exists, for example `WCR Audio Source 1` becoming `WCR Audio Source 1 2`.

`configureAudioSources()` already uses the OBS-returned name for setup and adding the source to the scene, but it stored the original configured id in `this.audioSources`. Later cleanup then tried to remove/delete the original id instead of the actual OBS source name.

This stores the OBS-returned source name in the active audio source list, so cleanup removes the same source that was added.

Validation:
- Verified locally with an Electron/OBS/noobs smoke test: old behavior removed the requested base name, fixed behavior removed the OBS-returned renamed source.

### DIFF
--- a/src/main/Recorder.ts
+++ b/src/main/Recorder.ts
@@ -788,7 +788,7 @@ export default class Recorder extends EventEmitter {
       }
 
       noobs.AddSourceToScene(name);
-      this.audioSources.push(src);
+      this.audioSources.push({ ...src, id: name });
     });
 
     const mics = this.audioSources.filter(


aza547:
This looks very promising. Agree the code looks broken for the removal case where OBS renames the device. Not sure how I didn't spot this. Thanks for the submission.

aza547:
fixes #799

PR #825, opened 2026-05-24
Fix unintended video player resize handles

### LINKED ISSUE #791: Uninteded player resizing
**Describe the bug**
In the intersection between the play button and the side bar, you get an icon as if you are supposed to be able to drag the sidebar. However, what it actually does is grab the player so that you can resize it horizontally. 

**To Reproduce**
1. Go to webapp
2. Click on and hold in the intersection between the play button and the side bar
3. Scroll down to drag to the right

**Expected behavior**

**Screenshots**

Image

Image

**Additional context**

Alkwer:
## Summary
Fixes #791.

This restricts the video player resize behavior to the intended bottom grip handle only.

## Root cause
`re-resizable` treats omitted `enable` directions as enabled in the bundled webapp logic.

The previous config used `enable={{ bottom: true }}`, which left side and corner handles active, including `left`, `right`, and `bottomLeft`. That allowed the player to be resized horizontally from the edges.

## Changes
- Explicitly disables all resize handles except `bottom`.
- Adds an `onResizeStart` guard so non-bottom resize directions are rejected.
- Adds focused renderer tests for the resize handle config and guard.

## Validation
- `./node_modules/.bin/jest.cmd --runInBand --testPathPattern src/__tests__/renderer/CategoryPage.test.tsx`
- `./node_modules/.bin/eslint.cmd src/renderer/CategoryPage.tsx src/__tests__/renderer/CategoryPage.test.tsx`
- `npm.cmd run build:renderer`

## Note
This also disables horizontal resizing from the right edge, since the current UI only exposes and persists height resizing through the bottom grip.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,13 @@
   "jest": {
     "transform": {
       "\\.(ts|tsx|js|jsx)$": "ts-jest"
-    }
+    },
+    "modulePaths": [
+      "<rootDir>/src"
+    ],
+    "modulePathIgnorePatterns": [
+      "<rootDir>/release/app"
+    ]
   },
   "dependencies": {
     "@electron/notarize": "^3.0.0",
--- a/src/__tests__/renderer/CategoryPage.test.tsx
+++ b/src/__tests__/renderer/CategoryPage.test.tsx
@@ -1,0 +1,264 @@
+/** @jest-environment jsdom */
+
+import React from 'react';
+import { render } from '@testing-library/react';
+import {
+  AppState,
+  Flavour,
+  Pages,
+  RendererVideo,
+  StorageFilter,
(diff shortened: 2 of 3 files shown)


aza547:
This looks like the right fix, but the website lives in a different (and private) codebase. I think for some unknown reason this is actually not an issue in the electron client. Although it probably makes sense to take this PR anyway and keep them in step.

aza547:
Also I don't really want the AI generated tests setting the testing precedent in this repo so I'm going to close this and make the minimal change myself. Thanks for the nudge though.

aza547:
https://github.com/aza547/wow-recorder/commit/86fae464189ac8e8365f5e27233931da78ac19ac

Issue #812, opened 2026-04-15
Kill Video Creator Fails (AV1?)


aza547:
https://discord.com/channels/1004860808737591326/1493756562324586649

PR #826, opened 2026-05-24
Make kill video rendering handle mixed POV sources

Alkwer:
Refs #812.

## Summary

This PR improves kill video rendering robustness for mixed POV source files:

- maps the selected POV audio stream as optional (`-map X:a?`) so FFmpeg does not fail when that source has no audio stream
- scales each POV into the target canvas with `force_original_aspect_ratio=decrease` before padding
- catches FFmpeg failures inside the kill-video queue worker so they are logged instead of becoming unhandled promise rejections

## Context

The original report does not include full FFmpeg stderr or the reporter's source files, so I cannot confirm this is the exact root cause for #812.

The generated command in the report exposed a couple of fragile FFmpeg assumptions around mixed POV sources and the selected audio stream. This PR addresses those failure modes without changing unrelated video processing behavior.

## Validation

- `npm.cmd run build:main`
- `git diff --check`

### DIFF
--- a/src/main/VideoProcessQueue.ts
+++ b/src/main/VideoProcessQueue.ts
@@ -478,6 +478,11 @@ export default class VideoProcessQueue {
         const item: UploadQueueItem = { path: videoPath };
         this.queueUpload(item);
       }
+    } catch (error) {
+      console.error(
+        '[VideoProcessQueue] Error processing kill video:',
+        String(error),
+      );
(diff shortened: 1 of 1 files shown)


aza547:
Sorry, I'm not convinced this is useful.

Alkwer:
Sure no problem :)

PR #828, opened 2026-05-28
Require NTFS for WoW log paths

### LINKED ISSUE #767: Detect WoW installed on exFAT filesystem
https://discord.com/channels/1004860808737591326/1234205219634548796/1436987012472246303

Alkwer:
Closes #767

## Summary
- Detect the filesystem format for configured WoW log paths on Windows.
- Mark active WoW log paths as invalid unless the detected filesystem is NTFS.
- Include the specific Retail / Classic / PTR / Era log path label in the localized config error.

## Notes
- This intentionally does not change `CombatLogWatcher` or switch file watching to polling.
- If filesystem detection fails, the path is treated as non-NTFS so the configuration is rejected.
- Added localized error text for English, German, Korean, and Simplified Chinese.

## Test Plan
- [x] `git diff --check`
- [x] Scoped eslint on changed config/localisation files
- [x] Scoped eslint on `src/main/util.ts` with only the pre-existing `no-explicit-any` rule disabled
- [x] `npm run build:main`
- [x] `npm run build:renderer`
- [ ] `npm test -- --runInBand` - currently fails before running tests due existing repo issues: missing `uiohook-napi`, unresolved `storage/CloudClient`, and stale activity constructor test calls


src/main/filesystem.ts
```diff
@@ -0,0 +1,66 @@
+import { execFile } from 'child_process';
```
aza547: I dont think this needs to be its own file. Can you move this to utils.ts?

src/utils/configUtils.ts
```diff
@@ -260,6 +261,21 @@ const getLocaleError = (phrase: Phrase) => {
...
+  if (!isExFat) {
+    return;
+  }
+
+  console.error(
+    '[Util] Unsupported exFAT filesystem for WoW log path',
+    logPath,
+  );
+  const error = getLocaleError(Phrase.UnsupportedWowFilesystem);
+  throw new Error(error);
```
aza547, line 278: This won't make it clear what log path is the issue. I think you need to pass that up in this error. i.e. it needs to say the retail / classic / whatever log path is the problem. a user might have 4 paths configured and one of the 4 may be wrong.

Alkwer:
1. Removed the separate src/main/filesystem.ts file, moved it to src/main/util.ts.
2. Made the exFAT config error identify the specific WoW log path - validateLogPathFilesystem now receives the relevant label, so the message can say things like Retail Log Path, Classic Log Path, Retail PTR Log Path, etc..
3. Updated the localized exFAT message to use a {logPath} placeholder - this keeps one shared translated message while still pointing to the exact problematic WoW version/path..

Think these are probably all the places that had to be changed.

src/main/util.ts
```diff
@@ -751,6 +753,66 @@ const checkDisk = async (dir: string, req: number) => {
...
+        }
+
+        console.info('[Util] Drive format', targetPath, driveFormat);
+        resolve(driveFormat);
+      },
+    );
+  });
+};
+
+const isExFatPath = async (targetPath: string): Promise<boolean> => {
```
aza547: I wonder if we should just be checking filesystem is NTFS rather than not exFat. I've never tested on anything other than NTFS. What do you think?

src/utils/configUtils.ts
```diff
@@ -260,6 +261,39 @@ const getLocaleError = (phrase: Phrase) => {
   return getLocalePhrase(lang, phrase);
 };
 
+const getFormattedLocaleError = (
```
aza547: I don't think you need this function if you just join the path with the translated message.

src/main/util.ts
Alkwer: In that case it might be better to do this indeed, to alert about nonNTFS filesystem

Alkwer:
Updated in the latest push:

- changed the filesystem check to require detected WoW log paths to be on NTFS rather than only rejecting exFAT
- kept the fail-open behavior when filesystem detection fails, to avoid false positives
- removed the formatting helper and now prefix the localized filesystem error with the specific log path label
- updated the PR title/body to match the NTFS behavior

src/main/util.ts
```diff
@@ -65,7 +65,9 @@ const setupApplicationLogging = () => {
   return path.dirname(logPath);
 };
 
-const { exec } = require('child_process');
+const { exec, execFile } = require('child_process');
```
aza547: Please put this at the top of the file.

src/main/util.ts
```diff
@@ -65,7 +65,9 @@ const setupApplicationLogging = () => {
...
+
+const DRIVE_FORMAT_COMMAND = '[System.IO.DriveInfo]::new($args[0]).DriveFormat';
```
aza547: This can probably just go inside the function. Doubt anyone else is going to reuse it given how specific it is.

src/main/util.ts
```diff
@@ -751,6 +753,66 @@ const checkDisk = async (dir: string, req: number) => {
...
+        'Bypass',
+        '-Command',
+        DRIVE_FORMAT_COMMAND,
+        root,
+      ],
+      { windowsHide: true },
+      (error, stdout, stderr) => {
+        if (error) {
+          console.warn('[Util] Failed to get drive format', targetPath);
+          console.warn(String(error));
```
aza547: Please don't do this - it may be hard to tell where the error log is from. You can put it into the previous line so that it's greppable (i.e. I can search the codebase for "Failed to get drive format" if I see this in logs.

src/main/util.ts
```diff
@@ -751,6 +753,66 @@ const checkDisk = async (dir: string, req: number) => {
...
+
+          if (stderr) {
+            console.warn(String(stderr).trim());
```
aza547: Ditto - ungreppable log.

src/main/util.ts
```diff
@@ -751,6 +753,66 @@ const checkDisk = async (dir: string, req: number) => {
...
+        resolve(driveFormat);
+      },
+    );
+  });
+};
+
+const isNtfsPath = async (targetPath: string): Promise<boolean | undefined> => {
+  const driveFormat = await getDriveFormat(targetPath);
+
+  return driveFormat ? driveFormat.toLowerCase() === 'ntfs' : undefined;
```
aza547: I think this can be better written to avoid the possibility of undefined

```
if (driveFormat && driveFormat.toLowerCase() === 'ntfs') {
 return true
}

return false
```

src/utils/configUtils.ts
```diff
@@ -260,6 +261,23 @@ const getLocaleError = (phrase: Phrase) => {
   return getLocalePhrase(lang, phrase);
 };
 
+const validateLogPathFilesystem = async (
+  logPath: string,
+  logPathLabel: Phrase,
+) => {
+  const isNtfs = await isNtfsPath(logPath);
+
+  if (isNtfs !== false) {
```
aza547: `if (isNtfs)` surely? You miss the undefined case here. But you should avoid that possibility also as per earlier comment

aza547 (changes requested):
Can you also show a screenshot of how this error will look in the UI? Suggest you could just flip the logic in your check to make it reproduce.

Alkwer:
Updated based on the latest review comments:

- Moved the `child_process` require to the top of `src/main/util.ts`.
- Inlined the PowerShell drive format command inside `getDriveFormat`.
- Merged the drive format failure logs into a single greppable `console.warn`.
- Changed `isNtfsPath` to return a strict `boolean` instead of `boolean | undefined`.
- Updated config validation to require an explicit NTFS result.
- Updated the PR description to reflect that filesystem detection failures are treated as non-NTFS.

Also screenshot:

image

aza547:
Hm latest looks broken?
```
[1] 15:32:47.428 > [Util] Failed to get drive format C:\Program Files\World of Warcraft\_retail_\Logs Error: Command failed: powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -Command [System.IO.DriveInfo]::new($args[0]).DriveFormat C:\
[1] At line:1 char:50
[1] + [System.IO.DriveInfo]::new($args[0]).DriveFormat C:\
[1] +                                                  ~~~
[1] Unexpected token 'C:\' in expression or statement.
[1]     + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
[1]     + FullyQualifiedErrorId : UnexpectedToken
[1]  
[1]  At line:1 char:50
[1] + [System.IO.DriveInfo]::new($args[0]).DriveFormat C:\
[1] +                                                  ~~~
[1] Unexpected token 'C:\' in expression or statement.
[1]     + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
[1]     + FullyQualifiedErrorId : UnexpectedToken
[1] 15:32:47.429 > [Util] Unsupported filesystem for WoW log path C:\Program Files\World of Warcraft\_retail_\Logs
[1] 15:32:47.430 > [Manager] Failed to configure base on startup Error: Retail Log Path: The configured WoW log path is not on an NTFS drive. Combat log updates can be unreliable on other filesystems. Move WoW to an NTFS drive and update this path.
```

Not sure I understand how the error in your screenshot is possible. Shouldn't it be "Retail log path:" not just "Retail:"?

Alkwer:
Both true, sorry for that - pushed another commit and the screenshot was wrongly taken, this one should be better

image

Alkwer:
- Fixed the PowerShell drive format check so the drive root is passed as an explicit script parameter.
- This avoids PowerShell parsing the drive root, e.g. C:\, as part of the command expression.
- The NTFS detection now correctly returns the filesystem format instead of failing with an “Unexpected token” error.

src/main/util.ts
```diff
@@ -751,6 +751,70 @@ const checkDisk = async (dir: string, req: number) => {
...
+  return new Promise((resolve) => {
+    execFile(
+      'powershell.exe',
+      [
+        '-NoProfile',
+        '-NonInteractive',
+        '-ExecutionPolicy',
+        'Bypass',
+        '-Command',
+        '& { param([string]$DriveRoot) [System.IO.DriveInfo]::new($DriveRoot).DriveFormat }',
```
aza547, line 776: Personally I would avoid writing something like this as the syntax is horrendous. You could just interpolate the Javascript string you are passing. 

```
`[System.IO.DriveInfo]::new("${DriveRoot"}).DriveFormat }`
```

aza547: But maybe that's personal preference...

aza547 (approved):

aza547:
FYI I made a few tweaks: https://github.com/aza547/wow-recorder/commit/8b1b92fd326e3544a2453ee50dae38d5848e2633 

Some tidy up but an important functional change to not fail the whole install if we fail to run the powershell command.

PR #829, opened 2026-05-28
[codex] Fix PTR recording detection

Alkwer:
## Summary

Fix PTR-only recording startup by including the PTR recording flags in process detection.

## Details

- Treat recordRetailPtr as a valid match for the Retail process state emitted by rust-ps.
- Treat recordClassicPtr as a valid match for the Classic process state emitted by rust-ps.
- Store the Classic PTR log handler in classicPtrLogHandler instead of overwriting classicLogHandler.

## Validation

- git show --check --stat --oneline HEAD

No tests were added; this PR intentionally keeps the patch to the minimal runtime fix.

### DIFF
--- a/src/main/Manager.ts
+++ b/src/main/Manager.ts
@@ -504,7 +504,7 @@ export default class Manager {
     }
 
     if (config.recordClassicPtr) {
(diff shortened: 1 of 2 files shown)


aza547:
LGTM

PR #831, opened 2026-05-31
Prevent native window menu on title bar

### LINKED ISSUE #682: Rightclick of title bar is a bit broken
Image

Alkwer:
## Summary

Prevent Windows from opening the native system window menu when right-clicking draggable title bar regions in the frameless Electron window.

Fixes #682

## Why

Electron treats `-webkit-app-region: drag` areas as non-client frame regions on Windows, so a right-click can trigger the native system menu. Handling `system-context-menu` in the main process suppresses that menu at the source.

## Validation

- `npx.cmd eslint src/main/main.ts`
- `npm.cmd run prestart`
- Manual GUI check on the actual local app: started the Electron app, right-clicked the title bar, and confirmed no native popup menu was created (`PopupMenusAfter: 0`).

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -198,6 +198,9 @@ const createWindow = async () => {
     },
   });
 
+  // Prevent Windows from opening the native window menu on draggable regions.
+  window.on('system-context-menu', (event) => event.preventDefault());
+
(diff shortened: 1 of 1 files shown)


aza547:
LGTM!

Issue #794, opened 2026-02-08
Webapp generate link doesn't work on iOS


aza547:
<img width="1284" height="2778" alt="Image" src="https://github.com/user-attachments/assets/17d80d9e-41dc-4b9c-809b-1d86f40ca383" />

https://discord.com/channels/1004860808737591326/1468786978815213619/1468786978815213619

aza547:
I just made a fix. Safari should now get a dialog:

<img width="1909" height="995" alt="Image" src="https://github.com/user-attachments/assets/f7ac9b4c-b791-4d43-83f1-12fedbe7b5ec" />

and the copy button is synchronous. Needs a real test though. I'll ask the user.

PR #832, opened 2026-06-01
[codex] Fix iOS share link clipboard copy

Alkwer:
## Summary
- Add a renderer clipboard helper that starts Clipboard API writes synchronously using a Promise-backed `ClipboardItem`.
- Update the share-link button flow to pass the pending link promise to the clipboard helper instead of awaiting it first.
- Return the generated shareable link from the Electron `getShareableLink` IPC handler while preserving the existing Electron clipboard write fallback.

## Root Cause
On iOS Safari/WebKit, `navigator.clipboard.writeText()` can be rejected when it runs after awaited async work because the original user activation has been lost.

## Related Issue
- https://github.com/aza547/wow-recorder/issues/794

## Validation
- `eslint` on the changed production files
- `npm run build:renderer`

## Notes
- No automated regression test is included in this PR.
- Manual verification on iOS Safari is still needed before opening a PR to the upstream repository.

### DIFF
--- a/src/renderer/VideoPlayer.tsx
+++ b/src/renderer/VideoPlayer.tsx
@@ -62,6 +62,7 @@ import CloudOffIcon from '@mui/icons-material/CloudOff';
 import Separator from './components/Separator/Separator';
 import { toast } from './components/Toast/useToast';
 import { Phrase } from 'localisation/phrases';
+import { copyTextPromiseToClipboard } from './clipboard';
 
 interface IProps {
   videos: RendererVideo[];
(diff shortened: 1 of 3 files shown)


Alkwer:
Some idea, but, as mentioned in description of PR, I don't have any iOS device to test this fix

aza547:
This is another one that is a webapp bug not a desktop client bug, so the fix belongs elsewhere. Although the fix may be the same, it's hard to reason about a PR like this that is untested.

aza547:
I've fixed this separately.

PR #835, opened 2026-06-06
[codex] Handle delayed OBS stop recovery

Alkwer:
## Summary

This makes OBS stop recovery less brittle when OBS emits `stop` quickly but takes longer to fully deactivate/finalize the output.

In the observed failure mode, WCR timed out during the normal stop path, called `ForceStopRecording()`, then only waited 3 seconds for recovery. OBS can still be finalizing at that point, so WCR may discard a recording even though a later `deactivate` signal would have completed the stop.

## Changes

- Create the stop wait promise before calling `StopRecording()`.
- Increase the forced-stop recovery wait from 3 seconds to 60 seconds.
- Keep the change scoped to `Recorder.ts`.

## Validation

- `npm run build`
- `npx eslint src/main/Recorder.ts --no-warn-ignored`
- Real dev app smoke test with Retail WoW running: WCR detected WoW, started the buffer, recorded a simulated Retail Solo Shuffle, stopped OBS cleanly, processed the MP4, wrote metadata, and did not queue cloud upload.

### DIFF
--- a/src/main/Recorder.ts
+++ b/src/main/Recorder.ts
@@ -1000,8 +1000,8 @@ export default class Recorder extends EventEmitter {
     }
 
     this.stopQueue.empty();
-    noobs.StopRecording();
     const wrote = this.stopQueue.shift();
+    noobs.StopRecording();
 
     try {
       await Promise.race([wrote, getPromiseBomb(60, 'Failed to stop')]);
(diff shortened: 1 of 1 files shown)


Alkwer:
Some idea for this issue: https://discord.com/channels/1004860808737591326/1362788447546511451/threads/1512847278417903916

aza547:
Why do you think it is acceptable to wait more than 3 seconds when calling ForceStopRecording?

aza547:
I would like to spent some time looking at this sort of stability issue but I don't think this is correct. Force stopping should be basically instant (in-fact, 3s is already really very long to wait here). It is an instruction to immediately abandon the recording.

Alkwer:
Yeah, you're right, extending ForceStopRecording wasn't a good idea, sorry.

Anyway, I see that the issue is solved.

PR #837, opened 2026-06-06
[codex] Add delete after cloud upload option

### LINKED ISSUE #496: Toggle option to never show disk videos and delete them as soon as uploaded
Title

Alkwer:
﻿## Summary
- Add a `Delete After Upload` cloud setting so users can opt into removing local copies after successful cloud upload.
- Hide local videos while they are staged for upload, then mark them for local deletion only after cloud metadata is committed.
- Recover hidden staging videos on cloud startup/config failures and treat duplicate cloud metadata (`409`) as an idempotent upload success.

## Validation
- `npm.cmd run build` passed.
- Local app smoke test passed: renderer, main process, and OBS reached ready state.
- `git diff --check` passed.
- `src/__tests__` intentionally unchanged.

## Notes
- Draft PR because this keeps the feature diff minimal. Full `npm test`, lint, and standalone `tsc --noEmit` require existing repo/test maintenance changes that are intentionally not included here.

Fixes #496

image


aza547:
This feature needs careful consideration to make sure the user experience is acceptable. I probably can't accept an AI PR for this I'm afraid.

> Hide local videos while they are staged for upload, then mark them for local deletion only after cloud metadata is committed.

But the whole point of WCR is that videos are immediately available after you finish a pull/arena/dungeon. This means you won't be able to see your M+ run for several minutes if you have slow upload.

aza547:
Happy to discuss further if you like.

PR #833, opened 2026-06-01
[codex] Add chat POV timestamp links

Alkwer:
## Summary

Closes #808.

Video chat supports POV links alongside the existing timestamp links:

- `0:12` seeks the current playback.
- `@Player 0:12` or `@Player-Realm 0:12` selects that POV and seeks to the timestamp.
- `@Player` selects that POV while preserving the playback position.

Typing `@` opens suggestions for the current activity, with mouse selection, arrow keys, Enter/Tab completion, and Escape dismissal. Suggestions and actionable chat links show class icons when metadata is available. Unknown or ambiguous POV references remain plain text. Matching prefers disk copies, and completion includes the realm only when the short player name is ambiguous.

## Implementation

- Keep parsing and player matching in `videoChatLinks.ts`.
- Route chat actions through `CategoryPage` and the existing `VideoPlayer` seek and persistent-progress behavior.
- Restore textarea focus and caret position after mention completion.
- Integrate `main` at `0566da15`, preserving React Table 9, the current dialog handling, and filtered playback state. Resolve the `CategoryPage` import and chat-POV conflicts.
- Keep the issue entry under `Unreleased` in `CHANGELOG.md`.

## Validation

Windows, Node 24.19.0, npm 11.17.0. Existing installed dependencies were reused; the root package manifest and lockfile match the merged tree.

- `npm run build` passed for both main and renderer production bundles.
- `npm exec -- eslint src/renderer/CategoryPage.tsx src/renderer/VideoChat.tsx src/renderer/videoChatLinks.ts src/renderer/components/TextArea/textarea.tsx` passed with no errors and one existing `VideoChat` exhaustive-deps warning.
- `git diff --cached --check 0566da15` passed for the PR changes relative to current `main`.
- Independent review confirmed that the merge preserves the existing POV handlers and the newer table/dialog integration.
- `node integration.cjs` (temporary local harness) passed 8 React/jsdom checks using the merged `CategoryPage` and `VideoChat`: named timestamp switch/seek, standalone mention position preservation, plain timestamp seek, leaving multi-player mode paused on one POV, keyboard and mouse completion, caret/focus restoration and suffix preservation, ignoring old mentions away from the caret, and Escape dismissal through keyup.

## Existing baseline and limits

[...]

PR #838, opened 2026-06-07
[codex] Add clip source navigation

Alkwer:
## Summary

- Persist clip source metadata, including the source category, source video name, and clip offset.
- Add a source button to the Clips table that jumps back to the original recording at the captured offset.
- Disable the source action when the original recording is unavailable, while preserving a legacy hash/start-time fallback for older clips.
- Add localized tooltip text for the new source action.

## Validation

- `npm.cmd run build`
- Targeted ESLint for the changed renderer/localisation/main files that are lint-clean. `src/main/util.ts` still has existing `no-explicit-any` lint failures outside this change when linted directly.
- Packaged app smoke test: verified the Clips table source button navigates from a clip to the original recording at the saved offset.

### DIFF
--- a/src/main/VideoProcessQueue.ts
+++ b/src/main/VideoProcessQueue.ts
@@ -37,13 +37,19 @@ const isDebug = devMode || process.env.DEBUG_PROD === 'true';
 // Use the dynamically linked ffmpeg.exe we package with OBS in noobs. This
 // allows us to avoid including a static ffmpeg.exe which is an extra 60MB.
 const ffmpegPathRel = 'node_modules/noobs/dist/bin/ffmpeg.exe';
+const ffprobePathRel = 'node_modules/noobs/dist/bin/ffprobe.exe';
 
 let ffmpegPathAbs = devMode
   ? path.resolve(__dirname, '../../release/app/', ffmpegPathRel)
   : path.resolve(__dirname, '../../', ffmpegPathRel);
+let ffprobePathAbs = devMode
+  ? path.resolve(__dirname, '../../release/app/', ffprobePathRel)
+  : path.resolve(__dirname, '../../', ffprobePathRel);
 
 ffmpegPathAbs = fixPathWhenPackaged(ffmpegPathAbs);
+ffprobePathAbs = fixPathWhenPackaged(ffprobePathAbs);
 ffmpeg.setFfmpegPath(ffmpegPathAbs);
+ffmpeg.setFfprobePath(ffprobePathAbs);
 
 /**
  * A queue for cutting videos to size.
@@ -264,6 +270,7 @@ export default class VideoProcessQueue {
       // covers the cases where we're cutting a section off the end of
       // the video due to a timeout.
       const videoPath = await this.cutVideo(data, outputDir);
+      await VideoProcessQueue.updateClipTimingMetadata(data, videoPath);
 
       // Add the size of the newly cut video. We can't do this earlier
       // as the size will change when we cut/remux.
@@ -754,6 +761,73 @@ export default class VideoProcessQueue {
     await VideoProcessQueue.ffmpegWrapper(fn, 'Video cut');
     console.timeEnd('[VideoProcessQueue] Video cut took:');
     return outputPath;
+  }
+
+  private static async getVideoDurationSeconds(
+    videoPath: string,
+  ): Promise<number | undefined> {
+    return new Promise((resolve) => {
+      ffmpeg.ffprobe(videoPath, (error, data) => {
+        if (error) {
+          console.warn(
+            '[VideoProcessQueue] Failed to probe video duration',
+            videoPath,
+            String(error),
+          );
+          resolve(undefined);
+          return;
+        }
+
+        const duration = data.format.duration;
+        resolve(
+          Number.isFinite(duration) && duration > 0 ? duration : undefined,
+        );
+      });
+    });
+  }
+
+  private static async updateClipTimingMetadata(
+    data: VideoQueueItem,
+    videoPath: string,
+  ) {
+    if (!data.clip) return;
+
+    const actualDuration =
+      await VideoProcessQueue.getVideoDurationSeconds(videoPath);
+    if (!actualDuration) return;
+
+    data.metadata.duration = actualDuration;
+
+    const parentOffset = data.metadata.parentOffset;
+    if (typeof parentOffset !== 'number' || !Number.isFinite(parentOffset)) {
+      return;
+    }
+
+    const requestedDuration = data.duration;
+    if (!Number.isFinite(requestedDuration) || requestedDuration <= 0) {
+      return;
+    }
+
+    // Copy-based clipping can only start on a keyframe, so ffmpeg may include
+    // frames from before the requested offset. Match source navigation to the
+    // clip that was actually written, not just the requested slider position.
+    const keyframeLeadIn = Math.min(
+      parentOffset,
+      Math.max(0, actualDuration - requestedDuration),
+    );
+
+    if (keyframeLeadIn <= 0) return;
+
+    const adjustedParentOffset = parentOffset - keyframeLeadIn;
+    data.metadata.parentOffset = Math.round(adjustedParentOffset * 1000) / 1000;
+
+    console.info('[VideoProcessQueue] Adjusted clip source offset:', {
+      requestedOffset: parentOffset,
+      adjustedOffset: data.metadata.parentOffset,
+      requestedDuration,
+      actualDuration,
+      keyframeLeadIn,
+    });
   }
 
   /**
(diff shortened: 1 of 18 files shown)


Alkwer:
image
image
image

src/main/types.ts
```diff
@@ -191,9 +191,11 @@ type VideoQueueItem = {
 type Metadata = {
   category: VideoCategory;
   parentCategory?: VideoCategory; // present if it's a clip
+  parentVideoName?: string; // present if a clip knows its source video
```
aza547, line 195: I wonder if we can also work this out retroactively for already clipped videos but just stripping the  "- Clipped At ..." part from the file name (or videoName property for cloud videos).

src/main/util.ts
```diff
@@ -626,13 +626,22 @@ const getPromiseBomb = (fuse: number, reason: string) => {
...
+) => {
   const final = initial;
   final.duration = duration;
+  // Preserve the source category before this metadata is converted into a clip.
   final.parentCategory = initial.category;
+  final.parentVideoName = parentVideoName;
   final.category = VideoCategory.Clips;
   final.protected = true;
   final.clippedAt = date.getTime();
+  final.clipOffset = offset;
```
aza547: reckon this should share the parent prefix - the rest of the metadata in the clips is sortof bogus because it just copies from the parent on clipping. would be nice to make it obvious that these are trusted fields in clips.

src/renderer/components/Tables/Cells.tsx
```diff
@@ -103,6 +106,8 @@ export const populateDetailsCell = (
   language: Language,
   cloudStatus: CloudStatus,
   setVideoState: Dispatch<SetStateAction<RendererVideo[]>>,
+  allVideos?: RendererVideo[],
```
aza547: Should this not be videoState?

src/renderer/components/Tables/Cells.tsx
```diff
@@ -103,6 +106,8 @@ export const populateDetailsCell = (
...
+  goToClipParent?: (clip: RendererVideo) => void,
```
aza547: Curious why this needs passed in. Was maybe expecting to see setAppState here.

src/renderer/components/Tables/Cells.tsx
```diff
@@ -214,10 +219,47 @@ export const populateDetailsCell = (
...
+    return (
+      <Tooltip content={tooltip}>
+        <div>
+          <Button
+            variant="ghost"
+            size="xs"
+            onClick={goToSource}
+            disabled={disabled}
+          >
+            <Search size={18} />
```
aza547: Reckon there is a better icon for this. https://lucide.dev/icons/external-link maybe?

src/renderer/components/Tables/VideoSelectionTable.tsx
```diff
@@ -281,9 +281,19 @@ const VideoSelectionTable = (props: IProps) => {
    * Render an individual row of the table.
    */
   const renderRow = (row: Row<RendererVideo>, sortedIndex: number) => {
+    const selectedVideo = appState.selectedVideos[0];
```
aza547: dont understand why this exists

src/renderer/CategoryPage.tsx
```diff
@@ -108,8 +110,40 @@ const CategoryPage = (props: IProps) => {
     return correlatedState.filter(queryFilter);
   }, [correlatedState, dateRangeFilter, videoFilterTags, language]);
 
+  const goToClipParent = useCallback(
```
aza547: I enabled React compiler for the project at the weekend, so you don't need to useCallback / useMemo. React compiler will handle that at build time.

aza547 (changes requested):
A bit short on time but a few comments. I think this a really neat feature though, keen to have it.

Alkwer:
Addressed the review comments:
- renamed the clip offset metadata to parentOffset
- grouped trusted clip source fields under the parent* naming
- added legacy source lookup by stripping the " - Clipped at ..." suffix
- changed the icon to ExternalLink
- simplified the table cell API so it does not receive full videoState
- removed the new useCallback usage
- added a comment explaining row highlighting after programmatic navigation

aza547:
I tried to have a play around with this but the app seems pretty broken - looks like it's not setting the src property correctly on the video player? (I can't play any videos)

Also clicking "go to source" doesn't change the side menu (although from my limited testing did correctly seek to the parent video). 

One thing I found a bit weird was that if I seek to an older parent it doesn't scroll the video selection table to that so it's a bit hard to tell what is selected. That table is using tanstack table which is a really powerful library - I it's possible to adjust the scroll on seek so it's onscreen. 

Lesser comment - but should the "seek to source" button be on the far right like the "create kill video" button is in the raid category? The leftmost column is looking a bit busy imo and there is precedent for going to the far right.

aza547:
Realized the video player not working might be the update to ReactPlayer I took - you probably just want to rebase on main. The component's API changed quite a lot and I didn't do an `npm install` on this branch.

Alkwer:
- Rebased the branch onto the latest `main` / 7.8.1
- Fixed source-recording navigation opening at `t=0` by adding an explicit initial seek for ReactPlayer 3 / React 19.
- Fixed source timestamps for newly created clips by probing the actual written clip duration after ffmpeg cutting and adjusting for keyframe lead-in.
- Updated generated/legacy clip handling so source navigation is only enabled when there is trusted source metadata, while still supporting legacy `- Clipped at ...` clips.
- Moved the source-recording action into its own Source column on the right side of the clips table.
- Improved source navigation behavior so it clears conflicting filters, switches to the source category, selects the source recording, and scrolls/pages to the row.
- Synced the menu highlighted state when navigation changes categories programmatically.

src/main/VideoProcessQueue.ts
```diff
@@ -264,6 +270,7 @@ export default class VideoProcessQueue {
       // covers the cases where we're cutting a section off the end of
       // the video due to a timeout.
       const videoPath = await this.cutVideo(data, outputDir);
+      await VideoProcessQueue.updateClipTimingMetadata(data, videoPath);
```
aza547, line 273: I don't understand  any of this. Why do you need `ffprobe`? Surely we have all the info to hand already? What is it giving you? The video duration? We know that. It's on `data`. 

```
type VideoQueueItem = {
  name: string; // Can be an OBS timestamp if recording or more complicated if clipping.
  source: string; // Can be either a path or a URL.
  suffix: string; // Typically details of the recording, but can also be a "clipped at ..." description.
  offset: number;
  duration: number;
  clip: boolean;
  metadata: Metadata;
};
```

There is loads of logic here for no reason as far as I can tell. Same comment for keyframes. WCR puts keyframes deliberately every second (which is more frequent than OBS defaults). You will only ever be 0.5s inaccurate worst case if you omit this logic. That's totally fine and I'd much rather that than this. 

Lastly, the metadata should be written in one shot. Not this write then rewrite approach.

src/renderer/components/Menu/Menu.tsx
```diff
@@ -45,11 +45,10 @@ export const Menu = ({
...
   };
 
-  // This is a bit of a janky way to handle it I guess, but it accounts for
-  // the fact that the two sections of content and settings are two different menus
-  // so when one gets selected, the other needs to be cleared
+  // Keep the highlighted item in sync when navigation is driven from outside
+  // the menu, such as jumping from a clip to its source recording.
   React.useEffect(() => {
-    if (!initialValue) setCurrentValue(undefined);
+    setCurrentValue(initialValue || undefined);
```
aza547, line 51: this is logically different? why change it?

aza547 (changes requested):
Sorry, I think I've been a bit grumpy here. I'm finding the code in a few of these PRs now is pretty poor and is missing key considerations. I assume it's AI. It feels pretty bad to spend my time reviewing PRs like this when I could have probably implemented this myself in the time I've spent here now. Happy to be corrected if I'm misunderstanding this...? But the ffprobe / keyframes stuff going on here just seems like nonsense?

aza547:
You can probably tell my feelings towards AI from the above. But to be clear, I don't object to AI usage. I'm using it myself. However I do object to spending my weekends reviewing AI code that hasn't had due care applied to it by a human to check it is sensible, useful, well tested. Ultimately accepting PRs into WCR is a trust based thing, and I certainly don't trust AI on its own, so I need to trust the human behind the PR is thinking about what they are submitting, and I don't see it here.

aza547:
As I said earlier I do really like the idea here, so I would like to see this PR develop to a point where I can merge it. If you have questions about how things work feel free to reach out to me, the keyframe stuff for example isn't obvious without some digging. But is a very important part of how WCR works.

Alkwer:
That’s fair. I generally try to test each iteration end-to-end whenever I have access to the necessary test environment, but I can’t honestly say how thoroughly I tested this particular revision a month later. I’m trying to improve my process with every PR.
I’ll take a step back on both this PR and #833, revisit them from first principles, simplify them where necessary, and test them properly again. Of course, I don’t want to create unnecessary review work for you.

aza547:
Spent some time on this today myself. Expecting my PR here to supersede this: https://github.com/aza547/wow-recorder/pull/872.

PR #839, opened 2026-06-07
Fix date range picker close interactions

### LINKED ISSUE #629: Date range filter buttons can be off-screen
Image

Alkwer:
Closes #629

## Summary
- Adds a local downstream fix for the bundled date range picker close behavior.
- Allows the picker to close when pressing `Escape` or clicking the calendar button again while the picker is open.

## Why
The bundled `react-tailwindcss-datepicker` version does not include the upstream fix from onesine/react-tailwindcss-datepicker#318 yet, so this keeps the workaround local to the app's date filter wrapper.

## Validation
- Ran `git diff --check`.
- Ran `node node_modules\eslint\bin\eslint.js src\renderer\DateRangePicker.tsx`.
- Ran `npm run build`.
- Verified in the real Electron app that `Escape` closes the datepicker.
- Verified in the real Electron app that clicking the calendar button again closes the datepicker.

### DIFF
--- a/src/renderer/DateRangePicker.tsx
+++ b/src/renderer/DateRangePicker.tsx
@@ -1,7 +1,14 @@
 import { getLocalePhrase } from 'localisation/translations';
 import { Phrase, Language } from 'localisation/phrases';
 import { AppState } from 'main/types';
-import { Dispatch, SetStateAction, useCallback } from 'react';
+import {
+  Dispatch,
+  MouseEvent,
+  SetStateAction,
+  useCallback,
+  useEffect,
+  useRef,
+} from 'react';
 import Datepicker, { DateValueType } from 'react-tailwindcss-datepicker';
 
 interface IProps {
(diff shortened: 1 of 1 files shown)


aza547:
Hm I'm not sure if a CSS fix is the right thing here. It doesn't look very good - I'm unsure why its so far up the screen in your screenshot given there is lots of space below still. 

Keep in mind this is a very minor issue and simply maximizing the window is likely a good enough workaround to not bother fixing this.

Note the upstream fix which enables closing on escape and re-selecting the button. https://github.com/onesine/react-tailwindcss-datepicker/pull/318. I think you could probably also implement that downstream if you are keen to fix.

Alkwer:
- Reworked PR #839 from the previous CSS-only workaround to a local downstream fix in DateRangePicker.tsx.

- Added local close handling for the bundled react-tailwindcss-datepicker, since the upstream fix is not available in the released package version yet.

- The date picker now closes when the user presses Escape.

- The date picker now closes when the user clicks the calendar button again while it is already open.

- Removed the previous CSS workaround from the final PR diff.

- Updated the PR title and description to match the new behavior-focused fix.

aza547:
Hm the code for this is kind of hideous I'm afraid, honestly I'd rather have the bug 🙈

aza547:
if this can't be as simple as an on-key-press handler I suggest we just leave it

Alkwer:
I don't think there is a simple solution like the one you mentioned - let's close the topic then :thumbsup:

PR #842, opened 2026-06-13
[RFC] OSX Support

Jackatb:
This PR is adding support for OSX with wow-recorder. I have been using these
patches on my machine for a while, but finally got around to tidying them up
and submitting them upstream.

This requires a few little "hacks" with some patching of libs to make it work on OSX.
Most of these are nicely wrapped behind helpers.

Sorry they're a bit of a mess but hopefully they'll be helpful for you.

Thanks Jack

### DIFF
--- a/native/macos/README.md
+++ b/native/macos/README.md
@@ -1,0 +1,80 @@
+# macOS OBS Runtime
+
+Warcraft Recorder uses a patched OBS `obs-ffmpeg` plugin on macOS so replay
+buffer saves can use the same `StartRecording(offset)` flow as Windows.
+
+Generated OBS checkouts, build output, and staged binary artifacts are ignored
+by git:
+
+```text
+native/macos/build/
+native/macos/artifacts/
+```
+
+The source of truth is the versioned patch set in `native/macos/patches/` plus
+the build scripts in `scripts/macos/`.
+
+## Rebuild Patched OBS Artifacts
+
+To clone OBS `30.2.3`, apply Warcraft Recorder's patches, build the required
+plugin/helper targets, and stage the outputs:
+
+```bash
+scripts/macos/build-obs-ffmpeg.sh
+```
+
+The staged layout is:
+
+```text
+native/macos/artifacts/obs-30.2.3-arm64/
+  app/Contents/MacOS/obs-ffmpeg-mux
+  obs/OBS.app/Contents/WCRPlugIns/obs-ffmpeg.plugin
+```
+
+To stage already-built local OBS outputs:
+
+```bash
+scripts/macos/stage-obs-artifacts.sh \
+  /path/to/obs-ffmpeg.plugin \
+  /path/to/obs-ffmpeg-mux
+```
+
+The plugin must be built against the same OBS/FFmpeg ABI as the bundled
+`OBS.app`. The default packaging path expects OBS `30.2.3`.
+
+## Package Runtime
+
+During Electron packaging, `scripts/macos/after-pack.js` copies `OBS.app` into
+the generated app bundle, overlays the patched plugin, copies
+`obs-ffmpeg-mux` beside the Electron executable, and copies the rebuilt
+`noobs.node` addon into `noobs/dist`.
+
+Defaults:
+
+```text
+OBS.app source: /Applications/OBS.app
+artifacts: native/macos/artifacts/obs-30.2.3-<arch>/
+```
+
+Override paths when needed:
+
+```bash
+WCR_MACOS_OBS_APP_PATH=/path/to/OBS.app
+WCR_MACOS_OBS_ARTIFACT_DIR=/path/to/native/artifacts/obs-30.2.3-arm64
+```
+
+## Verify
+
+After packaging, run:
+
+```bash
+npm run smoke:macos-app
+npm run smoke:macos-obs
+```
+
+Pass a specific app bundle with:
+
+```bash
+npm run smoke:macos-app -- --app /path/to/WarcraftRecorder.app
+npm run smoke:macos-obs -- --app /path/to/WarcraftRecorder.app
+```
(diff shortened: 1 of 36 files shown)


aza547:
Very cool you have got this working. 

Need to have a think about what is possible long term in the way of this feature. Realistically relying on patch files and manual instructions won't be something I'd be comfortable merging or being the basis for official MacOS support in WCR. 

Unfortunately I'm very busy with other feature work just now, and I won't be able to give this feature the consideration it needs for a while. Also I don't think I can take this on without actually having a MacOS machine to test on which I currently don't own. That said, I have been considering buying one.

native/macos/patches/obs-30.2.3-replay-buffer-convert.patch
```diff
@@ -0,0 +1,1048 @@
...
+@@ -183,23 +191,17 @@ static void add_video_encoder_params(struct ffmpeg_muxer *stream,
+ 		spc = AVCOL_SPC_BT2020_NCL;
+ 	}
+
+-	const enum AVColorRange range = (info->range == VIDEO_RANGE_FULL)
+-						? AVCOL_RANGE_JPEG
+-						: AVCOL_RANGE_MPEG;
+-	const enum AVChromaLocation chroma_location = determine_chroma_location(
+-		obs_to_ffmpeg_video_format(info->format), spc);
++	const enum AVColorRange range = (info->range == VIDEO_RANGE_FULL) ? AVCOL_RANGE_JPEG : AVCOL_RANGE_MPEG;
```
aza547, line 87: These changes look superfluous to me? Is this just a copy of my own OBS fork or is there anything more going on here?

native/macos/README.md
```diff
@@ -0,0 +1,80 @@
...
+```text
+OBS.app source: /Applications/OBS.app
+artifacts: native/macos/artifacts/obs-30.2.3-<arch>/
+```
+
+Override paths when needed:
+
+```bash
+WCR_MACOS_OBS_APP_PATH=/path/to/OBS.app
+WCR_MACOS_OBS_ARTIFACT_DIR=/path/to/native/artifacts/obs-30.2.3-arm64
```
aza547, line 63: Is this the same version of OBS noobs uses currently? (I can't remember)

src/main/Manager.ts
```diff
@@ -165,6 +165,35 @@ export default class Manager {
     this.reconfiguring = false;
   }
 
+  /**
+   * Release long-lived process resources before the app exits.
+   */
+  public shutdown() {
+    console.info('[Manager] Shutting down');
```
aza547, line 172: Why does this matter?

src/main/Manager.ts
```diff
@@ -566,13 +595,6 @@ export default class Manager {
       }
     });
 
-    // Test listener, to enable the test button to start a test.
-    ipcMain.on('test', (_event, args) => {
```
aza547, line 570: This seems suspicious

src/main/obsPlatform.ts
```diff
@@ -0,0 +1,82 @@
...
+      return 'coreaudio_input_capture';
+    default:
+      return type;
+  }
+};
+
+const isAppleH264Encoder = (encoder: string) =>
+  MAC_VIDEO_TOOLBOX_H264_ENCODERS.includes(encoder as ESupportedEncoders);
+
+const getVideoToolboxBitrate = (quality: string) => {
```
aza547, line 53: Bitrate is not a good rate control for WCR. It does not consider various factors like screen resolution or fps. 30k might be ultra for a 1920x1080 screen but certainly isn't for an ultrawide screen. It's why WCR users CQP.

src/main/Recorder.ts
```diff
@@ -979,6 +1037,14 @@ export default class Recorder extends EventEmitter {
 
     // The native code expects an integer.
     const rounded = Math.round(offset);
+
+    if (!this.replayBufferConvertSupported) {
```
aza547, line 1041: Don't understand how this could happen assuming the build process is followed correctly?

src/main/Recorder.ts
```diff
@@ -1472,12 +1576,15 @@ export default class Recorder extends EventEmitter {
...
+    const normalizedName = item.name.toLowerCase();
+
     return (
       item.name.startsWith('[Wow.exe]: ') ||
       item.name.startsWith('[WowT.exe]: ') ||
       item.name.startsWith('[WowB.exe]: ') ||
       item.name.startsWith('[WowClassic.exe]: ') ||
-      item.name.startsWith('[WowClassicT.exe]: ')
+      item.name.startsWith('[WowClassicT.exe]: ') ||
+      normalizedName.includes('world of warcraft')
```
aza547: Won't this match on any browser window with "World of Warcraft" in the title?

src/utils/Poller.ts
```diff
@@ -31,6 +32,27 @@ export default class Poller extends EventEmitter {
...
+  private scanning = false;
+
+  /**
+   * True while process scan results are allowed to emit state changes.
+   */
+  private running = false;
+
+  /**
+   * We do not need real-time process detection. A few seconds keeps the app
+   * responsive without continuously waking the system.
```
aza547: err, what? 5s seems pretty continuous!

src/utils/Poller.ts
```diff
@@ -80,24 +108,82 @@ export default class Poller extends EventEmitter {
...
+    if (this.scanning) {
+      return;
+    }
+
+    this.scanning = true;
+
+    try {
+      const stdout = await new Promise<string>((resolve, reject) => {
+        execFile(
+          'ps',
```
aza547: Wouldn't pgrep be better? assuming it's available on macos which I think it is

aza547:
I had a really quick skim and made a few comments which jumped out at me. In general lots of it looks sensible. But to be clear I still think the only way this can be viable is if noobs is updated to have macos support.

PR #845, opened 2026-06-20
added tbc anniversary support

Mirix5:
I wanted to add support for the burning crusade anniversary servers.


aza547:
I don't really understand this PR I'm afraid, this looks like lots of scaffolding around the TBC log hander which does nothing other than extend the classic log handler and add some IDs that already exist? 

Why does this need a new log handler and settings? Why isn't this PR a one liner to add "anniversary" to the list of classic flavours that we accept (without log path validation disabled). See `validClassicFlavours` in `ConfigUtils.validateBaseConfig()`. 

The boss IDs in `constants.ts` is probably a bit misleading from previous iterations of the app that I never removed, but they are not actually required. The boss ID doesn't need to be there to record a boss fight anymore. So I don't think adding all the boss encounter IDs as you have done actually gives you any benefit. See `RaidEncounter.raid()` which sets `"zoneName": "Unknown Raid",` in the metadata (but that's fine - we don't actually show that anywhere). 

Further, the arena and BG IDs you have added are just a duplicate of the IDs in `classicArenas` / `classicBattlegrounds` so I think the existing classic log handler should work on arenas, BGs and raid encounters without any changes.

aza547:
Happy to take a one liner change that adds `wow_anniversary` to the list of accepted classic log paths, that improves behaviour slightly as it avoids users needing to disable validate log paths. In-fact I'm just going to make that and close this PR. I guess a benefit of having this would be allowing configuration of both paths at once if you were switching between TBC anniversary and MOP classic. But I'm not convinced enough people are doing that to make the extra complexity in the settings worthwhile.

aza547:
https://github.com/aza547/wow-recorder/commit/478b23d30eaff4725fdebf6ef8038b044daff562

PR #846, opened 2026-06-20
Keep pull lock state across cloud POV updates

### LINKED ISSUE #843: Pull doesn't stay locked
Image

Alkwer:
## Summary

Fixes #843.

- Treat a pull as locked when any POV in the pull group is protected.
- Keep lock state consistent for grouped disk/cloud POVs in row actions, bulk actions, sorting, and cloud protection updates.
- Await disk/cloud lock persistence before updating renderer state, so the UI does not mark videos as changed when storage rejects the operation.

## Testing

- `npx.cmd eslint src/storage/DiskClient.ts src/storage/CloudClient.ts src/renderer/App.tsx src/renderer/CategoryPage.tsx src/renderer/components/Tables/Cells.tsx src/renderer/components/Tables/Sorting.ts src/renderer/videoProtection.ts src/renderer/videoProtectionActions.ts`
- `git diff --check -- src/storage/DiskClient.ts src/storage/CloudClient.ts src/renderer/App.tsx src/renderer/CategoryPage.tsx src/renderer/components/Tables/Cells.tsx src/renderer/components/Tables/Sorting.ts src/renderer/videoProtection.ts src/renderer/videoProtectionActions.ts`
- `npm.cmd run build`
- Local app smoke test with real videos: verified local lock persistence and grouped pull lock behavior after a simulated cloud POV update.

### DIFF
--- a/src/activitys/Activity.ts
+++ b/src/activitys/Activity.ts
@@ -28,6 +28,8 @@ export default abstract class Activity {
 
   protected _overrun: number = 0;
 
+  protected hash = crypto.createHash('md5');
+
   protected cfg = ConfigService.getInstance();
 
   constructor(startDate: Date, category: VideoCategory, flavour: Flavour) {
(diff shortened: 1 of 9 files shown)


aza547:
I've not looked at the code yet but very suspicious of this as I was expecting this to be a one liner

aza547:
Notes in #843 but I don't think this is the right fix.

PR #847, opened 2026-06-21
[codex] Fix repeatable activity unique hash

Alkwer:
## Summary

Fixes Activity.getUniqueHash() so repeated calls on the same activity instance remain valid and deterministic.

The previous implementation stored a Node Hash instance on the activity. Node hash instances are one-shot after digest(), so a second call could fail with Digest already called. The method now creates a fresh hash instance per call.

## Scope

- Changed only src/activitys/Activity.ts
- No AI-created or AI-modified tests included

## Validation

- git diff --check
- git diff --cached --check
- git diff --check origin/main...HEAD on the isolated PR branch
- npm run build
- Local ad-hoc runtime check confirming two consecutive getUniqueHash() calls return the same hash without throwing

### DIFF
--- a/src/activitys/Activity.ts
+++ b/src/activitys/Activity.ts
@@ -28,8 +28,6 @@ export default abstract class Activity {
 
   protected _overrun: number = 0;
 
-  protected hash = crypto.createHash('md5');
-
   protected cfg = ConfigService.getInstance();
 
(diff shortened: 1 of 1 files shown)


aza547:
Changes like this worry me but afaik we only ever call this once currently so I think this is fine and does prevent future bugs with repeated calls.

PR #848, opened 2026-06-21
Fix challenge mode timer finalization

Alkwer:
## Problem

`ChallengeModeDungeon` previously started `getKeystoneTimers()` for every dungeon flavour. For Classic/MoP Challenge Mode, the constructor then replaced timings with local MoP challenge mode timers, but the earlier remote request could still resolve later and overwrite them with Retail keystone timers.

That made Classic results unsafe, and also made Retail `upgradeLevel` timing-dependent: metadata could use local fallback timers or remote timers depending on when the remote response arrived.

## Summary

Classic/MoP Challenge Mode now uses only local challenge mode timers and no longer starts the Retail remote keystone timer request. Retail still starts with local fallback timers, but finalizes remote timer updates before metadata and filename generation.

Late remote responses after metadata finalization are ignored so `upgradeLevel`, `resultInfo`, metadata, and filename generation do not depend on remote response timing.

## Scope

- Updated `Activity` with a default `prepareMetadata()` hook
- Awaited `prepareMetadata()` in `LogHandler.endActivity()` before metadata/filename reads
- Restricted remote keystone timer fetching/finalization to Retail `ChallengeModeDungeon`
- Added comments around the race-sensitive timer flow

## Validation

- `git diff --check`
- `npm run build`
- Local ad-hoc runtime check for:
  - Classic does not fetch remote timers
  - Retail fast remote response updates timers before metadata
  - Retail slow remote response uses local fallback
  - Late remote response does not change finalized timers

No AI-created or modified tests are included.

### DIFF
--- a/src/activitys/Activity.ts
+++ b/src/activitys/Activity.ts
@@ -42,6 +42,10 @@ export default abstract class Activity {
   abstract getMetadata(): Metadata;
   abstract getFileName(): string;
 
+  prepareMetadata(): Promise<void> {
+    return Promise.resolve();
+  }
+
   get zoneID() {
     return this._zoneID;
   }
--- a/src/activitys/ChallengeModeDungeon.ts
+++ b/src/activitys/ChallengeModeDungeon.ts
@@ -20,6 +20,8 @@ import CloudClient from 'storage/CloudClient';
 export default class ChallengeModeDungeon extends Activity {
   private static readonly TIMER_GRACE_SECONDS = 1;
 
+  private static readonly REMOTE_TIMER_FINALIZATION_TIMEOUT_MS = 3000;
+
   private _mapID: number;
 
   private _level: number;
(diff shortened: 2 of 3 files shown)


aza547:
I don't understand the problem this is fixing

aza547:
Oh I see, in the MOP case:


```
    this.getKeystoneTimers();

    if (flavor === Flavour.Classic) {
      console.info('[ChallengeModeDungeon] Using Classic timers for', mapID);
      this._timings = mopChallengeModesTimers[mapID];
    }
 ```
 
Should really be 

```
    this.getKeystoneTimers();
    
    if (flavor === Flavour.Retail) {
      this.getKeystoneTimers();    
    } else if (flavor === Flavour.Classic) {
      this._timings = mopChallengeModesTimers[mapID];
    } 
 ```
 
 It's not valid to call that WCR API for a MOP dungeon in the way this is doing.

aza547:
I'm going to make that fix. I think this solution is over-engineered. The CloudClient (uses axios) is responsible for rejecting a promise that runs too long and I don't think there is realistically a race here between the promise that is invoked by the constructor resolving and the end of the dungeon. The former should be less than a second, the latter is typically 20 mins.

aza547:
https://github.com/aza547/wow-recorder/commit/983abd5a12410d6a92fdfddeb2f0e4cedac0a50d

PR #850, opened 2026-06-22
Sanitize kill video output filenames

Alkwer:
## Summary
- Reuses the existing filename sanitizer when building kill/multiview output paths.
- Prevents encounter metadata with Windows-invalid filename characters from being passed to `path.join()`.

## Problem
Normal video cuts already sanitize output filenames, but kill/multiview renders built the filename directly from encounter metadata and timestamps. If `encounterName` or related metadata contained characters such as `:`, `?`, `*`, `/`, or `|`, the rendered video path could be invalid on Windows.

## Validation
- `git diff --check -- src/main/VideoProcessQueue.ts`
- `npm.cmd run build`

### DIFF
--- a/src/main/VideoProcessQueue.ts
+++ b/src/main/VideoProcessQueue.ts
@@ -823,7 +823,7 @@ export default class VideoProcessQueue {
   }
 
   /**
-   * Prepare and return the audio map for a kill video.
+   * Prepare and return the output path for a kill video.
    */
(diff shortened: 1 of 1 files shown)


aza547:
I think there is fairly low risk of this as the videoName value is quite strictly put together from the encounter and difficulty but can't see this doing any harm.

PR #851, opened 2026-06-24
3 Features: Annotation persistence, seamless buffer replay, disk read bugfix

spencerneutron:
I apologize in advance for doing this as a single PR, but I've been working with, testing on, and using this integration branch locally for a few weeks now while raiding and wanted to share.

I'll let the code mostly speak for itself but here's an at-a-glance look:

## What this PR changes

| Change-set | Before (upstream) | After (this PR) |
|---|---|---|
| **Time-keyed VOD annotations** | An Excalidraw drawing overlay existed, but drawings were **ephemeral** and never saved, and lost on close or POV switch. One whole-VOD canvas, no time dimension, and no way to clear it. | Drawings **persist per video file** as a time-indexed **keyframe** record: each edit snapshots at the playhead, and on playback a keyframe **flashes near its timestamp** (and holds while paused). Adds scrubber keyframe markers, an adjustable display-duration slider (persisted in config), a clear-all control, and auto-removal of a keyframe once its contents are erased. Disabled in multi-POV grid mode so markup stays tied to a single viewpoint's file. |
| **Review while relocating to storage** | A freshly cut recording was moved from the local buffer to permanent storage (e.g. a NAS) **before it became reviewable**. You had to wait for the transfer to finish to watch the latest VOD. | The VOD is **reviewable immediately** from a local staging copy while it is copied to storage in the background; playback **cuts over seamlessly** to the permanent copy when the move completes (no stall on seek). A progress bar under the status indicator shows the move, and staging copies are cleaned up mid-session. |
| **In-memory metadata index** | Every video-list refresh **re-scanned the storage dir and re-parsed every `.json` sidecar** (O(N) disk I/O), amplified per operation and costly on network storage where each file is a round-trip, so cost grew with library size on every refresh. | A metadata index is **built once and kept coherent via write/delete hooks**. Refreshes are **delta-only** (a single `readdir`, parsing only newly-seen files) and effectively in-memory; annotation saves rebuild from the index without re-scanning storage. |

## Anecdotes

I've been using this branch against Retail WoW captures for over a month now. The main values adds are as follows:

- Users with a fast buffer and slow storage location and/or extremely large capture sizes (various quality settings) were generally unable to review VODs until they were moved. This is seamless, let's the move happen cleanly in the background, and generally makes the experience that much smoother when you _really_ need to start reviewing a wipe as soon as its ended. 

- Annotations are sweet, but they were mostly a real-time tool that had minimal value outside of a "hey watch my stream real quick, here's a laser pointer". Now they'll persist on the VOD, you can replay and jump around and it'll stick with it.

I added a few controls to support this, as well as the markers on the scrubber. See image:

annotations-saved-timeline-controls

- The bugfix was mainly a poor scaling on the VOD directory reads. It was a freebie to fix and the cache is durable, lightweight and better supports responsiveness of the app as the VOD library grows.

-----

I'm open to any feedback here. I definitely leaned on Claude to try and keep this as clean as I could especially since I **don't personally use the Pro features**.

If you hate the big-ass PR, I'll totally understand.

Thanks for the awesome app, hopefully my contribution here is helpful!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,13 @@ buffer
 
 # Some stuff OBS window capture creates at runtime?
 win-capture/
+
+# Local agent, planning, and editor artifacts
+.claude/
+.vs/
+AGENTS.md
+CLAUDE.md
+EXECUTION.md
+PLAN.md
+PLAN-*.md
+Stop-Node.ps1
--- a/src/config/configSchema.ts
+++ b/src/config/configSchema.ts
@@ -53,6 +53,7 @@ export type ConfigurationSchema = {
   deathMarkers: number;
   encounterMarkers: boolean;
   roundMarkers: boolean;
+  annotationFlashSeconds: number;
   pushToTalk: boolean;
   pushToTalkKey: number;
   pushToTalkMouseButton: number;
@@ -428,6 +429,14 @@ export const configSchema = {
     default: 5,
     minimum: 0,
     maximum: 60,
+  },
+  annotationFlashSeconds: {
+    // Set via the in-player annotation duration slider, not the Settings UI, so
+    // no description Phrase is needed.
+    type: 'integer',
+    default: 5,
+    minimum: 1,
+    maximum: 30,
   },
   cloudStorage: {
     description: Phrase.CloudStorageDescription,
--- a/src/main/VideoProcessQueue.ts
+++ b/src/main/VideoProcessQueue.ts
@@ -1,4 +1,5 @@
 import path from 'path';
+import { ipcMain } from 'electron';
 import {
   getBaseConfig,
   getStagingDir,
@@ -128,6 +129,19 @@ export default class VideoProcessQueue {
   private inProgressKillVideos: string[] = [];
 
   /**
+   * Resolved paths of the video sources currently loaded in the frontend
+   * player(s). Reported by the renderer. Used so we never delete a local
+   * staging copy that is currently being reviewed.
+   */
+  private activeSources = new Set<string>();
+
+  /**
+   * Staging copies that have been relocated to storage but couldn't be deleted
+   * yet because they were in use. Cleaned up once playback moves off them.
+   */
+  private pendingStagingDeletes = new Set<string>();
+
+  /**
    * Constructor.
    */
   private constructor() {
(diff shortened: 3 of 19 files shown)


aza547:
Hi thanks for the PR, some neat stuff you're trying to add here. 

# Time-keyed VOD annotations 

I like the idea a lot. Although I see it's value as limited in the local only case - am I correct in understanding that all you can do with it is draw yourself, and then review yourself? Obviously, the excalidraw feature is already pretty limited to the case you described. While you change is certainly an improvement, I reckon this would need to work for local and remote videos too. 

I wrote "Hi" and the "annotations" property in the metadata was 11kB (I presume a heavily annotated video may have up to a MB of annotation storage given that). That's going to cause problems on the backend and probably needs dedicated storage outside of the metadata database. Certainly that's solvable, but the backend is not public so means it needs to be done by me. A few more minor bits I would like to tweak in the way of buttons and styling but nothing major. 

# Review while relocating to storage

I'll need to take a look at the code to have a real opinion on this but I'm interested in how this can be implemented and provide a good UX. It's not a problem I have personally (and I'm kind of of the mindset that this is acceptable for people who opt to move their videos to NAS storage). Actually though the separation of buffers is basically pointless now that WCR doesn't constantly record. This feature has just stuck around for days before WCR had an in-memory video buffer (and everything was written to disk all the time). Why not just record straight to your NAS? In-fact, should the "separate buffer folder" feature just be removed entirely? 

# In-memory metadata index

This does sound sensible but I worry about the filesystem semantics you may be relying on, what NAS protocol are you using? NFS? I've never actually set NAS up on Windows myself, is it basically just NFS? Even within NFS though there are quite big differences between v3 and v4. What have you tested? I've still not actually looked at the code here so may have further thoughts when I get there. I did some googling, apparently "SMB" is a thing on Windows. I've never heard of it. How many videos do you have stored that this has become a problem?

spencerneutron:
Hi @aza547,

I'll get a more thorough response probably tomorrow (I'm a little tied up today) but I have some initial thoughts based on your questions scoped only to the annotations feature.

To eliminate any need for your backend code to change, I have an _idea_ that I could store the annotation metadata in the finished .MP4 file as a legal and spec-valid `uuid` box. This box can accept any user payload easily without diminishing the file or getting in the way of playback.

**My question for you**: does your backend ever remux, repackage, fragment, or transcode the MP4 or does it store and serve the file byte-for-byte?

If the answer is no, I can pursue that and mitigate any work in figuring out where to store the annotations file separately.

P.S.  I think I might close this PR and resubmit with just the annotations branch for now to separate the concerns and keep this more manageable.

aza547:
> My question for you: does your backend ever remux, repackage, fragment, or transcode the MP4 or does it store and serve the file byte-for-byte?

It does not, the backend never touches the video data which are stored as blobs in R2. But how could you modify such data once it's uploaded without a full reupload? S3/R2 doesn't allow for modifications to objects without replacing the whole thing. There is also problem of concurrent modifications to be dealt with.

aza547:
> Why not just record straight to your NAS? In-fact, should the "separate buffer folder" feature just be removed entirely?

Thinking about this a bit more there is still the remux step which would need to read from and write to NAS. So that's the reason it still exists, to keep the MKV local before remuxing.

aza547:
Going to close this as I've not had a response. Definitely some cool ideas here I might come back to but don't have the time to drive it right now.

PR #857, opened 2026-06-27
Fix repeated playback toggles when holding spacebar

### LINKED ISSUE #856: Holding spacebar breaks the video player
Weird. It's like it queues up a bunch of play/pause actions and then takes a while to recover.

Alkwer:
## Summary

This fixes a playback hotkey issue where holding Space while viewing a video could repeatedly toggle play/pause and leave the user stuck in a playback loop.

The fix keeps the scope intentionally small:

- Playback hotkeys still call `preventDefault()` / `stopPropagation()` so the browser/player does not handle Space separately.
- Repeated `keydown` events for Space/K are ignored, so holding the key only toggles playback once.
- The video selection table now returns early for repeated or unrelated key events before building sorted/selected row models.

## Root Cause

The video player handled playback hotkeys from a global `keydown` listener. Browsers keep firing repeated `keydown` events while a key is held, and the old handler treated each repeat as a fresh playback toggle.

There was also a separate global key handler in the video selection table that did table row work before confirming the key was relevant. While holding Space, those repeated events could make the UI feel unresponsive even though Space is not a table navigation key.

## Notes

This PR does not attempt to change behavior for extremely rapid manual Space/K tapping. That case appears to be separate from the reported hold-to-repeat issue and is left out to keep this fix low-risk.

Fixes #856

## Testing

- `npx.cmd eslint src/renderer/VideoPlayer.tsx src/renderer/components/Tables/VideoSelectionTable.tsx`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/src/renderer/components/Tables/VideoSelectionTable.tsx
+++ b/src/renderer/components/Tables/VideoSelectionTable.tsx
@@ -1,12 +1,7 @@
 import { AppState, RendererVideo } from 'main/types';
 
 import { Cell, flexRender, Header, Row, Table } from '@tanstack/react-table';
-import React, {
-  Fragment,
-  RefObject,
-  useCallback,
-  useEffect,
-} from 'react';
+import React, { Fragment, RefObject, useCallback, useEffect } from 'react';
 import {
   ArrowDown,
   ArrowUp,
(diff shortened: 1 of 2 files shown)


src/renderer/components/Tables/VideoSelectionTable.tsx
```diff
@@ -122,10 +117,20 @@ const VideoSelectionTable = (props: IProps) => {
    */
   useEffect(() => {
     const handleKeyDown = (event: KeyboardEvent) => {
-      const sortedRows = table.getSortedRowModel().rows;
-      const selectedRows = table.getSelectedRowModel().rows;
+      if (
+        event.repeat ||
```
aza547: I think this can just be 

if (event.repeat) {
  return;
}

without the other conditions - they are checked below

aza547 (changes requested):
minor change but otherwise looks good - i'll apply this to the website too when it merges

Alkwer:
Updated:

- simplified the early return to only skip repeated `keydown` events
- left the existing `Ctrl+A` / arrow-key branches to decide which keys are relevant

PR #858, opened 2026-06-27
[codex] Fix cross-source log activity ownership

Alkwer:
## Summary

- Track the active combat log source that owns the current recording.
- Ignore source-specific timeout, force-stop, start, end, and metadata mutation events from other log sources.
- Wire Retail PTR and Classic PTR handlers with distinct sources so PTR events cannot stop or mutate base-client recordings.

## Validation

- `git diff --check HEAD`
- TypeScript syntax transpile check for the five changed files
- Local manual append test with real `CombatLogWatcher` and temporary Retail/Classic `WoWCombatLog.txt` files: Classic timeout was ignored for an active Retail recording, and Retail `ENCOUNTER_END` still stopped the Retail recording.

## Notes

- No test files are included in this PR.

### DIFF
--- a/src/main/Manager.ts
+++ b/src/main/Manager.ts
@@ -44,7 +44,7 @@ import {
   runRetailRecordingTest,
 } from '../utils/testButtonUtils';
 import VideoProcessQueue from './VideoProcessQueue';
-import LogHandler from 'parsing/LogHandler';
+import LogHandler, { LogHandlerSource } from 'parsing/LogHandler';
 import { PTTKeyPressEvent } from 'types/KeyTypesUIOHook';
 import { send } from './main';
 import DiskClient from 'storage/DiskClient';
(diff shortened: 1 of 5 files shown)


aza547:
300 lines for this bug is a bit wild. I don't think I can consider this PR. Sorry.

Alkwer:
Sure, I'll see if I can squeeze this fix into smaller amount of lines

aza547:
I did have a think about it and decided it wasn't super obvious. LogHandler.activity being static is the route of the complication. That seemed like a good idea at the time, it might still be, but not clear. 

Ideologically there can only be one activity in progress so that's why it's like that. Maybe the log timeout should be static too? Then there is weird cases where swapping between games quickly might not end an activity from a previous flavor of game. 

It's not necessarily the number of lines I object to here FWIW - it's the tiny scope of the bug versus the large complexity in the diff.

Alkwer:
That makes sense. The large complexity came from me initially trying to model ownership at the log handler/source level. That meant introducing a new source concept for retail/classic/era/PTR handlers, tagging activities on start, and making force-end paths source-aware.

After looking again, that was more machinery than this bug needs. Since activities already carry Activity.flavour, I reworked this into a much smaller approach where a timeout ignores the active activity when the flavour does not match.

I’ll open a replacement PR with that smaller version rather than continuing with this closed one.

PR #860, opened 2026-06-28
fix: ignore timeouts from other log flavours

### LINKED ISSUE #855: Log timeout from one game mode can stop recordings of another & stop manual recordings
[WarcraftRecorder-2026-06-26.log](https://github.com/user-attachments/files/29397923/WarcraftRecorder-2026-06-26.log)

I was running the tests today while logged into classic. I think what happened is this:
1. Logged into classic. Must have written a combat log. 
2. Started tests. Got to beloren test. 
3. Classic log time out fired.
4. Interrupted tests and discarded beloren pull early. 

In summary... that could be annoying for people who are switching between games. The interrupt should be isolated to the game mode being recorded. 

```
[2026-06-26 23:33:31.231] [info]  [LogHandler] Haven't received data for combatlog in 120 seconds.
[2026-06-26 23:33:31.232] [info]  [LogHandler] Force ending activity, timedelta: -120
[2026-06-26 23:33:31.232] [info]  [Activity] Setting overrun to 0
[2026-06-26 23:33:31.232] [info]  [LogHandler] Ending recording video for category: Raids
[2026-06-26 23:33:31.233] [info]  [Recorder] Queued stop
[2026-06-26 23:33:31.233] [info]  [Recorder] Stop recording
[2026-06-26 23:33:31.234] [info]  [LogHandler] Queue buffer start as WoW still running
[2026-06-26 23:33:31.234] [info]  [Recorder] Queued start buffer
[2026-06-26 23:33:31.235] [info]  [Recorder] Got signal { type: 'output', id: 'stopping', code: 0 }
[2026-06-26 23:33:31.235] [info]  [Recorder] No action needed on this signal
[2026-06-26 23:33:31.369] [info]  [Recorder] Got signal { type: 'output', id: 'stop', code: 0 }
[2026-06-26 23:33:31.369] [info]  [Recorder] No action needed on this signal
[2026-06-26 23:33:31.397] [info]  [Recorder] Got signal { type: 'output', id: 'deactivate', code: 0 }
[2026-06-26 23:33:31.398] [info]  [Recorder] State is now: none
[2026-06-26 23:33:31.398] [info]  [Recorder] Stopped successfully
[2026-06-26 23:33:31.399] [info]  [Recorder] Start
[2026-06-26 23:33:31.447] [info]  [Recorder] Get and clear last file C:/Users/Alex/Videos/Warcraft Recorder/.temp/2026-06-26 23-33-15.mkv
[2026-06-26 23:33:31.447] [warn]  Encounter not found in known raids, can't get raid name
[2026-06-26 23:33:31.449] [warn]  Encounter not found in known raids, can't get raid name
[2026-06-26 23:33:31.449] [info]  [LogHandler] Discarding raid encounter, too short
```

Alkwer:
Fixes #855.

This is a smaller replacement for #858.

The previous approach tried to model ownership at the log handler/source level, which added more complexity than this bug needs. This version uses the existing `Activity.flavour` instead.

A combat log timeout now only force-ends the active recording when the timeout comes from the same game flavour as the active activity. For example, a Classic timeout will no longer end a Retail recording.

This keeps:
- `LogHandler.activity` static
- global `forceEndActivity()` behavior unchanged
- `WARCRAFT_RECORDER_FORCE_STOP` behavior unchanged

Validation:
- `npx eslint src/parsing/LogHandler.ts src/parsing/RetailLogHandler.ts src/parsing/ClassicLogHandler.ts src/parsing/EraLogHandler.ts`
- local append-event simulation using real `WoWCombatLog.txt` writes:
  - Classic timeout ignored during Retail recording
  - same-flavour Retail timeout still force-ends recording
  - force-stop event still ends recording globally

### DIFF
--- a/src/parsing/ClassicLogHandler.ts
+++ b/src/parsing/ClassicLogHandler.ts
@@ -5,7 +5,7 @@ import {
   classicUniqueSpecSpells,
   mopChallengeModes,
 } from '../main/constants';
-import LogHandler from './LogHandler';
+import LogHandler, { type LogHandlerSource } from './LogHandler';
 import { Flavour } from '../main/types';
 import ArenaMatch from '../activitys/ArenaMatch';
 import { isUnitFriendly, isUnitPlayer, isUnitSelf } from './logutils';
@@ -20,8 +20,8 @@ import ConfigService from 'config/ConfigService';
  * Classic log handler class.
  */
 export default class ClassicLogHandler extends LogHandler {
-  constructor(logPath: string) {
-    super(logPath, 2);
+  constructor(logPath: string, source: LogHandlerSource = 'classic') {
+    super(logPath, 2, source);
 
     /* eslint-disable prettier/prettier */
     this.combatLogWatcher
@@ -217,7 +217,7 @@ export default class ClassicLogHandler extends LogHandler {
       Flavour.Classic,
     );
 
-    await LogHandler.startActivity(activity);
+    await this.startActivity(activity);
   }
 
   private static calculateArenaResult(arenaMatch: ArenaMatch) {
@@ -381,7 +381,7 @@ export default class ClassicLogHandler extends LogHandler {
       Flavour.Classic,
     );
 
-    await LogHandler.startActivity(activity);
+    await this.startActivity(activity);
   }
 
   private async battlegroundEnd(line: LogLine) {
@@ -489,7 +489,7 @@ export default class ClassicLogHandler extends LogHandler {
       Flavour.Classic,
     );
 
-    await LogHandler.startActivity(activity);
+    await this.startActivity(activity);
   }
 
   private async handleChallengeModeEndLine(line: LogLine) {
--- a/src/parsing/LogHandler.ts
+++ b/src/parsing/LogHandler.ts
@@ -42,12 +42,25 @@ import { ESupportedEncoders } from 'main/obsEnums';
  * typically have up to 4 child classes, we don't want multiple concurrent
  * activities.
  */
+export type LogHandlerSource =
+  | 'retail'
+  | 'retailPtr'
+  | 'classic'
+  | 'classicPtr'
+  | 'era';
+
 export default abstract class LogHandler {
   public static activity: Activity | undefined;
+
+  // Tracks which watcher started the shared activity so only that watcher's
+  // timeout can end it. Global force stops deliberately ignore this.
+  private static activitySource: LogHandlerSource | undefined;
 
   public static overrunning = false;
 
   public combatLogWatcher: CombatLogWatcher;
+
+  protected readonly source: LogHandlerSource;
 
   protected player: Combatant | undefined;
 
@@ -60,7 +73,8 @@ export default abstract class LogHandler {
    */
   protected logProcessQueue = new AsyncQueue(Number.MAX_SAFE_INTEGER);
 
-  constructor(logPath: string, dataTimeout: number) {
+  constructor(logPath: string, dataTimeout: number, source: LogHandlerSource) {
+    this.source = source;
     this.combatLogWatcher = new CombatLogWatcher(logPath, dataTimeout);
     this.combatLogWatcher.watch();
     const lpq = this.logProcessQueue;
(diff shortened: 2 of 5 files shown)


src/parsing/LogHandler.ts
```diff
@@ -364,9 +367,23 @@ export default abstract class LogHandler {
...
 
-    if (LogHandler.activity) {
-      await LogHandler.forceEndActivity(-ms / 1000);
+    if (!LogHandler.activity) {
+      return;
     }
+
+    // Timeouts belong to one combat log watcher, while activity is shared
+    // globally. Ignore stale timeouts from another game flavour.
+    if (LogHandler.activity.flavour !== this.flavour) {
```
aza547: this is still problematic as timeouts from retail ptr would end retail content, although it's is certainly an improvement to isolate retail/classic

Alkwer:
Updated the patch after the review feedback:

- Timeout ownership is now tracked by concrete log source, not just flavour.
- Added distinct sources for `retail`, `retailPtr`, `classic`, `classicPtr`, and `era`.
- A Retail PTR timeout no longer ends an active Retail recording, and likewise for Classic/Classic PTR/Era.
- Same-source timeout behavior is unchanged: the watcher that started the activity can still end it on timeout.
- Global force-stop behavior is unchanged and still ends the active recording regardless of source.

Alkwer:
As per update in the issue:
- Added explicit protection for manual recordings in `dataTimeout()`.
- Combat log timeouts now return early while the active recording is `Manual`.
- Explicit/global stop paths are unchanged, so manual stop, force stop, and shutdown handling still work.

src/parsing/ClassicLogHandler.ts
```diff
@@ -217,7 +217,7 @@ export default class ClassicLogHandler extends LogHandler {
       Flavour.Classic,
     );
 
-    await LogHandler.startActivity(activity);
+    await this.startActivity(activity);
```
aza547, line 220: This change isn't functionally different, right?

src/parsing/RetailLogHandler.ts
```diff
@@ -34,8 +34,8 @@ import RaidEncounter from 'activitys/RaidEncounter';
 export default class RetailLogHandler extends LogHandler {
   private isPtr = false;
 
-  constructor(logPath: string) {
-    super(logPath, 10);
+  constructor(logPath: string, source: LogHandlerSource = 'retail') {
+    super(logPath, 10, Flavour.Retail, source);
```
aza547: Seems like there is duplicated knowledge here between the 3rd and 4th args.

src/parsing/LogHandler.ts
```diff
@@ -208,7 +230,15 @@ export default abstract class LogHandler {
     LogHandler.activity.addDeath(playerDeath);
   }
 
-  protected static async startActivity(activity: Activity) {
+  // Instance starts tag the shared activity with this watcher's source.
+  protected async startActivity(activity: Activity) {
```
aza547, line 226: Oh I see there is a non-static wrapper now

aza547 (commented):
Not sure how I feel about this one in general. Still think there might be a more elegant way to do this but want to have a think about it myself when I get some time.

src/parsing/ClassicLogHandler.ts
Alkwer: Correct, this should not change how the recording starts. The reason for switching to this.startActivity(activity) is only to remember which log handler started the activity, so timeout handling can ignore timeouts from other handlers.

src/parsing/RetailLogHandler.ts
Alkwer: Ah true, I'll change that to keep PR clean for any potential future use

Alkwer:
Sure, happy to give you some potential inspiration. If u want u can share some other idea when you will have one for this issue

aza547:
Had a look at this today and did this: https://github.com/aza547/wow-recorder/pull/871

I think it's better as the ownership of the timer is now clearer. Will need to do some testing of it but let me know if you disagree with the approach :)

PR #866, opened 2026-07-12
Handle unknown dungeon map IDs

Alkwer:
## Summary

- Make `getDungeonName` always return a string.
- Fall back to `"Unknown Map"` when a dungeon map ID cannot be resolved.

## Why

The current recording path rejects unknown map IDs, but persisted or cloud metadata is loaded without runtime validation. If it contains an unrecognized map ID, `getDungeonName` previously returned `undefined`. Clip activity sorting later calls `localeCompare` on the returned value, which could throw and break the clips table.

## Validation

- Local regression checks for unknown Retail and Classic map IDs and clip sorting.
- `npm run build:renderer`
- Targeted ESLint and Prettier checks for the changed file.
- `git diff --check`

### DIFF
--- a/src/renderer/rendererutils.ts
+++ b/src/renderer/rendererutils.ts
@@ -321,22 +321,24 @@ const getEncounterNameById = (encounterId: number): string => {
 };
 
 /**
- * Get the dungeon name if possible, else an empty string.
+ * Get the dungeon name if possible, else "Unknown Map".
  */
-const getDungeonName = (video: RendererVideo) => {
+const getDungeonName = (video: RendererVideo): string => {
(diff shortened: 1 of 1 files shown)


aza547:
I don't think it's possible to record unknown maps so not sure how this can ever be hit?

Although if did, shouldn't this return "Unknown Map" rather than an empty string?

Alkwer:
You're right — the current recording flow rejects unknown map IDs. This is only a defensive fallback for legacy or invalid metadata though this is prob some super edge case. Also, Unknown Map makes more sense.

aza547:
Looks safe. Thanks.

PR #870, opened 2026-07-17
Add manual disk size monitor action

### LINKED ISSUE #867: add "run now" button for disk size monitor
Image

Alkwer:
## Summary

- add a **Run now** action next to disk usage in General Settings
- invoke the existing disk size monitor through the typed preload/IPC bridge
- prevent overlapping monitor runs and refresh disk status and videos before completion

## Why

This lets users run storage cleanup immediately instead of waiting for the automatic trigger.

Fixes #867

## Validation

- exercised the full UI → preload → IPC → disk monitor flow in an isolated Electron profile
- verified repeated runs complete without renderer exceptions
- `npm run build`
- targeted ESLint checks for the changed TypeScript files
- `git diff --check`

### DIFF
--- a/src/renderer/GeneralSettings.tsx
+++ b/src/renderer/GeneralSettings.tsx
@@ -286,8 +286,6 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
 
     try {
       await ipc.runDiskSizeMonitor();
-    } catch (error) {
-      console.error('[GeneralSettings] Failed to run disk size monitor', error);
     } finally {
       setIsRunningDiskSizeMonitor(false);
     }
@@ -300,7 +298,6 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
     if (perc > 100) perc = 100;
     const text = max === 0 ? `${usage}GB / ∞` : `${usage}GB / ${max}GB`;
 
-    // Compare unrounded bytes so the action only appears after the configured limit is exceeded.
     const shouldShowRunButton =
       config.maxStorage > 0 &&
       appState.diskStatus.usage > appState.diskStatus.limit;
@@ -311,7 +308,9 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
           {getLocalePhrase(language, Phrase.DiskUsageDescription)}
         </Label>
 
-        <div className="flex flex-row items-center justify-start w-[430px] gap-x-2 py-2">
+        <div
+          className={`flex flex-row items-center justify-start ${shouldShowRunButton ? 'w-[430px]' : 'w-80'} gap-x-2 py-2`}
+        >
           <Tooltip
             content={getLocalePhrase(language, Phrase.DiskUsageDescription)}
           >
@@ -326,7 +325,6 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
               size="xs"
               onClick={runDiskSizeMonitor}
               disabled={isComponentDisabled() || isRunningDiskSizeMonitor}
-              aria-busy={isRunningDiskSizeMonitor}
             >
               <span
                 className={`mr-1.5 inline-flex ${isRunningDiskSizeMonitor ? 'animate-spin' : ''}`}
--- a/src/renderer/GeneralSettings.tsx
+++ b/src/renderer/GeneralSettings.tsx
@@ -300,6 +300,11 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
     if (perc > 100) perc = 100;
     const text = max === 0 ? `${usage}GB / ∞` : `${usage}GB / ${max}GB`;
 
+    // Compare unrounded bytes so the action only appears after the configured limit is exceeded.
+    const shouldShowRunButton =
+      config.maxStorage > 0 &&
+      appState.diskStatus.usage > appState.diskStatus.limit;
+
     return (
       <div className="flex-col">
         <Label className="flex items-center">
@@ -316,23 +321,21 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
           <span className="text-[11px] text-foreground font-semibold whitespace-nowrap">
             {text}
           </span>
-          <Button
-            size="xs"
-            onClick={runDiskSizeMonitor}
-            disabled={
-              isComponentDisabled() ||
-              isRunningDiskSizeMonitor ||
-              config.maxStorage <= 0
-            }
-            aria-busy={isRunningDiskSizeMonitor}
-          >
-            <span
-              className={`mr-1.5 inline-flex ${isRunningDiskSizeMonitor ? 'animate-spin' : ''}`}
+          {shouldShowRunButton ? (
+            <Button
+              size="xs"
+              onClick={runDiskSizeMonitor}
+              disabled={isComponentDisabled() || isRunningDiskSizeMonitor}
+              aria-busy={isRunningDiskSizeMonitor}
             >
-              <RefreshCw size={14} />
-            </span>
-            {getLocalePhrase(language, Phrase.RunDiskSizeMonitorButtonLabel)}
-          </Button>
+              <span
+                className={`mr-1.5 inline-flex ${isRunningDiskSizeMonitor ? 'animate-spin' : ''}`}
+              >
+                <RefreshCw size={14} />
+              </span>
+              {getLocalePhrase(language, Phrase.RunDiskSizeMonitorButtonLabel)}
+            </Button>
+          ) : null}
         </div>
       </div>
     );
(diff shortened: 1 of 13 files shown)


Alkwer:
image

src/storage/DiskSizeMonitor.ts
```diff
@@ -78,8 +91,11 @@ export default class DiskSizeMonitor {
     );
 
     if (filesForDeletion.length > 0) {
-      DiskClient.getInstance().refreshStatus();
-      DiskClient.getInstance().refreshVideos();
+      // Resolve only after renderer-facing state reflects the completed cleanup.
+      await Promise.all([
```
aza547, line 165: I don't think this really does anything other than wait a bit longer, the await doesn't actually wait on the frontend rendering, just on some of the filesystem calls in these methods.

src/storage/DiskSizeMonitor.ts
```diff
@@ -17,9 +17,22 @@ const asyncFilter = async (fileStream: FileInfo[], filter: any) => {
 };
 
 export default class DiskSizeMonitor {
+  // Call sites create separate instances, so share the in-flight run to avoid overlapping cleanup.
+  private static activeRun: Promise<void> | undefined;
```
aza547: Good spot that this is required. I guess there is also some risk of this happening concurrently already, probably very unlikely but if the call in videoProcessQueue is invoked in quick succession I'm not sure on the behaviour. Might be harmless but certainly wasn't intended to be ran concurrently.

But I'm not sure this does the right thing. Clicking the button mid-run will not pick up the newest settings? It will just not trigger an extra run. I reckon you might also want to queue another run. Perhaps the right way to implement this is to use the existing pattern:
```
const diskSizeMonitorQueue= new AsyncQueue(1); // Limit 1 means we drop any unrequired work
diskSizeMonitorQueue.add( ... ) ;
```

That could live in the videoProcessQueue and also fix the risk of concurrent runs there.

aza547:
Also should probably only show the button when the used > max ? Don't think there is a reason to show it if not

Alkwer:
Summary of the last 2 commits, as per comments:
- Replaced the shared in-flight monitor promise with AsyncQueue(1), preventing concurrent cleanup runs while retaining one pending rerun.
- Extended AsyncQueue to return completion promises, propagate task failures, and let excess callers await the latest accepted task.
- Restored disk status and video refreshes to their previous non-blocking behavior.
- Changed the Run now button to appear only when disk usage exceeds the configured limit.
- Used unrounded byte values for the visibility check and added comments explaining the queue and visibility behavior.

src/renderer/GeneralSettings.tsx
```diff
@@ -278,20 +281,37 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
     );
   };
 
+  const runDiskSizeMonitor = async () => {
+    setIsRunningDiskSizeMonitor(true);
+
+    try {
+      await ipc.runDiskSizeMonitor();
+    } catch (error) {
+      console.error('[GeneralSettings] Failed to run disk size monitor', error);
```
aza547: these logs don't go anywhere, you should log errors from the main process.

src/renderer/GeneralSettings.tsx
```diff
@@ -278,20 +281,37 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
...
+  };
+
   const getDiskUsageBar = () => {
     const usage = Math.round(appState.diskStatus.usage / 1024 ** 3);
     const max = Math.round(appState.diskStatus.limit / 1024 ** 3);
     let perc = max === 0 ? 100 : (100 * usage) / max;
     if (perc > 100) perc = 100;
     const text = max === 0 ? `${usage}GB / ∞` : `${usage}GB / ${max}GB`;
 
+    // Compare unrounded bytes so the action only appears after the configured limit is exceeded.
```
aza547: this comment isn't saying anything that isn't obvious

src/renderer/GeneralSettings.tsx
```diff
@@ -278,20 +281,37 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
...
+      appState.diskStatus.usage > appState.diskStatus.limit;
+
     return (
       <div className="flex-col">
         <Label className="flex items-center">
           {getLocalePhrase(language, Phrase.DiskUsageDescription)}
         </Label>
 
-        <div className="flex flex-row items-center justify-start w-80 gap-x-2 py-2">
+        <div className="flex flex-row items-center justify-start w-[430px] gap-x-2 py-2">
```
aza547: you are changing the width regardless of if the button is showing?

src/renderer/GeneralSettings.tsx
```diff
@@ -301,6 +321,21 @@ const GeneralSettings: React.FC<IProps> = (props: IProps) => {
           <span className="text-[11px] text-foreground font-semibold whitespace-nowrap">
             {text}
           </span>
+          {shouldShowRunButton ? (
+            <Button
+              size="xs"
+              onClick={runDiskSizeMonitor}
+              disabled={isComponentDisabled() || isRunningDiskSizeMonitor}
+              aria-busy={isRunningDiskSizeMonitor}
```
aza547: the rest of the codebase doesn't apply aria props so no need here

src/utils/AsyncQueue.ts
```diff
@@ -2,16 +2,34 @@ export default class AsyncQueue {
...
   }
 
-  public add(task: () => Promise<void>) {
+  public add(task: () => Promise<void>): Promise<void> {
     // Just drop any tasks added over the limit.
-    if (this.queue.length >= this.limit) return;
-    this.queue.push(task);
+    if (this.queue.length >= this.limit) return this.latestCompletion;
+
+    const completion = new Promise<void>((resolve, reject) => {
```
aza547: hmm I need to think about this. I get the need for feedback when finished for this button but I don't like touching this code

Alkwer: I think the other option, if you want to keep AsyncQueue unchanged, would be to implement the completion signal locally in DiskSizeMonitor instead of changing AsyncQueue - would that be preferable or do you want to think about another way?

aza547: yeah i think that is a better change for now, i'm aware that adding this to AsyncQueue is not necessarily wrong but it just opens the scope of this PR to possibly impacting loads of important parts of the processing and if we can avoid that worry I think it's better

aza547: I guess the run task can just send a "done running size monitor" style message to the frontend that is ignored if the button isn't shown / is not disabled (as the trigger may be automatic run on end of recording, or manual click of the button)

Alkwer: Generally yes, though then I guess could be some edge case scenario where someone would run manual monitor right after automatic one and not get message because automatic monitor would hide the button

aza547: Not sure that's a concern if the response to the signal is to show the button (if other conditions are still met)? then you can handle the disabling on the frontend. You don't need to disable it when an automatic run is happening because it's concurrency is handled by the queue. Although I guess you only want to fire the signal on the queue being empty else it might re-enable the button pre-emptively.

aza547: maybe the automatic triggers just don't send the signal though which might be a better simplification than sending a signal on end of queue (which would once again probably need modifications to the AsyncQueue class)

Alkwer:
I did a feedback message as a separate one, automatic monitor runs bypass IPC handler and therefore do not send the event

src/storage/DiskSizeMonitor.ts
```diff
@@ -16,10 +17,48 @@ const asyncFilter = async (fileStream: FileInfo[], filter: any) => {
   return fileStream.filter((_, index) => results[index]);
 };
 
+// All monitor instances share the queue, retaining at most one rerun while cleanup is active.
+const diskSizeMonitorQueue = new AsyncQueue(1);
```
aza547, line 23: Ah I was still hoping we could do without this. I've sold you a lemon here and I can see why it's not achievable with the suggestion / constraints I gave.

aza547: Specifically the handling below re the completion.

aza547:
This existing behaviour here is actually worse than I thought - looks like it's totally possible to have concurrent runs of the size monitor as it's called both in `videoQueueEmpty` and `downloadQueueEmpty`. It will be good to resolve the chance of that.

src/storage/DiskSizeMonitor.ts
Alkwer: Got it. Should I remove the completion handling and diskSizeMonitorComplete, but keep AsyncQueue(1) to prevent concurrent runs? This would mean videoQueueEmpty no longer waits for the cleanup to finish — is that okay?

aza547: I'd like to have another look at this when I get a chance first.

PR #872, opened 2026-07-19
Clip source seeking

aza547:
Implements clip source seeking. 

Builds on https://github.com/aza547/wow-recorder/pull/838. 

Also added some hotkey tips that I meant to add for a while.

Still need:
- [x] A bunch of translations.
- [x] Server side handling (probably a schema change :x)

I think I've solved some long-standing tech debt here where a common state of the VideoSelectionTable was no rows selected. I'm a bit scared to change the handling of the no rows selected case though because it was a common source of renderer errors (and subsequent blank screens - these days that would hit the error boundary but still not good).

These changes will also need applied to the webapp.


src/renderer/VideoPlayer.tsx
```diff
@@ -553,7 +553,6 @@ export const VideoPlayer = forwardRef<VideoPlayerRef, IProps>((props, ref) => {
       return;
     }
 
-    persistentProgress.current = player1.current.currentTime;
```
aza547, line 556: Pretty sure this was just a mistake in the ReactPlayer v3 refactor. Can't see why this would be desirable.

aza547: It also explains why it might have appeared extra logic was required but simply setting persistent progress should be enough to seek the player on mount.

aza547:
Feel free to let me know if you have any thoughts @Alkwer but happy now to merge this so I'm going for it.

aza547:
also raised https://github.com/aza547/wow-recorder/issues/874 so i don't forget to port this to the webapp too

src/renderer/StorageFilterToggle.tsx
```diff
@@ -32,11 +32,8 @@ const StorageFilterToggle = (props: IProps) => {
       return;
     }
 
-    table.toggleAllRowsSelected(false);
-
     setAppState((prevState) => ({
       ...prevState,
-      selectedVideos: [],
```
Alkwer, line 39: Isn't it the case that now, after removing this line, changing the storage filter no longer clears either the table row selection or appState.selectedVideos? If the selected item is excluded by the new filter, the player can keep showing the old video while the table and bulk actions operate on a different visible row.

src/renderer/components/Tables/Cells.tsx
```diff
@@ -220,6 +221,41 @@ export const populateDetailsCell = (
...
+  };
+
+  return (
+    <Box className="inline-flex">
+      <Tooltip content={tooltip}>
+        <Button
+          variant="ghost"
+          size="xs"
+          onClick={goToSource}
+          disabled={disabled}
```
Alkwer, line 250: Small thing - button applies disabled:pointer-events-none, and the tooltip uses this button directly as its trigger. When the source is unavailable, the disabled button therefore cannot receive hover events and the “Source unavailable” tooltip will not appear. Prob would be good to wrap it in div or span.

Alkwer:
Looks good to me - added some small comments but not the most critical things

src/renderer/StorageFilterToggle.tsx
aza547: yeah you are right. I think that's not the end of the world but I do intend to revisit the storage filter under https://github.com/aza547/wow-recorder/issues/875

src/renderer/components/Tables/Cells.tsx
aza547: good point! I think similar is true for the kill video creator icon which i also noticed seemed to be missing a tooltip when disabled

PR #878, opened 2026-07-23
Keep Instant Replay available during overrun

### LINKED ISSUE #877: instant replay should be available in overrun

Alkwer:
## What changed

- Preserve the current Instant Replay state while a recording is in overrun.
- Reset Instant Replay normally once the overrun has finished.

Fixes #877.

## Root cause

When an activity ended, `LogHandler.activity` was cleared before the configured overrun completed. `Manager.refreshStatus()` treated the missing activity as a completed recording and reset the renderer's Instant Replay state, even though the recorder was still producing a valid replay during overrun.

## Impact

Instant Replay remains visible and available for the full recording overrun, then disappears when the recorder returns to its ready state.

## Validation

- `npm run build`
- `npx eslint src/main/Manager.ts`
- Temporary regression test against `Manager.refreshStatus()`: failed before the fix and passed after it
- Simulated Mythic raid E2E: recording produced successfully, boss HP metadata was correct, and the UI showed `Overrunning` with an active `Instant Replay` button

### DIFF
--- a/src/main/Manager.ts
+++ b/src/main/Manager.ts
@@ -307,7 +307,7 @@ export default class Manager {
 
     if (LogHandler.activity) {
       refreshInstantReplayState(LogHandler.activity);
-    } else {
+    } else if (!inOverrun) {
       resetInstantReplayState();
     }
   }


aza547:
looks right - only concern about this is if there is a timing window between the state update on the frontend where this is clickable that could result in an error

aza547:
I guess if that window did exist it is already present for the 0 overrun case...

PR #882, opened 2026-07-27
Record audio in separate tracks

### LINKED ISSUE #730: Record Audio in seperate Tracks
When selecting to record audio from individual applications, there should be a checkbox to create separate audio tracks for each source, allowing more options using the videos for editing.

OBS in general supports this kind of setting: https://www.reddit.com/r/obs/comments/11pnn29/how_to_separate_audio_tracks_when_recording/

I have looked into the code of obs-studio-node and it seems like the recording settings do not allow to set this option yet. I created a related issue in this project: https://github.com/streamlabs/obs-studio-node/issues/1574

Alkwer:
Closes #730

## Summary
- add per-source routing across OBS audio tracks 1-6 using the OBS/noobs mixer bitmask
- default existing and newly created audio sources to track 1 when no assignment is stored
- expose compact Source and Tracks columns in the audio source table
- apply routing through `noobs` 0.0.204
- preserve every audio stream when cutting finished recordings and clips
- track audio sources created and deleted through the UI so OBS lifecycle cleanup does not leave duplicate sources

Recordings remain a single MP4 file. Compatible players and editors can select any of the recorded audio streams.

## Validation
- `npm run build`
- recorded a real WoW session with sources routed to separate tracks
- verified the resulting MP4 contains six stereo AAC audio streams with `ffprobe`
- verified the streams contain the audio sources selected in the UI
- verified routing persists across restart
- launched and exercised an unsigned production `win-unpacked` build
- ran an isolated end-to-end recording through the real app, noobs, OBS, and `VideoProcessQueue`
- verified track controls remain editable while the recorder is ready
- reproduced the existing dynamic-source cleanup bug, then verified added sources are removed and recreated without duplication
- visually verified the final Source/Tracks table layout and control alignment

## Local limitations
- repository-wide lint and Jest commands currently report existing unrelated errors
- installer signing was unavailable locally, so production-mode verification used the unsigned unpacked build


Alkwer:
image

src/config/configSchema.ts
```diff
@@ -14,6 +14,7 @@ export type ConfigurationSchema = {
   monitorIndex: number;
   selectedCategory: number;
   audioSources: AudioSource[];
+  separateAudioTracks: boolean;
```
aza547: Do we need this?

src/config/configSchema.ts
```diff
@@ -176,16 +177,23 @@ export const configSchema = {
         device: 'default',
         volume: 1,
         type: AudioSourceType.OUTPUT,
+        audioTracks: [1, 2],
```
aza547: Reckon this should be a number typed bitmask to match the noobs/obs API

src/main/Manager.ts
```diff
@@ -613,6 +613,7 @@ export default class Manager {
           offset,
           duration,
           clip: true,
+          preserveAllAudioStreams: true,
```
aza547: A bit unsure why this is required but i'll read on

src/main/preload.ts
```diff
@@ -145,8 +147,12 @@ contextBridge.exposeInMainWorld('electron', {
     },
 
     // Also returns the properties.
-    createAudioSource(id: string, type: AudioSourceType): Promise<string> {
-      return ipcRenderer.invoke('createAudioSource', id, type);
+    createAudioSource(
+      id: string,
+      type: AudioSourceType,
+      audioTracks: number[],
```
aza547: `tracks: number`

src/main/types.ts
```diff
@@ -318,6 +319,7 @@ type AudioSource = {
   friendly?: string; // A user-friendly name for the source
   device?: string | number; // Machine friendly identifier for the device or window, I think this can only be a string in practice.
   volume: number; // Current volume setting (0-1)
+  audioTracks?: number[]; // OBS output tracks this source is routed to.
```
aza547: `tracks: number`

src/config/ConfigService.ts
```diff
@@ -229,6 +232,23 @@ export default class ConfigService
     );
   }
 
+  /**
+   * Add default track assignments to audio sources saved by older versions.
+   */
+  private migrateAudioSourceTracks(): void {
```
aza547: I don't think this will need to exist, if the `tracks` property of `AudioSource` doesn't exist then we should just default to track 1

src/main/Recorder.ts
```diff
@@ -370,11 +375,18 @@ export default class Recorder extends EventEmitter {
 
     ipcMain.handle(
       'createAudioSource',
-      (event, id: string, type: AudioSourceType) => {
+      (_event, id: string, type: AudioSourceType, tracks: number[]) => {
```
aza547: tracks are already included as a property on AudioSourceType?  Why pass separately?

src/main/Recorder.ts
```diff
@@ -370,11 +375,18 @@ export default class Recorder extends EventEmitter {
...
         const name = noobs.CreateSource(id, type);
         console.info('[Manager] Created audio source', name);
+        const audioTracks = normalizeAudioTracks(tracks);
+        const separateAudioTracks = this.cfg.get<boolean>(
+          'separateAudioTracks',
+        );
+
+        this.configureAudioSourceTracks(name, audioTracks, separateAudioTracks);
         noobs.AddSourceToScene(name);
+        this.audioSources.push({ id: name, type, volume: 1, audioTracks });
```
aza547: Hmm what is going on here. Why are we not already pushing this to `this.audioSources`? Is this an existing bug? ...and removing it in the deleteAudioSources method?

src/main/Recorder.ts
```diff
@@ -387,6 +399,9 @@ export default class Recorder extends EventEmitter {
     ipcMain.on('deleteAudioSource', (_event, id: string) => {
       console.info('[Manager] Deleting audio source', id);
       noobs.DeleteSource(id);
+      this.audioSources = this.audioSources.filter(
```
aza547, line 399: This looks right. I'm confused why it doesn't already exist. I think the following is broken without this
- Launch WCR
- Add audio source in settings
- Leave audio tab
- Audio sources should all be removed, but the newly created one is not tracked in this.audiosources so it isnt
- Launch WoW... all sources are created from config, duplicated audio?

src/utils/audioTracks.ts
```diff
@@ -0,0 +1,81 @@
+import type { AudioSource } from 'main/types';
+
+const OBS_AUDIO_TRACKS = [1, 2, 3, 4, 5, 6] as const;
+type AudioTrackNumber = (typeof OBS_AUDIO_TRACKS)[number];
+
+const MIXED_AUDIO_TRACK = 1;
+const MIXED_AUDIO_TRACK_MASK = 1;
+const ISOLATED_AUDIO_TRACKS = OBS_AUDIO_TRACKS.slice(1);
+
+const normalizeAudioTracks = (tracks: unknown): AudioTrackNumber[] => {
```
aza547: I think you should just avoid this and keep it as a bitmask the whole time. Applies to this whole file?  See test/track.js in noobs

src/renderer/AudioSourceControls.tsx
```diff
@@ -86,6 +96,14 @@ const AudioSourceControls = (props: IProps) => {
     }
   });
 
+  const audioTrackRoutingLocked = [
```
aza547: Why are you locking this? I think all this can be changed mid-recording?

aza547 (changes requested):
Quick first pass. 
- Think the config is more complex than it needs to be.
- Think you are locking audio settings for no reason.
- Expect simplifying this removes a lot of the code here.

Also confused by the existing lack of adding/removing to this.audioSources when IPC calls are made. That might be an existing bug.

UI looks sensible, probably have a few minor thoughts on the styling.

src/main/VideoProcessQueue.ts
```diff
@@ -733,6 +733,7 @@ export default class VideoProcessQueue {
     }
 
     const outputPath = VideoProcessQueue.getOutputVideoPath(data, outputDir);
+    const audioMap = data.preserveAllAudioStreams ? '-map 0:a?' : '-map 0:a:0?';
```
aza547: Not sure why you need to pass this bool. Why wouldn't you want to preserve the tracks?

aza547:
> image

Overall looks really good - I wonder about the tracks label. Does it need to exist? I'd like to try to keep this compact if possible. I reckon a tooltip might do, or maybe a table header? Also should the track selections be taller to match the row?

src/config/configSchema.ts
```diff
@@ -176,13 +176,15 @@ export const configSchema = {
         device: 'default',
         volume: 1,
         type: AudioSourceType.OUTPUT,
+        tracks: 1,
```
aza547: Probably good to be clear here that this is a bitmask?

`1 << 0`

would do that even though it's not incorrect.

Alternatively some constants?

Alkwer: Updated both default audio sources to use 1 << 0

src/main/Recorder.ts
```diff
@@ -370,11 +370,13 @@ export default class Recorder extends EventEmitter {
 
     ipcMain.handle(
       'createAudioSource',
-      (event, id: string, type: AudioSourceType) => {
+      (_event, id: string, type: AudioSourceType) => {
         console.info('[Manager] Creating audio source', id, 'of type', type);
         const name = noobs.CreateSource(id, type);
         console.info('[Manager] Created audio source', name);
+        this.configureAudioSourceTracks(name, 1);
```
aza547: I want to avoid the confusion here: this is not a number 1-6 but a bitmask. Let's add some constants to an appropriate location that we can use. 

const track1  = 1 << 0;
const track2 = 1 << 1;

etc. 

All the references to tracks should use the constants.

src/config/configSchema.ts
```diff
@@ -176,13 +176,15 @@ export const configSchema = {
         device: 'default',
         volume: 1,
         type: AudioSourceType.OUTPUT,
+        tracks: 1 << 0,
```
aza547: Let's use a constant as per later comment

src/main/Recorder.ts
```diff
@@ -387,6 +389,9 @@ export default class Recorder extends EventEmitter {
     ipcMain.on('deleteAudioSource', (_event, id: string) => {
       console.info('[Manager] Deleting audio source', id);
       noobs.DeleteSource(id);
```
aza547, line 398: Also a bit surprised this doesn't call removeSourceFromScene. Should it?

src/main/Recorder.ts
```diff
@@ -865,6 +882,7 @@ export default class Recorder extends EventEmitter {
         console.warn('[Recorder] Unable to configure audio source', src);
       }
 
+      this.configureAudioSourceTracks(name, src.tracks ?? 1);
```
aza547: another place for a constant

src/main/types.ts
```diff
@@ -318,6 +318,7 @@ type AudioSource = {
   friendly?: string; // A user-friendly name for the source
   device?: string | number; // Machine friendly identifier for the device or window, I think this can only be a string in practice.
   volume: number; // Current volume setting (0-1)
+  tracks?: number; // OBS tracks represented as a mixer bitmask.
```
aza547: worth a comment that there are 6 tracks exactly, represented by the 6 least significant bits

src/main/VideoProcessQueue.ts
```diff
@@ -741,6 +741,8 @@ export default class VideoProcessQueue {
       // re-encoding which would take time and CPU.
       .withVideoCodec('copy')
       .withAudioCodec('copy')
+      .outputOption('-map 0:v:0')
```
aza547, line 758: this worries me a bit. can you tell me why this is required?

aza547: ditto all the other ffmpeg args here

src/renderer/AudioSourceControls.tsx
```diff
@@ -48,6 +52,8 @@ import { Phrase } from 'localisation/phrases';
 
 const ipc = window.electron.ipcRenderer;
 let debounceTimer: NodeJS.Timeout | undefined;
+const OBS_AUDIO_TRACKS = [1, 2, 3, 4, 5, 6];
```
aza547: prefer to keep the casing consistent please - obsAudioTracks / defaultAudioTrack

src/renderer/AudioSourceControls.tsx
```diff
@@ -447,6 +453,25 @@ const AudioSourceControls = (props: IProps) => {
     }
   };
 
+  const setSourceAudioTracks = (src: AudioSource, values: string[]) => {
+    const tracks = values.reduce(
+      (mask, value) => mask | (1 << (Number(value) - 1)),
+      0,
+    );
+    if (tracks === 0) return;
```
aza547, line 462: can this ever happen? fine to be defensive, but worth a comment if it's impossible atleast

src/renderer/AudioSourceControls.tsx
```diff
@@ -447,6 +453,25 @@ const AudioSourceControls = (props: IProps) => {
...
+
+    setConfig((prev) => {
```
aza547: I think you can simplify this to

```
setConfig((prev) => ({
  ...prev,
  audioSources: prev.audioSources.map((source) =>
    source.id === src.id ? { ...source, tracks } : source,
  ),
}));
```

aza547 (changes requested):
Looks really good. A few minor changes. Curious to discuss the the ffmpeg args. Also you should add a change log entry and reference the issue - this is a highly in demand feature, the people will want to know!

src/main/Recorder.ts
Alkwer: True, good catch, added that in local patch

src/main/Recorder.ts
Alkwer: Yes, this appears to have been an existing inconsistency. Sources created through IPC were not added to this.audioSources, so they could remain in OBS after leaving the Audio tab - the patch now tracks them immediately.

src/main/VideoProcessQueue.ts
Alkwer: FFmpeg's automatic stream selection keeps only one audio stream, so without explicit mapping tracks 2-6 would be dropped when cutting a recording. `-map 0:v:0` selects the first video stream, while `-map 0:a?` preserves all audio streams; the `?` allows files without audio. Both streams are copied without re-encoding.
Kill-video rendering is different because it concatenates multiple POV segments into an output with a single audio track. The `:a:0` selectors explicitly use OBS track 1, preserving the previous single-track behaviour. The other existing FFmpeg options were not changed.

src/renderer/AudioSourceControls.tsx
Alkwer: It can happen if user tries to deselect the last selected track

src/main/VideoProcessQueue.ts
Alkwer: Updated in 3acee4eb. Kill-video rendering now probes the inputs and preserves all corresponding audio streams. In the normal mode, track N from each POV is spliced into output track N, limited to the stream count shared by every input. In Single Player Audio mode, all audio streams from the selected POV are mapped for the whole video. Explicit mapping is still required because FFmpeg automatic stream selection would otherwise keep only one audio stream.

src/main/Recorder.ts
aza547: great

src/main/VideoProcessQueue.ts
```diff
@@ -741,6 +753,10 @@ export default class VideoProcessQueue {
       // re-encoding which would take time and CPU.
       .withVideoCodec('copy')
       .withAudioCodec('copy')
+      // FFmpeg otherwise selects only one audio stream. OBS recordings can
+      // contain six, so preserve them all while allowing sources without audio.
+      .outputOption('-map 0:v:0')
+      .outputOption('-map 0:a?')
```
aza547, line 759: I think both these options can collapse to `-map 0`?

src/main/VideoProcessQueue.ts
```diff
@@ -44,6 +44,7 @@ let ffmpegPathAbs = devMode
 
 ffmpegPathAbs = fixPathWhenPackaged(ffmpegPathAbs);
 ffmpeg.setFfmpegPath(ffmpegPathAbs);
+ffmpeg.setFfprobePath(path.join(path.dirname(ffmpegPathAbs), 'ffprobe.exe'));
```
aza547, line 47: i'd rather not use ffprobe unless it's a last resort, it's just extra complexity that doesn't need to be here. the video is aware of what version it is recorded at, we can just check that

src/main/VideoProcessQueue.ts
```diff
@@ -850,25 +866,74 @@ export default class VideoProcessQueue {
...
   /**
-   * Prepare and return the audio map for a kill video.
+   * Return the number of audio streams that can be preserved in a kill video.
+   * Spliced audio is limited to streams shared by every input, while the
+   * single-player mode can preserve every stream from the selected input.
    */
-  private static prepareKillVideoAudioMap(item: KillVideoQueueItem) {
-    const map =
+  private static async getKillVideoAudioTrackCount(
+    item: KillVideoQueueItem,
```
aza547, line 874: this whole thing can be `num = video.appVersion > n ? 6 : 1`

src/main/VideoProcessQueue.ts
```diff
@@ -850,25 +866,74 @@ export default class VideoProcessQueue {
...
+            });
+          }),
+      ),
+    );
+
+    const count =
       item.audioTrackIndex === -1
-        ? '-map [a]'
-        : `-map ${item.audioTrackIndex}:a`;
+        ? Math.min(...counts)
```
aza547, line 902: don't understand this but i think this function should just be deleted

src/main/VideoProcessQueue.ts
```diff
@@ -429,31 +430,42 @@ export default class VideoProcessQueue {
 
     const first = item.segments[0].video;
     const videoPath = VideoProcessQueue.prepareKillVideoPath(first);
-    const audioMap = VideoProcessQueue.prepareKillVideoAudioMap(item);
-    const filter = VideoProcessQueue.prepareKillVideoComplexFilter(item);
-
-    const fn = ffmpeg()
```
aza547, line 435: i can't see the reason for refactoring this? ill read on

aza547:
Thanks for the efforts - I'd like to pick this up to get it over the line. Should have some time this weekend. Mostly I'm nervous about the VideoProcessQueue stuff and want to consider it carefully. The rest looks good.

src/main/VideoProcessQueue.ts
Alkwer: True, you can also do this that way

src/main/VideoProcessQueue.ts
Alkwer: That is also the way if we establish which what number will have the version from which there will be cutoff

src/main/VideoProcessQueue.ts
Alkwer: If you want to use appVersion approach then you don't need it

src/main/VideoProcessQueue.ts
Alkwer: I moved it because the audio track count was determined asynchronously using ffprobe. With ffprobe removed, this refactor is no longer needed and the original structure can be restored.

Alkwer:
Sure, np, happy to help if you would have any questions or doubts

aza547:
Merged this under https://github.com/aza547/wow-recorder/pull/885

Issue #886, opened 2026-08-02
Multidelete dialog table selection is broken


aza547:
<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/dcbbc0d3-6e0a-467f-8a88-a6fa7cdcb826" />

Guessing react compiler is the cause, clicking buttons targets the wrong row.

Alkwer:
Apparently TanStack works with React Compiler only from v9 which is still in beta - would it be the best to turn off Compiler for the whole module DeleteDialog.tsx ?

aza547:
bleh, nice spot. i might just try to take v9 as i'm reworking a bunch of things here anyway.

PR #888, opened 2026-08-09
Revamp Lock, Tag, Delete Dialogs

### LINKED ISSUE #843: Pull doesn't stay locked

aza547:
Will fix https://github.com/aza547/wow-recorder/issues/843 but will need some special attention to locking. Specifically after this ships, want to wait a few months, then lock every video that has a locked correlated video and lift the housekeeper restrictions so it can remove unlocked videos in a group with correlated videos (I know, what a mess). 

Fixing https://github.com/aza547/wow-recorder/issues/886 as part of this too. 

Should also fix while here https://github.com/aza547/wow-recorder/issues/843 although need to look closer. 

Migrates react table to the latest v9: https://tanstack.com/table/latest/docs/framework/react/guide/migrating.


src/renderer/DeleteDialog.tsx
```diff
@@ -193,33 +381,30 @@ const DeleteDialog = ({
...
-              variant="destructive"
-              type="submit"
-              onClick={doDelete}
-              disabled={table.getSelectedRowModel().rows.length < 1}
-            >
-              {getLocalePhrase(appState.language, Phrase.DeleteButtonTooltip)}
-            </Button>
+            <Button variant="ghost">Close</Button>
           </DialogClose>
+          <Button variant="destructive">Delete ({count})</Button>
```
Alkwer: The Delete button has no onClick, so deleting currently does not work - prob you still need to connect it to doDelete.

src/renderer/DeleteDialog.tsx
```diff
@@ -21,129 +21,317 @@ import {
...
+  };
+
+  const columns: ColumnDef<typeof stockFeatures, RendererVideo, unknown>[] = [
+    {
+      id: 'Lock',
+      accessorKey: 'isProtected',
+      accessorFn: (v) => v,
+      cell: populateLockCell,
+    },
+    {
```
Alkwer, line 95: In TanStack Table v9, accessorFn takes precedence over accessorKey, so getValue() returns the whole video object and every row displays the cloud icon

src/renderer/DeleteDialog.tsx
```diff
@@ -21,129 +21,317 @@ import {
...
+      id: 'Status',
+      accessorFn: (v) => v,
+      accessorKey: 'encounterName',
+      cell: (ctx) => populateTagStatusCell(ctx),
+    },
+  ];
+
+  const data = useMemo<Array<RendererVideo>>(() => {
+    // TODO handle multi row delete
+    const parent = deleteDialogVideoTargetIds[0]
```
Alkwer: I see you have a comment in the code but maybe it's good to have the remainder also as a review comment, not to forget to implement handling multiple rows, since it's quite important part

src/renderer/DeleteDialog.tsx
```diff
@@ -193,33 +381,30 @@ const DeleteDialog = ({
...
   return (
-    <Dialog>
+    <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogTrigger asChild>{children}</DialogTrigger>
       <DialogContent>
         <DialogHeader>
-          <DialogTitle>
-            {getLocalePhrase(appState.language, Phrase.AreYouSure)}
-          </DialogTitle>
+          <DialogTitle>Delete Manager</DialogTitle>
```
Alkwer: Might be better to use getLocalPhrase instead?

src/renderer/LockDialog.tsx
```diff
@@ -0,0 +1,351 @@
...
+          </div>
+        </ScrollArea>
+      </div>
+    );
+  };
+
+  const renderLockAllButton = () => {
+    const actionIsLock = data.some((v) => !v.isProtected);
+    const includesCloud = data.some((v) => v.cloud);
+    const label = actionIsLock ? 'Lock All' : 'Unlock All';
```
Alkwer: Same localization question as in DeleteDialog.tsx

src/renderer/TagDialog.tsx
```diff
@@ -8,115 +15,412 @@ import {
...
 
   return (
     <Dialog open={open} onOpenChange={handleOpenChange}>
-      <DialogTrigger asChild>{children}</DialogTrigger>
       <DialogContent>
         <DialogHeader>
-          <DialogTitle>
-            {getLocalePhrase(language, Phrase.AddADescription)}
-          </DialogTitle>
+          <DialogTitle>Tag Manager</DialogTitle>
```
Alkwer: Same localization question as in DeleteDialog.tsx

src/renderer/DeleteDialog.tsx
aza547: yeah, generally i leave the translations to the end as they often change a bit

src/renderer/DeleteDialog.tsx
aza547: nice spot!

aza547:
I think just localization to do now. Need to have a review but taking a break. Did consider trying to fix the lack of checking for rsp.ok on a bunch of these calls but this PR is already getting too large for my liking.

src/renderer/DeleteDialog.tsx
```diff
@@ -21,205 +21,417 @@ import {
...
       language,
       Phrase.From,
-    )} ${Math.max(selectedRowCount, 1)} ${getLocalePhrase(language, Phrase.Rows)}.`;
+    )} ${rows} ${getLocalePhrase(language, Phrase.Rows)}.`;
 
-    return (
-      <div className="text-sm">
-        <p className="inline ">{warning}</p>
+    const selected = table.getSelectedRowModel().rows;
+    const inScopeLocked = selected.filter((row) => row.original.isProtected);
```
Alkwer: Could we derive the locked warning from the same set as getVideosToDelete()? In multi-row mode the table only contains the first target group, while deletion expands every targetVideoId. This means the dialog can say there are no locked recordings even though a locked recording from another selected group will be deleted.

src/renderer/CategoryPage.tsx
```diff
@@ -531,15 +565,18 @@ const CategoryPage = (props: IProps) => {
...
       return (
         <Tooltip content={tooltip}>
           <div>
             <DeleteDialog
-              key={toDelete.map((v) => v.videoName).join(',')} // Forces a remount on selection change.
-              inScope={toDelete}
-              appState={appState}
+              open={dialog === DialogType.DELETE}
+              onOpenChange={setDeleteDialog}
+              targetVideoIds={ids}
```
Alkwer, line 580: Could we derive targetVideoIds from the same effective selection as selectedViewpoints, or update the row selection when filters change? If the selected row is filtered out, selectedViewpoints falls back to the first visible row but selectedRows is empty, so Delete remains enabled while the dialog receives an empty array and shows Delete (0).

src/renderer/KillVideoDialog.tsx
```diff
@@ -25,46 +24,70 @@ import { Info } from 'lucide-react';
...
-  });
+
+    setSegments(segs);
+  }, [open]);
+
+  // Our select component only accepts strings annoyingly.
+  const [fps, setFps] = useState('60');
+  const [singleAudio, setSingleAudio] = useState(false);
+
+  const [audioTrackPlayer, setAudioTrackPlayer] = useState('');
```
Alkwer: Should we initialize or reset audioTrackPlayer whenever this dialog opens for a new target? It currently starts empty and can also retain a player from the previous group. If Single Player Audio is enabled in either case, findIndex() returns -1 and the render silently falls back to using all audio tracks.

Alkwer:
I found those 3 additional points to think about, aprat from localization, the rest looks fine

src/renderer/KillVideoDialog.tsx
aza547: Yes, good spot! I've fixed that, and also reset the other fields on open changing too.

src/renderer/components/Tables/Cells.tsx
```diff
@@ -173,7 +190,35 @@ export const populateDetailsCell = (
     );
   };
 
-  const renderTagIcon = () => {
+  const renderProtectedIconMulti = () => {
+    const tooltip = 'Open Multilock dialog';
```
aza547: needs localised

src/renderer/components/Tables/Cells.tsx
```diff
@@ -200,31 +245,72 @@ export const populateDetailsCell = (
...
+
+    let tooltip = noPermission
+      ? getLocalePhrase(language, Phrase.GuildNoPermission)
+      : getLocalePhrase(language, Phrase.TagButtonTooltip);
+
+    const foundTag = toTag.map((v) => v.tag).find((t) => t);
+
+    if (foundTag) {
+      icon = <FolderMessageSquareMore size={18} />;
+      tooltip = `Open multitag dialog`;
```
aza547: needs localised

src/renderer/components/Tables/Cells.tsx
```diff
@@ -200,31 +245,72 @@ export const populateDetailsCell = (
...
-            </Button>
-          </TagDialog>
+            {icon}
+          </Button>
         </div>
       </Tooltip>
     );
   };
 
+  const n = [video, ...video.multiPov].length;
```
aza547: better var name

src/renderer/CategoryPage.tsx
```diff
@@ -531,15 +566,28 @@ const CategoryPage = (props: IProps) => {
         ? getLocalePhrase(language, Phrase.GuildNoPermission)
         : getLocalePhrase(language, Phrase.BulkDeleteButtonTooltip);
 
+      // Only want to pass the parent IDs into the delete dialog.
+      const ids = [
+        ...new Set(
+          toDelete
+            .map((r) => r.uniqueId)
```
aza547: rv

src/renderer/KillVideoDialog.tsx
```diff
@@ -25,46 +24,75 @@ import { Info } from 'lucide-react';
...
+  const [resolution, setResolution] =
+    useState<keyof typeof obsResolutions>('1920x1080');
+
+  const [segments, setSegments] = useState<KillVideoSegment[]>([]);
+
+  useEffect(() => {}, [open]);
```
aza547: ?

src/renderer/LockDialog.tsx
```diff
@@ -0,0 +1,382 @@
...
+  const setLock = (videos: Array<RendererVideo>, lock: boolean) => {
+    const disk = videos.filter((v) => !v.cloud);
+    const cloud = videos.filter((v) => v.cloud);
+
+    ipc.sendMessage('videoButtonDisk', ['protect', lock, disk]);
+    ipc.sendMessage('videoButtonCloud', ['protect', lock, cloud]);
+
+    setVideoState((prev) => {
+      return prev.map((rv) => {
+        return videos.some((v) => v.uniqueId == rv.uniqueId)
```
aza547: confusing variable names rv / v / videos

src/renderer/LockDialog.tsx
```diff
@@ -0,0 +1,382 @@
...
+            {playerName}
+          </div>
+        </div>
+      );
+    };
+
+    return <div className="flex truncate">{renderSpecAndName()}</div>;
+  };
+
+  const populateLockedStatusCell = (
```
aza547: should review these functions to see if they can be shared across dialogs

src/renderer/LockDialog.tsx
```diff
@@ -0,0 +1,382 @@
...
+    }
+    return (
+      <div className="flex truncate text-sm">
+        {getLocalePhrase(language, Phrase.EligibleForAutomaticDeletion)}
+      </div>
+    );
+  };
+
+  {
+    /* // share this stuff where sensible with other dialogs */
```
aza547: delete this after reviewing

src/renderer/TagDialog.tsx
```diff
@@ -8,115 +15,482 @@ import {
...
+import SaveIcon from '@mui/icons-material/Save';
+import CloudIcon from '@mui/icons-material/Cloud';
+import CircularProgress from '@mui/material/CircularProgress/CircularProgress';
 
 interface IProps {
-  tag: string;
-  videos: RendererVideo[];
-  setVideoState: Dispatch<SetStateAction<RendererVideo[]>>;
-  children: React.ReactNode;
+  open: boolean;
```
aza547, line 40: probably need some permission stuff here - pass appstate and look at cloudstatus?

src/renderer/KillVideoDialog.tsx
```diff
@@ -25,46 +24,75 @@ import { Info } from 'lucide-react';
...
+        video: rv,
+        start: idx * segmentDuration,
+        stop: (idx + 1) * segmentDuration,
+      })),
+    );
+
+    setFps('60');
+    setResolution('1920x1080');
+    setSingleAudio(false);
+    setAudioTrackPlayer(sourcesRef.current[0]?.player?._name || '');
```
Alkwer: Could we use the same stable value, such as uniqueId, for the audio option, initial selection, and lookup? player._name is optional: for a source without it, the option displays videoName, while the state initializes to an empty string and findIndex() still compares only _name, so Single Player Audio silently falls back to all audio tracks.

src/renderer/TagDialog.tsx
```diff
@@ -8,115 +15,482 @@ import {
...
-  };
 
-  const clearTag = (event: React.MouseEvent<HTMLElement>) => {
-    event.stopPropagation();
-    saveTag('');
+    if (!row.getIsSelected()) {
+      row.getToggleSelectedHandler()(event);
+    }
+
+    setInnerTag(row.original.tag ?? '');
```
Alkwer: Could we only reload innerTag when the click actually changes the selected row? Clicking the already selected row while a save is pending restores the old tag in the textarea, but the timer still persists the newer value, leaving the dialog out of sync.

src/renderer/components/Tables/Cells.tsx
```diff
@@ -173,7 +190,35 @@ export const populateDetailsCell = (
     );
   };
 
-  const renderTagIcon = () => {
+  const renderProtectedIconMulti = () => {
+    const tooltip = 'Open Multilock dialog';
+
+    const icon = [video, ...video.multiPov].some((v) => v.isProtected) ? (
```
Alkwer: Should this use every() rather than some()? In a mixed group, some() shows the folder as locked even though at least one viewpoint remains unlocked. That seems to conflict with the behavior discussed in #843, where mixed groups should remain visually unlocked.

src/renderer/LockDialog.tsx
```diff
@@ -0,0 +1,382 @@
...
+    if (isProtected) {
+      return (
+        <div className="flex truncate text-sm">
+          {getLocalePhrase(language, Phrase.SafeFromAutomaticDeletion)}
+        </div>
+      );
+    }
+    return (
+      <div className="flex truncate text-sm">
+        {getLocalePhrase(language, Phrase.EligibleForAutomaticDeletion)}
```
Alkwer: Could this say “Unlocked” rather than “Eligible for automatic deletion” for now? The housekeeper still protects the whole correlated group whenever any member is locked, so an unlocked member of a mixed group is not actually eligible for automatic deletion.

aza547: That's true, but when I change the housekeeper behaviour it will become true. I'd rather get ahead of it and ship this change that is defensive (says eligible for deletion when its not) than the opposite. I think that makes sense. Let me know if you disagree. I also will spin over all the videos in the database and mark them locked if they currently are in a locked group too, before I change the behaviour (maybe at the time this ships? not 100% sure).

src/renderer/components/Tables/Cells.tsx
aza547: I think this is right the way it is. My intent here is that:
- any row with a locked video shows this
- the locked filter will show the rows that have a locked video within

appreciate it might lead users to think the whole row is locked though, but I think that's the best option still

src/renderer/TagDialog.tsx
aza547: Yes - good shout!

src/renderer/LockDialog.tsx
aza547: ...and the lock symbol i think makes it clear that it is unlocked. So this is really a bit of extra information for users who aren't familiar with what locked means.

src/renderer/KillVideoDialog.tsx
aza547: Yeah agree this is sensible. I'll see what I can do. Sadly the product of lots of added fields over the years that could possibly (but should really never) be undefined. Probably only possible for videos recorded like 4+ years ago with an ancient metadata format.

aza547: i've just disabled the render button while there is no valid seleciton, I think that is the best I can do for now

src/renderer/LockDialog.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const previousOpen = useRef(open);
+  const previousParentId = useRef(targetVideoId);
+
+  useEffect(() => {
+    // Close an open dialog if the parent video has been deleted by another
+    // user. That should be rare enough that this isn't too annoying.
+    if (
+      open &&
+      previousOpen.current &&
+      previousParentId.current !== targetVideoId
```
Alkwer: Could we base this check on whether targetVideoId still resolves in parentLookupMap, and apply the same fix in TagDialog? The target ID itself does not change when another user deletes that parent; only the lookup changes, so this condition remains false and the open dialog becomes empty.

src/localisation/phrases.ts
```diff
@@ -556,6 +556,22 @@ enum Phrase {
...
+  SafeFromAutomaticDeletion,
+  EligibleForAutomaticDeletion,
+  NoCustomTag,
+  DeleteSelectionContainsNoLocked,
+  DeleteIsPermanent,
+  TagDescription,
+  ClearAll,
+  Diagnostics,
+  UnknownDevice,
+  OpenLockDialog,
```
Alkwer, line 573: Could we add these two keys (this and the next one) to the German, Korean, and Chinese locale maps as well, using English fallbacks if translations are not available yet? getLocalePhrase indexes the selected locale directly, so these tooltips are currently undefined outside English.

src/renderer/TagDialog.tsx
```diff
@@ -8,115 +15,364 @@ import {
...
 import { Language, Phrase } from 'localisation/phrases';
 import { Textarea } from './components/TextArea/textarea';
+import {
+  ColumnDef,
+  flexRender,
+  Row,
+  stockFeatures,
+  useTable,
+} from '@tanstack/react-table';
+import { getVideoGroup, getVideoGroupIds } from './rendererutils';
```
Alkwer: Could we remove the unused getVideoGroupIds import, unless it will be reused for the remote-deletion check? It currently introduces a new no-unused-vars lint error.

PR #899, opened 2026-08-28
Handle mid-line combat log reads

sepdotgg:
# Summary

Combat log watcher hardening to help address [this reported issue on Discord](https://discord.com/channels/1004860808737591326/1542544430651678780).

## TL;DR

- Parse up to the last `\n` in the file. Throw away the rest.
- Update the position marker to be previous + bytes read. Next read starts where it left off.
- Add a unit test for various chunk sizes against existing log files in the repo.
  - Also verifies that both LF and CRLF variations behave the same way, just in case.
- Fix the existing [Basic.test.ts](https://github.com/aza547/wow-recorder/compare/main...sepdotgg:wow-recorder:handle-midline-win?expand=1#diff-4e19aefeef6675c9c9039fe7a11cf147bd21234523a68421818f2b6f2a1cfd65) parser test that was already broken.
- **Want feedback on this one:** Add a try/catch around the line handler so failures don't totally brick parsing until app restart. If it fires, it results in a degraded experience (ie, you might have missed an ENCOUNTER_END or something), but either way it would have been dead. Still logs an error.

## The root cause

The previous combat log watcher/reader assumed that all reads would land on a line ending (`\n`) character. On Windows, this is largely true, since ReadDirectoryChangesW doesn't fire until the entire write completes, and Windows won't update the file's size until that write happens. From the reader's perspective, the write is atomic, even if under the hood it's still laying down bytes in chunks.

On other platforms (potentially network filesystems on Windows as well) this is not guaranteed.

For modern non-FUSE filesystems (btrfs, ext4, modern kernel 7.1 ntfs), this is usually not an issue. Two things mitigate it: Firstly, VFS won't fire the inotify until all chunks have been written and secondly, these filesystems typically chunk in 128KB pages. So unless WoW flushes more than that at a time, _and_ you land an `fs.stat` in the microsecond window between the notify and a subsequent partial chunk, you won't hit the issue.

However, for FUSE based drivers, like the much older `ntfs-3g`, the default is 4KB chunks, and the notify can happen on every chunk write, rather than at the end. And that's the bug.

## The fix

The perfectly efficient correct fix is to work like a traditional tailing library that operates on full lines: Parse up to the last `\n` in the file, buffer the rest, then prepend it to to the next read.

I opted to go with a slightly simpler fix here that's just as correct:

- Parse up to the last `\n` in the file. Throw away the rest.
- Update the position marker to be previous + bytes read.

This results in a small amount of re-reading partial chunks on the aforementioned filesystem (basically just ntfs-3g) above, but avoids some extra complexity that comes with a full tail reader that optimizes for efficiency above all else. The cost is up to 4KB extra per read (but typically far less because you're reading to EoF anyway) + string parsing up to 4KB in memory, which is microseconds. On Windows, it's functionally equivalent, since the remainder is always 0 bytes.

### DIFF
--- a/src/parsing/CombatLogWatcher.ts
+++ b/src/parsing/CombatLogWatcher.ts
@@ -154,14 +154,25 @@ export default class CombatLogWatcher extends EventEmitter {
       return;
     }
 
-    await this.parseFileChunk(fullPath, bytesToRead, startPosition);
-    this.state[fullPath] = currentInfo;
+    const consumed = await this.parseFileChunk(
+      fullPath,
+      bytesToRead,
+      startPosition,
+    );
+    // for this path, the size represents the bytes parsed in total, not
+    // necessarily the reported fs.stat of the file, which could be from
+    // a partial line chunk write.
+    this.state[fullPath] = { ...currentInfo, size: startPosition + consumed };
   }
 
   /**
    * Parse a chunk of the file of length bytes from a specified position.
    */
-  private async parseFileChunk(file: string, bytes: number, position: number) {
+  private async parseFileChunk(
+    file: string,
+    bytes: number,
+    position: number,
+  ): Promise<number> {
     const buffer = Buffer.alloc(bytes);
     const handle = await open(file, 'r');
     const { bytesRead } = await read(handle, buffer, 0, bytes, position);
@@ -178,15 +189,42 @@ export default class CombatLogWatcher extends EventEmitter {
 
     this.emit('WARCRAFT_RECORDER_LOG_ACTIVITY');
 
-    const lines = buffer
+    // On some platforms (eg, anything VFS based, especially ntfs-3g), the notify can be sent mid-line.
+    // We can't assume that the last data of the read is a completed log line.
+    // Instead, only parse up to the last new line char return how much we parsed
+    // so that the next read can start from that position.
+    const valid = buffer.subarray(0, bytesRead);
+    const newline = valid.lastIndexOf(0x0a); // \n
+
+    if (newline < 0) {
+      return 0;
+    }
+
+    const consumed = newline + 1;
+
+    const lines = valid
+      .subarray(0, consumed)
       .toString('utf-8')
       .split('\n')
       .map((s) => s.trim())
       .filter((s) => s);
 
     lines.forEach((line) => {
-      this.handleLogLine(line);
+      // If parsing the line fails, at least allow the parser to continue.
+      // State could be undefined, but the alternative is that all
+      // log watching is bricked until an app restart.
+      try {
+        this.handleLogLine(line);
+      } catch (error) {
+        console.error(
+          '[CombatLogWatcher] Failed to handle log line:',
+          line,
+          error,
+        );
+      }
     });
+
+    return consumed;
   }
 
   /**
--- a/src/parsing/CombatLogWatcher.ts
+++ b/src/parsing/CombatLogWatcher.ts
@@ -159,14 +159,18 @@ export default class CombatLogWatcher extends EventEmitter {
       bytesToRead,
       startPosition,
     );
-    // for this path, the size represents the bytes parsed in total, not
-    // necessarily the reported fs.stat of the file, which could be from
-    // a partial line chunk write.
     this.state[fullPath] = { ...currentInfo, size: startPosition + consumed };
   }
 
   /**
-   * Parse a chunk of the file of length bytes from a specified position.
+   * Parse a chunk of the file of length bytes from a specified position to the final new line.
+   *
+   * On some platforms (eg, anything VFS based, especially ntfs-3g), the notify can be sent mid-line.
+   * We can't assume that the last data of the read is a completed log line. Instead, only parse
+   * up to the last new line char and return how much we parsed so that the next read can start from
+   * that position.
+   *
+   * @returns consumed bytes
    */
   private async parseFileChunk(
     file: string,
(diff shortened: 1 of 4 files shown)


src/parsing/CombatLogWatcher.ts
```diff
@@ -178,15 +189,42 @@ export default class CombatLogWatcher extends EventEmitter {
...
+    const lines = valid
+      .subarray(0, consumed)
       .toString('utf-8')
       .split('\n')
       .map((s) => s.trim())
       .filter((s) => s);
 
     lines.forEach((line) => {
-      this.handleLogLine(line);
+      // If parsing the line fails, at least allow the parser to continue.
```
aza547: nit: tiny bit neater for this to be in the handlerLogLine fn. 
Don't think the comment makes sense re state being undefined (how could that cause this to throw?)

src/parsing/CombatLogWatcher.ts
```diff
@@ -178,15 +189,42 @@ export default class CombatLogWatcher extends EventEmitter {
 
     this.emit('WARCRAFT_RECORDER_LOG_ACTIVITY');
 
-    const lines = buffer
+    // On some platforms (eg, anything VFS based, especially ntfs-3g), the notify can be sent mid-line.
+    // We can't assume that the last data of the read is a completed log line.
+    // Instead, only parse up to the last new line char return how much we parsed
+    // so that the next read can start from that position.
+    const valid = buffer.subarray(0, bytesRead);
+    const newline = valid.lastIndexOf(0x0a); // \n
```
aza547: so what's happening to \r\n here? the \r is included in the buffer passed to LogLine? Worth a comment?

src/parsing/CombatLogWatcher.ts
```diff
@@ -178,15 +189,42 @@ export default class CombatLogWatcher extends EventEmitter {
 
     this.emit('WARCRAFT_RECORDER_LOG_ACTIVITY');
 
-    const lines = buffer
+    // On some platforms (eg, anything VFS based, especially ntfs-3g), the notify can be sent mid-line.
+    // We can't assume that the last data of the read is a completed log line.
+    // Instead, only parse up to the last new line char return how much we parsed
+    // so that the next read can start from that position.
+    const valid = buffer.subarray(0, bytesRead);
```
aza547: why "valid"?

src/parsing/CombatLogWatcher.ts
```diff
@@ -154,14 +154,25 @@ export default class CombatLogWatcher extends EventEmitter {
...
     }
 
-    await this.parseFileChunk(fullPath, bytesToRead, startPosition);
-    this.state[fullPath] = currentInfo;
+    const consumed = await this.parseFileChunk(
+      fullPath,
+      bytesToRead,
+      startPosition,
+    );
+    // for this path, the size represents the bytes parsed in total, not
```
aza547: not sure this comment adds anything that isn't obvious?

src/parsing/CombatLogWatcher.ts
```diff
@@ -154,14 +154,25 @@ export default class CombatLogWatcher extends EventEmitter {
...
+    // necessarily the reported fs.stat of the file, which could be from
+    // a partial line chunk write.
+    this.state[fullPath] = { ...currentInfo, size: startPosition + consumed };
   }
 
   /**
    * Parse a chunk of the file of length bytes from a specified position.
```
aza547: "to the final new line"? 
`@returns consumed bytes`?

src/parsing/CombatLogWatcher.ts
```diff
@@ -178,15 +189,42 @@ export default class CombatLogWatcher extends EventEmitter {
 
     this.emit('WARCRAFT_RECORDER_LOG_ACTIVITY');
 
-    const lines = buffer
+    // On some platforms (eg, anything VFS based, especially ntfs-3g), the notify can be sent mid-line.
```
aza547: reckon this can be in the block comment rather than inline

aza547 (changes requested):
Thanks for submitting this. Looks fine in general, a few nits to tidy up.

How long do the tests take to run? A little nervous if they are reading those massive log files I've checked in, but I basically never run the tests in this repo anyway.

src/parsing/CombatLogWatcher.ts
sepdotgg: - I can move it to inside of the `handleLogLine` func, but to me that's hiding potential error paths. A `void handleLogLine` returning cleanly reads as success to me.

  `handleLogLine` is also public, so _everyone_ would get their errors swallowed. (eg, the Test combat log button would also swallow, may or may not be desirable). An alternative would be to create a separate private `tryHandleLogLine` if you'd prefer it not be in `parseFileChunk`.

- Regarding the comment: State is "undefined" in a spec behavior sense, not in a JS value sense. If the log line fails to construct, we've missed data, and therefore the the state of the combat log as we know it may or may not be accurate. 
  - I've reworded it to be more clear.

src/parsing/CombatLogWatcher.ts
sepdotgg: > the \r is included in the buffer passed to LogLine?

No, the `trim()` below removes it before it is passed to LogLine. This just finds the position where you start splitting. Will add a comment to call it out where the split happens.

The Chunking unit tests I added against the files exercises this case.

src/parsing/CombatLogWatcher.ts
sepdotgg: The buffer allocates size `bytes` but it might not be filled. `valid` is the portion that actually contains data. I've updated it to `populatedSlice` to hopefully be more clear.

src/parsing/CombatLogWatcher.ts
sepdotgg: I removed it. It's clear at this site, which is what counts.

My original reasoning was it is slightly unclear if you look at the `FileInfo` struct which implies `size` is the file's metadata size, which isn't strictly true when constructed here.

src/parsing/CombatLogWatcher.ts
sepdotgg: Updated.

src/parsing/CombatLogWatcher.ts
sepdotgg: Moved.

sepdotgg:
> Thanks for submitting this. Looks fine in general, a few nits to tidy up.
> 
> How long do the tests take to run? A little nervous if they are reading those massive log files I've checked in, but I basically never run the tests in this repo anyway.

~7s on my SSD, probably double on a spinning disk. Tests aren't wired into CI today, but even if they were, this isn't something that runs on every local build (as you said), so that felt like a worthwhile tradeoff for more realistic coverage of the parser path when it changes.

**Other**: Only one unresolved comment about `handleLogLine`.

PR #903, opened 2026-09-04
feat: add force stop hotkey

delasteve:


src/main/Manager.ts
```diff
@@ -898,6 +899,23 @@ export default class Manager {
...
+        // This user is updating their settings. Don't do anything.
+        return;
+      }
+
+      if (!isForceStopHotKey(event)) {
+        return;
+      }
+
+      console.info('[Manager] Force stop hotkey pressed');
+      this.forceStop();
```
Alkwer: This key event can also trigger the manual recording listener below. For example, with manual recording bound to K and force-stop bound to Ctrl+K, pressing Ctrl+K matches both because additional modifiers are accepted. Force-stop clears the current activity, then the manual listener starts a new recording. Can we ensure that one key event triggers only one action, with force-stop taking precedence?

src/renderer/HotKeySettings.tsx
```diff
@@ -0,0 +1,138 @@
...
+          keyPressEvent.type === PTTEventType.EVENT_MOUSE_PRESSED ||
+          keyPressEvent.type === PTTEventType.EVENT_MOUSE_RELEASED
+        ) {
+          // Don't accept mouse events
+          keyPressEvent = await getNextKeyOrMouseEvent();
+        }
+
+        setForceStopHotKeyFieldFocused(false);
+        setForceStopHotKey(keyPressEvent);
+        setForceStopKeyConfig(keyPressEvent);
```
Alkwer: The pending capture is not cancelled when the input loses focus or the component unmounts. Clicking the field, clicking elsewhere, and then pressing A still saves A as the force-stop hotkey. This was reproduced in a component test. Can we cancel the capture when binding ends and ignore any pending result?

src/renderer/HotKeySettings.tsx
```diff
@@ -0,0 +1,138 @@
...
+            side="right"
+          >
+            <Info size={20} className="inline-flex ml-2" />
+          </Tooltip>
+        </Label>
+        <Input
+          ref={forceStopHotKeyInputRef}
+          name="forceStopHotKeyInput"
+          value={getHotkeyString()}
+          onFocus={() => setForceStopHotKeyFieldFocused(true)}
```
Alkwer: Once a hotkey has been assigned, there is no way to disable it through the UI. Escape is captured as another binding rather than clearing it. Could we add a clear/unbind action that resets the key to -1 and clears its modifiers?

src/main/Manager.ts
```diff
@@ -898,6 +899,23 @@ export default class Manager {
       this.refreshCombatLoggingStatus();
     });
 
+    /**
+     * Force stop the current recording.
+     */
+    uIOhook.on('keydown', (event: UiohookKeyboardEvent) => {
```
aza547, line 905: reckon this should live inside the other listener (below) rather than duplicating the listener, less overhead

src/renderer/HotKeySettings.tsx
```diff
@@ -0,0 +1,138 @@
...
+          onBlur={() => setForceStopHotKeyFieldFocused(false)}
+          readOnly
+        />
+      </div>
+    );
+  };
+
+  return (
+    <div className="flex flex-row flex-wrap gap-x-4">
+      {getForceStopHotKeySelect()}
```
aza547, line 133: It might be nice to group the manual start/stop hotkey in this section too. Just the hotkey field that is. What do you think? The other hotkey is the PTT one but I think it makes sense for that to remain in the audio section.

src/renderer/rendererutils.ts
```diff
@@ -785,25 +785,34 @@ const getPTTKeyPressEventFromConfig = (
   };
 };
 
-const getManualRecordHotKeyFromConfig = (
-  config: ConfigurationSchema,
+const getHotKeyFromConfig = (
```
aza547, line 788: In general a little worried about this code - it does a bit more that it needs to (and it did before) and is called on every single keypress. Probably a "computers are actually really quite fast" moment and is fine but you are duplicating the work here to check if it's a match to two different key binds now. There are probably some easy performance wins in here if you fancied trying.

src/main/Manager.ts
```diff
@@ -898,6 +899,23 @@ export default class Manager {
...
+      if (this.manualHotKeyDisabled) {
```
aza547: This isn't relevant here - but do you need an equivalent for the force stop? What if you are setting this hotkey while in a run? That was the reason for this existing for the manual case. I can't remember exactly the condition but I think it just blocks the hotkey doing anything while the selector is active.

aza547 (changes requested):
Looks great overall. A few minor bits to consider in-line and more interesting questions below.

- Should we prevent players from setting the same hotkey to both this and manual stop start?

- What happens if a player is in a manual recording and hits force stop? If they hit manual stop AND force stop in quick succession? What should happen?

I'll do some testing of the latter too. Also interested in making sure the "spam force stop key" behaviour is sensible (and subsequent stops soon after are not acted on).

FWIW, separately I want to remove the need to hover the status to see the force stop button. Maybe as simple as this (with appropriate tooltip). There is a lot of conditional stuff in the status card though. 

image

src/main/Manager.ts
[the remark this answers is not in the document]
aza547: this is probably a good answer to my question earlier

PR #906, opened 2026-09-06
Fix empty overrun inputs persisting negative values

### LINKED ISSUE #863: overrun fields can be negative?

Alkwer:
Clearing either overrun input uses -1 as a temporary empty value. Both settings were added to the save payload before the validity checks, so clearing a field still persisted -1 and could shorten the recording duration.

Only add the overrun settings through the existing non-negative checks. Clearing a field now preserves its last saved value.

Fixes #863.

Validation:
- Targeted ESLint passed.
- Production build passed on Node.js 24.
- Verified in a real local Electron app with an isolated profile: clearing both inputs, saving 0 and 60, changing another setting while the inputs are empty, reopening settings, and restarting the app. This was checked on the local 7.13.0 build; the changed component is identical to this PR.
- Full lint reports 36 existing errors outside the changed component.

No test files are included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 ### Added
 ### Fixed
+- [Issue 863](https://github.com/aza547/wow-recorder/issues/863) - Preserve the last valid overrun setting when its input is cleared.
 - Add Nymrissa "World LFR" difficulty.
 
 ## [7.13.2] - 2026-08-30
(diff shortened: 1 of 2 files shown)


aza547:
hmm i actually cannot reproduce this bug now? not sure if it was fixed inadvertently 

either way this code seems correct, although I think it is missing the `minEncounterDuration`, i'll add that as well post-merge

aza547:
oh, I see, clearing the field sets to -1 but doesn't display it as -1

aza547:
https://github.com/aza547/wow-recorder/commit/5a5f145e395a5649b8c1846e9b3280d7d93a07eb

